Issue #16, opened 2019-10-17
Setup VSCode to run webpack on file save


gannonprudhomme:
Should find out how to configure VSCode to run webpack automatically anytime there's a change to a `.tsx` file.

firejake308:
Hint for whoever wants to work on this: I had this working for my last Vue project, and it worked by using `webpack-dev-server`, so you may want to try that out

gannonprudhomme:
Also my implementation: 
While it doesn't run on save, both vim & non-vim users could add:

To `.vscode/tasks.json`:

```
{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "npm",
            "script": "dev",
            "path": "autoscheduler/frontend/src/",
            "group": {
                "kind": "build",
                "isDefault": true
            }
        }
    ]
}
```

And to the global `keybindings.json`(do Ctrl+Shift+P then `Open Keyboard Shortcuts(JSON)`, add:

```     { "key": "ctrl+b", "command": "workbench.action.tasks.build"} ```

Which would run `npm run dev` when `Ctrl+b` is pressed.

While I did see some vim-specific keybindings/macros, I couldn't find anything that would run a command on `:w`. Also didn't look too deeply into running it on just a general save file, so not sure if that functionality exists.

firejake308:
Closed by #69

PR #17, opened 2019-10-18
Setup ESLint to work with VSCode

### LINKED ISSUE #13: Setup ESLint for Typescript
Need to configure ESLint for Typescript so that we can have a standard coding convention & catch basic syntax errors

Possibly integrate it into VSCode so that it runs on save? Might be able to do it with the ESLint VSCode extension

firejake308:
I have set up ESLint to work with VSCode. To run the linter, open `FrontendWorkspace.code-workspace` in VSCode. Once you have installed the ESLint plugin for VSCode, edit a JS, JSX, TS, or TSX file and hit save. The linter will auto fix as much of your code as it can and highlight the remaining errors. If you want to work in vim (like a real man) and lint your code, you can also run `npm run lint` from within the `frontend/src` folder, provided you have already run `npm install` to locally download the required packages for linting.

Closes #13


autoscheduler/frontend/src/App.tsx
```diff
@@ -1,9 +1,11 @@
...
\ No newline at end of file
+const App: React.SFC = function App() {
+  return (
+    <div>
+      Oi, put stuff here
+    </div>
+  );
+};
+
+export default App;
```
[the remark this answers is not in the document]
gannonprudhomme: Also, on the React tutorial this is how they suggest making the classes

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -0,0 +1,40 @@
+module.exports = {
+  env: {
+    es6: true,
+    node: true,
+  },
+  extends: [
+    'airbnb',
```
gannonprudhomme: On the tutorial I followed, it also had 
`        'plugin:react/recommended', `
`        'plugin:@typescript-eslint/recommended', `

As an extended plugin. Is this something we need to add?

autoscheduler/frontend/src/FrontendWorkspace.code-workspace
```diff
@@ -0,0 +1,20 @@
...
+		"eslint.run": "onSave",
+		"eslint.autoFixOnSave": true,
+		"eslint.validate": [
+			"javascript",
+			"javascriptreact",
+			{"language": "typescript", "autoFix": true},
+			{"language": "typescriptreact", "autoFix": true}
+		]
+	}
+}
```
gannonprudhomme: We're not going to include text-editor related configs in the project, as we're trying to let people use whatever they want (I know Scott uses Atom). Although, Ryan and I do both of course recommend VSCode. 

Instead, add add the file to the Wiki [here](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/wiki/VS-Code-Setup), with a bit of steps and background on what adding it will do.

gannonprudhomme: Likewise, remove the file from git and add it `.code-workspace` to the `.gitignore` when you get the chance 😄

gannonprudhomme:
Also, I did "request changes" on the pull request not really knowing what it'll do, so I think just whenever you make changes/comments on it do the "re-request review"(refresh button next to my name under reviewers)

autoscheduler/frontend/src/App.tsx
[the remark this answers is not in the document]
firejake308: ESLint for React recommends that stateless components be written as functional components rather than class-based components. If we expect that our top-level App component will never manage any state, then we should leave it as a functional component. If, on the other hand, we expect that in the future, the top-level App will be managing some aspects of state, then I will go ahead and throw in a dummy state to make the linter happy until we decide exactly what that state should be.

rconn478:
Unless I'm mistaken, this pull request doesn't actually close #16, it runs ESLint on save but not webpack.

gannonprudhomme:
> Unless I'm mistaken, this pull request doesn't actually close #16, it runs ESLint on save but not webpack.

Oh yeah I didn't even think about that, you're right

gannonprudhomme:
While it's not running on save, for both vim & non-vim users, could add:

To `.vscode/tasks.json`:

```
{
    "version": "2.0.0",
    "tasks": [
        {
            "type": "npm",
            "script": "dev",
            "path": "autoscheduler/frontend/src/",
            "group": {
                "kind": "build",
                "isDefault": true
            }
        }
    ]
}
```

And to the global `keybindings.json`(do Ctrl+Shift+P then `Open Keyboard Shortcuts(JSON)`, add:

```     { "key": "ctrl+b", "command": "workbench.action.tasks.build"} ```

Which would run `npm run dev` when `Ctrl+b` is pressed

File: autoscheduler/frontend/src/App.tsx (hunk 1 above)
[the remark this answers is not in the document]
Comment (gannonprudhomme) [reply autoscheduler/frontend/src/App.tsx]: > If we expect that our top-level App component will never manage any state, then we should leave it as a functional component.

I think this should be safe to assume but I'm not positive. As you have more front-end experience than both of us, I'll leave it up to you.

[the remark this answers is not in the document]
Comment (firejake308) [reply autoscheduler/frontend/src/App.tsx]: Since we will be using Redux to manage application-wide state, I have left App.tsx as a functional component.

File: autoscheduler/frontend/src/FrontendWorkspace.code-workspace (hunk 3 above)
Comment (firejake308) [reply autoscheduler/frontend/src/FrontendWorkspace.code-workspace]: I've updated the Wiki and explained how to set up ESLint in the `.vscode` settings so that no changes would have to be made to the `.gitignore`

File: autoscheduler/frontend/src/.eslintrc.js (hunk 2 above)
Comment (firejake308) [reply autoscheduler/frontend/src/.eslintrc.js]: Ah, yes, thank you. These two plugins overwrite several ESLint rules that conflict with TypeScript and React development, so I added them in

**Comment by firejake308** [comment]
> Unless I'm mistaken, this pull request doesn't actually close #16, it runs ESLint on save but not webpack.

You're right, I misread that. I removed #16 from the description and moved it back to the inbox in ZenHub.

**Comment by firejake308** [comment]
> > Unless I'm mistaken, this pull request doesn't actually close #16, it runs ESLint on save but not webpack.
> 
> Oh yeah I didn't even think about that, you're right. Regardless I do like this functionality.
> 
> For ESLint, we could just modify it so that it would run everytime you typed, but just wouldn't auto-fix it(at least, that's my preference - I'd rather have the underlines telling me to fix things then for it to just fix it automatically).

Since the VSCode ESLint plugin config has been moved to `.vscode/settings.json`, you can just edit the value of `eslint.run` to `onType`

File: autoscheduler/frontend/src/App.tsx (hunk 1 above)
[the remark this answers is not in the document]
Comment (gannonprudhomme) [reply autoscheduler/frontend/src/App.tsx]: Oh yeah good call, resolving conversation

**Comment by gannonprudhomme** [comment]
I added `*.code-workspace` to the `.gitignore` just so we don't run into this problem again in the future, and in case people want to have their VSCode workspace file within the project itself.

File: autoscheduler/frontend/src/package.json [hunk 4]
```diff
@@ -6,18 +6,25 @@
   "scripts": {
     "test": "echo \"Error: no test specified\" && exit 1",
     "dev": "webpack --mode development ./index.tsx --output ../static/main.js",
-    "build": "webpack --mode production ./index.tsx --output ../static/main.js"
+    "build": "webpack --mode production ./index.tsx --output ../static/main.js",
+    "lint": "eslint .",
+    "lint-init": "eslint --init"
```
gannonprudhomme: Do we need to include `lint-init`, if we already are including the `.eslintrc` file?

autoscheduler/frontend/src/FrontendWorkspace.code-workspace
gannonprudhomme: Looks good!

gannonprudhomme:
Also, after following all of these steps(added to `settings.json`, ran `npm install`), I changed the `App.tsx` to see if it would Any idea why? (Troubleshooting atm)

The ESLint plugin output says `ESLint server is running`, yet it doesn't auto fix the issues to the form you get, nor does it tell me to correct the spacing from 4 spaces to 2. Only thing it tells me is `Missing return type on function` for `render()`.

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -23,10 +23,10 @@ module.exports = {
     '@typescript-eslint',
     'import',
   ],
-  rules: {
-    "import/no-extraneous-dependencies": "off",
-    "react/jsx-filename-extension": [1, {"extensions": [".jsx", ".tsx"]}],
-  },
+  extends: [
```
gannonprudhomme: Why have two extends? Shouldn't they all be in one?

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -0,0 +1,46 @@
...
+  },
+  plugins: [
+    'react',
+    '@typescript-eslint',
+    'import',
+  ],
+  extends: [
+    'plugin:react/recommended',
+    'plugin:@typescript-eslint/recommended',
+  ],
```
gannonprudhomme: You have `extends` defined in two places, so I believe the second one is overwriting the first. Thus, the `airbnb` coding conventions are being applied

autoscheduler/frontend/src/package.json
firejake308: No, sorry, I only needed that to generate the config file

autoscheduler/frontend/src/.eslintrc.js
firejake308: Ah, that makes sense. I forgot that I had already defined `extends` when I wrote the second one.

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -23,10 +23,10 @@ module.exports = {
     '@typescript-eslint',
     'import',
   ],
-  rules: {
-    "import/no-extraneous-dependencies": "off",
-    "react/jsx-filename-extension": [1, {"extensions": [".jsx", ".tsx"]}],
```
gannonprudhomme: I believe we still need line 28, as I'm currently getting `JSX not allowed in files with extension '.tsx'` from `eslint(react/jsx-filename-extension)`

gannonprudhomme:
Also, did your ESLint autochange `App.tsx` to the functional form? It's giving me errors/problems for it, but it's not auto-correcting.

firejake308:
No, I changed App.tsx to the functional form manually, at the request of ESLint

autoscheduler/frontend/src/.eslintrc.js
firejake308: Correct, I had made that error go away by accidentally overwriting the Airbnb rules, but I will add back in the tsx rule

PR #21, opened 2019-10-20
Setup Redux

### LINKED ISSUE #18: Setup Redux
Need to add an outline of Redux on the frontend, so the "skeleton" is there once we begin the project.

firejake308:
Closes #18 

Redux dependencies have been added to `package.json`, and the necessary files were created.

### DIFF
--- a/autoscheduler/frontend/src/package.json
+++ b/autoscheduler/frontend/src/package.json
@@ -24,9 +24,6 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-react-hooks": "^1.7.0",
-    "prop-types": "^15.7.2",
-    "react": "^16.10.2",
-    "react-dom": "^16.10.2",
     "source-map-loader": "^0.2.4",
     "ts-loader": "^6.2.0",
     "typescript": "^3.6.4",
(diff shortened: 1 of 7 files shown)


autoscheduler/frontend/src/package.json
```diff
@@ -33,5 +33,10 @@
     "typescript-eslint": "0.0.1-alpha.0",
     "webpack": "^4.41.0",
     "webpack-cli": "^3.3.9"
+  },
+  "dependencies": {
+    "@types/react-redux": "^7.1.5",
+    "react-redux": "^7.1.1",
+    "redux": "^4.0.4"
```
gannonprudhomme: These are the dependencies that are going to be needed for like production right? Not just for development, so shouldn't we also have react, typescript, etc in here? And have the dependencies only needed for development, i.e. ESLint, in `devDependencies`?

I originally put them in `devDependencies` for no particular reason, so really just wondering if we should move the necessary ones into `dependencies`

firejake308: Yes, React needs to be moved to production dependencies, or else we could run into problems building for production. However, since Typescript is a development-time tool that just generates JS, Typescript is fine where it is. Jest should also be fine in devDependencies, when that gets added.

firejake308: Done

firejake308: For the record, I moved React, ReactDOM, and PropTypes

PR #26, opened 2019-10-21
Added basic scraper models

gannonprudhomme:
Added a basic outline for the scraper models, so that the backend team has a simple base to work off of.

Was also wondering if we should add the `Meta` class to it(which, among other things, determines what database the model will be stored in), but I say we leave it up to them to do it manually


gannonprudhomme:
For some reason Git wasn't set up on my laptop when I made this commit, so I'm going to close this one and make the same changes but with my actual account.

PR #28, opened 2019-10-21
Setup Pylint

### LINKED ISSUE #12: Setup Pylint
Need to set up Pylint so that it works for everyone out-of-the-box

Will do so by creating a `.pylintrc` file

gannonprudhomme:
One of the things we need to figure out with this is to ignore some of the errors that aren't necessary, such as the error `Missing module docstring - pylint(missing-docstring)` for the line `from django.db import models`. Also, I think we need to have the `.pylintrc` file in order for Pylint to enable in the workspace(at least for me)

This isn't complete yet, just making this so we can discuss.

Closes #12

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,570 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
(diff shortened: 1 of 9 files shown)


gannonprudhomme:
Ack i messed this up, meant to do that commit on a different branch, just gonna redo this

PR #30, opened 2019-10-22
Setup pylint

gannonprudhomme:
Basically just wanna check that all of these sound right.

~~I also need to add a step to the README or wiki, as to get it working with `pylint-django` I believe you have to do `pylint --load-plugins pylint-django /path/to/folder/with/__init__.py/inside/`~~

The above is totally wrong. We may be able to provide it in the `load-plugins` thing in the rcfile, alternatively we can set VSCode pylint arguments

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,570 @@
+[MASTER]
+
+# A comma-separated list of package or module names from where C extensions may
+# be loaded. Extensions are loading into the active Python interpreter and may
+# run arbitrary code.
+extension-pkg-whitelist=
+
+# Add files or directories to the blacklist. They should be base names, not
+# paths.
+ignore=CVS
+
+# Add files or directories matching the regex patterns to the blacklist. The
+# regex matches against base names, not paths.
+ignore-patterns=
+
+# Python code to execute, usually for sys.path manipulation such as
+# pygtk.require().
+#init-hook=
+
+# Use multiple processes to speed up Pylint. Specifying 0 will auto-detect the
+# number of processors available to use.
+jobs=1
+
+# Control the amount of potential inferred values when inferring a single
+# object. This can help the performance when dealing with large functions or
+# complex, nested conditions.
+limit-inference-results=100
+
+# List of plugins (as comma separated values of python modules names) to load,
+# usually to register additional checkers.
+load-plugins=
+
+# Pickle collected data for later comparisons.
+persistent=yes
+
+# Specify a configuration file.
+#rcfile=
+
+# When enabled, pylint would attempt to guess common misconfiguration and emit
+# user-friendly hints instead of false-positive error messages.
+suggestion-mode=yes
+
+# Allow loading of arbitrary C extensions. Extensions are imported into the
+# active Python interpreter and may run arbitrary code.
+unsafe-load-any-extension=no
+
+
+[MESSAGES CONTROL]
+
+# Only show warnings with the listed confidence levels. Leave empty to show
+# all. Valid levels: HIGH, INFERENCE, INFERENCE_FAILURE, UNDEFINED.
+confidence=
+
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+disable=print-statement,
+        parameter-unpacking,
+        unpacking-in-except,
+        old-raise-syntax,
+        backtick,
+        long-suffix,
+        old-ne-operator,
+        old-octal-literal,
+        import-star-module-level,
+        non-ascii-bytes-literal,
+        raw-checker-failed,
+        bad-inline-option,
+        locally-disabled,
+        file-ignored,
+        suppressed-message,
+        useless-suppression,
+        deprecated-pragma,
+        use-symbolic-message-instead,
+        apply-builtin,
+        basestring-builtin,
+        buffer-builtin,
+        cmp-builtin,
+        coerce-builtin,
+        execfile-builtin,
+        file-builtin,
+        long-builtin,
+        raw_input-builtin,
+        reduce-builtin,
+        standarderror-builtin,
+        unicode-builtin,
+        xrange-builtin,
+        coerce-method,
+        delslice-method,
+        getslice-method,
+        setslice-method,
+        no-absolute-import,
+        old-division,
+        dict-iter-method,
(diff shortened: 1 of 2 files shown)


.pylintrc
```diff
@@ -0,0 +1,577 @@
...
+# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
+expected-line-ending-format=
+# Naming style matching correct class names.
+class-naming-style=PascalCase
+
+# Maximum number of nested blocks for function / method body
+max-nested-blocks=5
+
+# Maximum number of characters on a single line.
+max-line-length=90
```
gannonprudhomme: What do we want for max-line-length?

.pylintrc
```diff
@@ -0,0 +1,577 @@
...
+
+# String used as indentation unit. This is usually "    " (4 spaces) or "\t" (1
+# tab).
+indent-string='    '
```
gannonprudhomme: Says to indent using spaces

gannonprudhomme:
Do we want indention to be with 2 spaces or 4? I tend to prefer 4 as it helps to differentiate blocks better, which is especially important in python.

.pylintrc
```diff
@@ -0,0 +1,577 @@
...
+
+# Number of spaces of indent required inside a hanging or continued line.
+indent-after-paren=4
```
gannonprudhomme: Do we want indention to be with 2 spaces or 4? I tend to prefer 4 as it helps to differentiate blocks better, which is especially important in python.

.pylintrc
```diff
@@ -0,0 +1,577 @@
...
+
+# Expected format of line ending, e.g. empty (any line ending), LF or CRLF.
+expected-line-ending-format=LF
```
gannonprudhomme: Do we want to force LF line endings?

.pylintrc
firejake308: ESLint defaults to 100, I believe the default in Vim is 80, somewhere in that ballpark. 90 is fine, imo.

.pylintrc
firejake308: 4 for Python. Definitely.

firejake308: But wait, maybe 2 could be okay for a continued line? To differentiate a continued line from a new block?

gannonprudhomme: That sounds great to me honestly. I never know how much to space a continued line but it makes it easy to differentiate.

gannonprudhomme: Also just realized that's the property I was commenting on originally. The general tab size is another property, so I'm gonna find that property and add it to the top as well

.pylintrc
gannonprudhomme: If I remember correctly I think that by default git automatically converts line endings when it downloads the files(LF on Unix & CRLF on Windows) so it shouldn't matter.

Only one I know it matters is the entry point script for Docker has to be LF line endings, for when we use GitHub Actions

.pylintrc
gannonprudhomme: @rconn478 thots?

rconn478: 90 seems like a good compromise.

gannonprudhomme:
Gonna look into the following:
- Requiring module doc strings, as I don't think they're entirely necessarily(can discuss though)
  - Likewise, ignoring stuff like `django.db` not having module doc strings
- `too-few-public-methods` as its irrelevant for the Meta classes for Django models
- Requiring spaces before and after the beginning of comments

Some of this stuff may be satisfied by `pylint-django`, so need to look into that further.

gannonprudhomme:
**Note**: I messed up a bit somewhere with git and had to rebase, so if you pulled this recently instead of pulling & merging, do `git reset --hard origin/backend/setup-pylint`

gannonprudhomme:
Oops requested a review a bit preemptively, gotta add those things from my 2 comments ago first

gannonprudhomme:
Just added pylint_django as a plug in, which at least ignores all of the pylint errors for the Meta classes on the models

gannonprudhomme:
Turns out adding pylint_django to the rcfile didn't do anything. What did the trick was adding it as an argument to pylint in VSCode, which you do by adding the following to `settings.json`:
```json
	"python.linting.pylintArgs": [
		"--load-plugins",
		"pylint_django"
	]
```

gannonprudhomme:
Also I rebased these commits off of `backend/master`, hence the force push

firejake308:
The linting looks good when I manually run `pylint autoscheduler/scraper`, it runs as expected, with the "too few class methods" error gone, as you indicated.  I also got VSCode to run it on save, but I had to manually `python3 -m pip install pylint-django` before that worked, so you might want to update the Wiki page to include that.

Also, the missing module docstrings are getting annoying, can we manually disable those?

gannonprudhomme:
> but I had to manually `python3 -m pip install pylint-django` before that worked

Yeah I had to add it to requirements.txt, so you should've ran `pip install -r requirements.txt` instead, although yeah I should've mentioned that.

> Also, the missing module docstrings are getting annoying, can we manually disable those?

Yeah I agree they are getting annoying, I'll add them to be ignored in the rc file

gannonprudhomme:
Upgraded pylint to `2.4.3` which allowed the differentiation for disabling `missing-module-docstring` vs `missing-class-docstring` and `missing-function-docstring`. Then I added the disabling of `C0114`, which corresponds to `missing-module-docstring`.

Would be nice if we could get it to ignore docstrings for subclasses of `django.db.models.Model`, but that's probably asking for too much.

Also I rebased onto `backend/master` to make this up to date(which was a bit pre-emptive of me I admit), so to check it you'll need to run `git reset --hard origin/backend/setup-pylint` again.

gannonprudhomme:
@rconn478 Can you run this so we can be sure it works for all of us?

.pylintrc
```diff
@@ -0,0 +1,575 @@
...
+# Disable the message, report, category or checker with the given id(s). You
+# can either give multiple identifiers separated by comma (,) or put this
+# option multiple times (only on the command line, not in the configuration
+# file where it should appear only once). You can also use "--disable=all" to
+# disable everything first and then reenable specific checks. For example, if
+# you want to run only the similarities checker, you can use "--disable=all
+# --enable=similarities". If you want to run only the classes checker, but have
+# no Warning level messages displayed, use "--disable=all --enable=classes
+# --disable=W".
+disable=C0114, # missing-module-docstring
```
rconn478:
```suggestion
disable=missing-module-docstring,
```
Using `missing-module-docstring` does the same thing and is self-explanatory, unless you ran into a problem using it that I'm not thinking about

.pylintrc
```diff
@@ -0,0 +1,575 @@
...
+
```
rconn478: remove this line

.pylintrc
gannonprudhomme: Agreed, fixing

gannonprudhomme:
Added the change Ryan suggested. Going to squash some of these commits, but not all of them as some of them I think are valuable to have individually(such as upgrading pylint and adding pylint_django to requirements)

PR #35, opened 2019-10-27
Create Department Model

### LINKED ISSUE #24: Create the Department Model
See #10 

Feel free to assign yourself!

btgathright:
Closes #24

### DIFF
--- a/autoscheduler/scraper/models/department.py
+++ b/autoscheduler/scraper/models/department.py
@@ -1,4 +1,9 @@
 from django.db import models
 
 class Department(models.Model):
-    """ Describe model here """
+    """ Data structure that stores the department tag and description """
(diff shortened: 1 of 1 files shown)


autoscheduler/scraper/models/department.py
```diff
@@ -1,4 +1,9 @@
 from django.db import models
 
 class Department(models.Model):
-    """ Describe model here """
+    """ Data structure that stores the department tag and description """
```
gannonprudhomme: I know I said to include it previously, but I honestly don't think this needs a doc-string, as it's super explanatory. Previously I wanted to do it just cause pylint said to, but I think it's overkill in this case at the least. Thoughts @rconn478

gannonprudhomme:
Also changed it so that it would be merged into `backend/master` rather than `master`.

PR #36, opened 2019-10-27
Created models for section and meeting

### LINKED ISSUE #25: Create the Section & Meeting model
See #10

rconn478:
Closes #25.
This is generally what the meeting and section models should look like, but there are a couple things we should think about together:
- A lot of fields are IntegerFields (IDs, course & section numbers, etc.), but everything is stored in the data is a string. For some of these, I feel like storing them as IntegerFields will be beneficial (eg. sorting by course_num and section_num), but others (like IDs) are debatable.
- Delete behavior for section: Right now, on_delete is set to CASCADE for the section field in a Meeting, meaning if the section if belongs to is deleted, the meeting will as well. I feel like this is the correct behavior.
- ArrayFields: These seem like the easiest way to store meeting days, but can only be used with PostgreSQL. This shouldn't be a problem, but before finalizing this, I want everyone to be aware.
- There is a one-to-many relationship between sections and meetings, so I made section a foreign key for a meeting. I believe this is the best way to do this, since this is the practice Django recommends for one-to-many relationships. If anyone has a different idea, let me know.
- The same applies for the relationship between instructors and sections

### DIFF
--- a/autoscheduler/scraper/models/section.py
+++ b/autoscheduler/scraper/models/section.py
@@ -1,9 +1,42 @@
+"""Models for meetings and sections.
+Together, they represent all the information for a particular meeting time.
+"""
+
+from django.contrib.postgres.fields import ArrayField
 from django.db import models
 
-class Meeting(models.Model):
-    """ Describe model here """
-
 class Section(models.Model):
-    """ Section will contain Meeting objects,
-        which is why its placed below it
+    """ Section contains data for a group of meetings.
     """
+    id = models.IntegerField(primary_key=True) # id is primary key in scraped data
+    subject = models.CharField(max_length=4, db_index=True)
+    course_num = models.IntegerField(db_index=True)
+    section_num = models.IntegerField(db_index=True)
+    term_code = models.IntegerField(db_index=True)
+    min_credits = models.IntegerField()
+    max_credits = models.IntegerField(null=True) #min_credits is never null, but max_credits can be
+    max_enrollment = models.IntegerField()
+    current_enrollment = models.IntegerField()
+    instructor = models.IntegerField()
+
+    class Meta:
+        db_table = "sections"
+
+class Meeting(models.Model):
+    """Describes a particular meeting time of a section.
+    Each section has one or more meetings.
+    """
+    id = models.IntegerField(primary_key=True) # id is primary key in scraped data
+    crn = models.IntegerField(db_index=True) # We'll probably going to be searching for CRNs a lot
+    building = models.CharField(max_length=4, null=True) # Online/research meetings have no building
+    meeting_days = ArrayField(
+        models.BooleanField(),
+        size=7
+    )
+    start_time = models.TimeField(null=True)
+    end_time = models.TimeField(null=True)
+    meeting_type = models.CharField(max_length=3) #Meeting types: LEC, LAB, RES, INS
+    section = models.ForeignKey(Section, on_delete=models.CASCADE)
+
+    class Meta:
+        db_table = "meetings"
--- a/autoscheduler/scraper/models/section.py
+++ b/autoscheduler/scraper/models/section.py
@@ -7,15 +7,14 @@ from django.db import models
 from scraper.models import Instructor
 
 class Section(models.Model):
-    """ Section contains data for a group of meetings.
-    """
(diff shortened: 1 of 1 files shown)


rconn478:
One more related question: I'm not sure if it's good practice to push migration files, so I didn't include them. I found conflicting answers [here](https://stackoverflow.com/questions/28035119/should-i-be-adding-the-django-migration-files-in-the-gitignore-file), so we should think about if we should or not. I'm leaning towards not, since it just adds more file changes that could be acquired by running `makemigrations`, and unnecessary commits in the case that someone doesn't `makemigrations` before pushing.

rconn478:
Another side note: this pull request should definitely be squashed

autoscheduler/scraper/models/section.py
```diff
@@ -1,9 +1,43 @@
...
-    """ Describe model here """
+from django.contrib.postgres.fields import ArrayField
+from django.db import models
+from scraper.models import Instructor
 
 class Section(models.Model):
-    """ Section will contain Meeting objects,
-        which is why its placed below it
+    """ Section contains data for a group of meetings.
+    """
```
gannonprudhomme: Might not need this docstring

autoscheduler/scraper/models/section.py
```diff
@@ -1,9 +1,43 @@
...
+    """Describes a particular meeting time of a section.
+    Each section has one or more meetings.
     """
+    id = models.IntegerField(primary_key=True) # id is primary key in scraped data
+    crn = models.IntegerField(db_index=True) # We'll probably going to be searching for CRNs a lot
+    building = models.CharField(max_length=4, null=True) # Online/research meetings have no building
+    meeting_days = ArrayField(
+        models.BooleanField(),
+        size=7
+    )
```
gannonprudhomme: Change to one line?

autoscheduler/scraper/models/section.py
```diff
@@ -1,9 +1,43 @@
...
+from django.contrib.postgres.fields import ArrayField
+from django.db import models
+from scraper.models import Instructor
 
 class Section(models.Model):
-    """ Section will contain Meeting objects,
-        which is why its placed below it
+    """ Section contains data for a group of meetings.
+    """
+    id = models.IntegerField(primary_key=True) # id is primary key in scraped data
```
gannonprudhomme:
```suggestion
    id = models.BigIntegerField(primary_key=True) # id is primary key in scraped data
```


Should be a BigIntegerField, as the ID

autoscheduler/scraper/models/section.py
```diff
@@ -1,9 +1,43 @@
...
+    instructor = models.ForeignKey(Instructor, on_delete=models.CASCADE)
+
+    class Meta:
+        db_table = "sections"
+
+class Meeting(models.Model):
+    """Describes a particular meeting time of a section.
+    Each section has one or more meetings.
     """
+    id = models.IntegerField(primary_key=True) # id is primary key in scraped data
```
gannonprudhomme:
```suggestion
    id = models.BigIntegerField(primary_key=True) # id is primary key in scraped data
```

Needs to be a `BigIntegerField`, as  `41266-201931-0` converted to an Integer & removing the redundant `20` in `2019` is 4.1 billion

autoscheduler/scraper/models/section.py
rconn478: Agreed

autoscheduler/scraper/models/section.py
rconn478: We could also make it a CharField so we can just leave it with the hypens instead of having to manipulate the strings, it would definitely be faster when we're scraping and probably the same performance doing queries, but I'm not sure honestly

gannonprudhomme:
> One more related question: I'm not sure if it's good practice to push migration files, so I didn't include them. I found conflicting answers [here](https://stackoverflow.com/questions/28035119/should-i-be-adding-the-django-migration-files-in-the-gitignore-file), so we should think about if we should or not. I'm leaning towards not, since it just adds more file changes that could be acquired by running `makemigrations`, and unnecessary commits in the case that someone doesn't `makemigrations` before pushing.

I'll look more into it, but from what I know about it I would be fine with not including it

PR #40, opened 2019-10-28
Implement Course model

### LINKED ISSUE #22: Create the Course model
See #10

ThatJuanGuy:
Course model. Added some fields for core curriculum tags. Combined min/max credits into just credit hours. Left some comments as comments.

Closes #22 .

### DIFF
--- a/autoscheduler/scraper/models/course.py
+++ b/autoscheduler/scraper/models/course.py
@@ -8,11 +8,10 @@ class Course(models.Model):
     description = models.TextField(blank=True)
     prerequisites = models.TextField(blank=True)
     corequisites = models.TextField(blank=True)
-    cd = models.BooleanField()#states whether course is a cd credit
-    icd = models.BooleanField()#states whether course is an icd credit
-    core_curriculum =models.CharField(max_length=32,blank=True)#ex'Creative Arts'Blank for NA
-    credit_hours = models.FloatField(null = True)#Number of credit hours the course is worth
+    cd = models.BooleanField() # states whether course is a cd credit
+    icd = models.BooleanField() # states whether course is an icd credit
+    core_curriculum = models.CharField(max_length=32, blank=True) # ex 'Creative Arts'Blank for NA
+    credit_hours = models.FloatField(null=True) # Number of credit hours the course is worth
 
     class Meta:
         db_table = "courses"
-        
--- a/autoscheduler/scraper/models/course.py
+++ b/autoscheduler/scraper/models/course.py
@@ -1,18 +1,17 @@
 from django.db import models
 
-class Course(models.Model): # Directly from GBS:w
+class Course(models.Model): 
     id = models.CharField(max_length=10, primary_key=True) # Could be dept+course_num
     dept = models.CharField(max_length=4, db_index=True) # CSCE
     course_num = models.CharField(max_length=5, db_index=True) # i.e. 221
     title = models.CharField(max_length=100) # Course title, i.e. "Data Structres & Algorithms"
-    description = models.TextField(null=True, blank=True)
-    prerequisites = models.TextField(null=True, blank=True)
-    corequisites = models.TextField(null=True, blank=True)
-    cd = models.CharField(max_length=3)#yes or no. states whether course is a cd credit
-    icd = models.CharField(max_length=3)#yes or no. states whether course is an icd credit
-    core_curriculum = models.CharField(max_length=32)#tag specifying what it counts as ex) 'Creative Arts' or 'American History'. If not a core curriculum class it can be blank or 'NA' or something
-    credit_hours = models.FloatField(null = True)#Number of credit hours the course is worth. I didn't really see a point in the min/max credits because I've yet to see the same course with variable credit hours in the same semester (probably because I'm a freshman though).
-    # distribution_of_hours = models.CharField(max_length=100) No idea what this is supposed to be but i left it as a comment. Explanation apprecitated. If this is a list of times the course meets I feel it's kind of redundant putting it here because that information will be in section
+    description = models.TextField(blank=True)
+    prerequisites = models.TextField(blank=True)
+    corequisites = models.TextField(blank=True)
+    cd = models.BooleanField()#states whether course is a cd credit
+    icd = models.BooleanField()#states whether course is an icd credit
+    core_curriculum =models.CharField(max_length=32,blank=True)#ex'Creative Arts'Blank for NA
+    credit_hours = models.FloatField(null = True)#Number of credit hours the course is worth
 
     class Meta:
         db_table = "courses"
--- a/autoscheduler/scraper/models/course.py
+++ b/autoscheduler/scraper/models/course.py
@@ -1,4 +1,18 @@
 from django.db import models
 
-class Course(models.Model):
-    """ Describe model here """
+class Course(models.Model): # Directly from GBS:w
+    id = models.CharField(max_length=10, primary_key=True) # Could be dept+course_num
+    dept = models.CharField(max_length=4, db_index=True) # CSCE
+    course_num = models.CharField(max_length=5, db_index=True) # i.e. 221
+    title = models.CharField(max_length=100) # Course title, i.e. "Data Structres & Algorithms"
+    description = models.TextField(null=True, blank=True)
+    prerequisites = models.TextField(null=True, blank=True)
+    corequisites = models.TextField(null=True, blank=True)
+    cd = models.CharField(max_length=3)#yes or no. states whether course is a cd credit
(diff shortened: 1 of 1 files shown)


autoscheduler/scraper/models/course.py
```diff
@@ -1,4 +1,18 @@
...
+    description = models.TextField(null=True, blank=True)
+    prerequisites = models.TextField(null=True, blank=True)
+    corequisites = models.TextField(null=True, blank=True)
+    cd = models.CharField(max_length=3)#yes or no. states whether course is a cd credit
+    icd = models.CharField(max_length=3)#yes or no. states whether course is an icd credit
+    core_curriculum = models.CharField(max_length=32)#tag specifying what it counts as ex) 'Creative Arts' or 'American History'. If not a core curriculum class it can be blank or 'NA' or something
+    credit_hours = models.FloatField(null = True)#Number of credit hours the course is worth. I didn't really see a point in the min/max credits because I've yet to see the same course with variable credit hours in the same semester (probably because I'm a freshman though).
+    # distribution_of_hours = models.CharField(max_length=100) No idea what this is supposed to be but i left it as a comment. Explanation apprecitated. If this is a list of times the course meets I feel it's kind of redundant putting it here because that information will be in section
```
gannonprudhomme: These 3 lines are too long, seem like they're above the 90-character limit

autoscheduler/scraper/models/course.py
```diff
@@ -1,4 +1,18 @@
...
+    id = models.CharField(max_length=10, primary_key=True) # Could be dept+course_num
+    dept = models.CharField(max_length=4, db_index=True) # CSCE
+    course_num = models.CharField(max_length=5, db_index=True) # i.e. 221
+    title = models.CharField(max_length=100) # Course title, i.e. "Data Structres & Algorithms"
+    description = models.TextField(null=True, blank=True)
+    prerequisites = models.TextField(null=True, blank=True)
+    corequisites = models.TextField(null=True, blank=True)
+    cd = models.CharField(max_length=3)#yes or no. states whether course is a cd credit
+    icd = models.CharField(max_length=3)#yes or no. states whether course is an icd credit
+    core_curriculum = models.CharField(max_length=32)#tag specifying what it counts as ex) 'Creative Arts' or 'American History'. If not a core curriculum class it can be blank or 'NA' or something
```
gannonprudhomme: > If not a core curriculum class it can be blank or 'NA' or something

Since it's possible to be blank, this should have `null=True` to indicate that the field can be empty

autoscheduler/scraper/models/course.py
```diff
@@ -1,4 +1,18 @@
 from django.db import models
 
-class Course(models.Model):
-    """ Describe model here """
+class Course(models.Model): # Directly from GBS:w
```
gannonprudhomme: > # Directly from GBS:w

Did you copy and paste this directly from my model? This comment isn't needed(also includes `:w` from vim lmao)

autoscheduler/scraper/models/course.py
```diff
@@ -1,4 +1,18 @@
...
-class Course(models.Model):
-    """ Describe model here """
+class Course(models.Model): # Directly from GBS:w
+    id = models.CharField(max_length=10, primary_key=True) # Could be dept+course_num
+    dept = models.CharField(max_length=4, db_index=True) # CSCE
+    course_num = models.CharField(max_length=5, db_index=True) # i.e. 221
+    title = models.CharField(max_length=100) # Course title, i.e. "Data Structres & Algorithms"
+    description = models.TextField(null=True, blank=True)
+    prerequisites = models.TextField(null=True, blank=True)
+    corequisites = models.TextField(null=True, blank=True)
```
gannonprudhomme: Should these be able to be blank *and* null? Wouldn't it make sense for them to be either one or the other? 

According to [this article](https://simpleisbetterthancomplex.com/tips/2016/07/25/django-tip-8-blank-or-null.html), it's best for character related fields(`TextField` or `CharField`) to have just `blank=True`, since this means that these fields can only have one possible state for being empty. Otherwise there would be two, one for null, and another for being blank.

autoscheduler/scraper/models/course.py
```diff
@@ -1,4 +1,18 @@
...
+    cd = models.CharField(max_length=3)#yes or no. states whether course is a cd credit
+    icd = models.CharField(max_length=3)#yes or no. states whether course is an icd credit
```
gannonprudhomme: If these are "yes" or "no", shouldn't they instead be a [BooleanField](https://docs.djangoproject.com/en/2.2/ref/models/fields/#booleanfield)?

autoscheduler/scraper/models/course.py
```diff
@@ -1,4 +1,18 @@
...
+    core_curriculum = models.CharField(max_length=32)#tag specifying what it counts as ex) 'Creative Arts' or 'American History'. If not a core curriculum class it can be blank or 'NA' or something
+    credit_hours = models.FloatField(null = True)#Number of credit hours the course is worth. I didn't really see a point in the min/max credits because I've yet to see the same course with variable credit hours in the same semester (probably because I'm a freshman though).
```
gannonprudhomme: > I didn't really see a point in the min/max credits because I've yet to see the same course with variable credit hours in the same semester (probably because I'm a freshman though).

Just by the way, comments like these are much more appropriate to include in the pull request/code review as a comment on this line(like I'm doing now), rather than being an actual comment in the code - as they won't really pertain to anyone else in a few weeks.

Also idk if this helps, but I literally did this *exact* same thing at my first internship this summer for my first pull request/code review, so I 100% know where you're coming from 😃 . What I typically do if I want to remember a comment is have them in the file like you have it now(except on a separate line), then when I'm committing them or about to make a pull request, copy them to a text file then add them as a comment on the code review.

gannonprudhomme:
Ps, I added "Closes #22" to the end of your description, which should close the according issue once this pull request is merged.

autoscheduler/scraper/models/course.py
rconn478: Yeah, TextFields should have `blank=True`, and other fields should have `null=True`.

autoscheduler/scraper/models/course.py
ThatJuanGuy: Yes. I thought I was supposed to adapt the old one for this. I kept the first few lines with the course name and stuff because they seemed good and then just added the core curriculum stuff at the end because that was the only thing I could think of that needed to be added at course level.

autoscheduler/scraper/models/course.py
ThatJuanGuy: Was not aware of this. Will fix. Thanks for letting me know.

autoscheduler/scraper/models/course.py
ThatJuanGuy: Not necessarily but they can be. Considering there are only two options in this situation, that's a better idea. I shall make the change.

autoscheduler/scraper/models/course.py
ThatJuanGuy: I wasn't sure if you wanted to go with 'NA' or blank or something else in the case of a class not being part of the core curriculum, so I just left it as default with max length being the length of the longest classification. I shall add the null = true now that I know which one you want to go with.

autoscheduler/scraper/models/course.py
ThatJuanGuy: Ok. Thanks for the tip. Will do next time I make a pull request.

autoscheduler/scraper/models/course.py
ThatJuanGuy: Will fix. This is because that's where I put my comments regarding the code. Will do it in a separate text file from now on.

autoscheduler/scraper/models/course.py
gannonprudhomme: Yeah nothing wrong with using it as a base that's what I intended it for, just meant remove the stuff that isn't necessary, such as that comment :)

gannonprudhomme:
I added some formatting stuff that I thought would be kind of nit picky to tell you to change individually, really just spacing changes and fixing the end of line at the end of the file

gannonprudhomme:
Another nit pick: instead of having your commits be `Create course model` and adding a description to them, make them more descriptive such that if they were in the master branch you would be able to understand what the commit change just by reading the commit message.

For instance, a more proper name for [this commit](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/pull/40/commits/b2b7b00cfd60e07f6fb6919a2cc74628b0f03daf) would be `Added end of line to course model`.

gannonprudhomme:
Also just btw, whenever you merge the PR from a branch feel free to delete the branch

PR #42, opened 2019-10-29
Instructor model

### LINKED ISSUE #23: Create the Instructor Model
Will fill out specifics soon

firejake308:
Added instructor model. If we use the PIDM, I believe it should fit in IntegerField, name is really the only thing that's required, and email address is optional. Eventually, grade distributions will have to be stored somewhere, but I think that ought to be a separate table, since a professor might have a good grade distro for one class but not for another.

Closes #23

### DIFF
--- a/autoscheduler/scraper/models/instructor.py
+++ b/autoscheduler/scraper/models/instructor.py
@@ -3,7 +3,7 @@ from django.db import models
 class Instructor(models.Model):
     """ Model that represents a professor, who may teach many courses """
     id = models.IntegerField(primary_key=True)
-    email_address = models.CharField(max_length=320, null=True)
+    email_address = models.CharField(max_length=48, null=True)
     name = models.CharField(max_length=64)
 
     class Meta:
--- a/autoscheduler/scraper/models/instructor.py
+++ b/autoscheduler/scraper/models/instructor.py
@@ -1,4 +1,10 @@
 from django.db import models
 
 class Instructor(models.Model):
-    """ Describe model here """
+    """ Model that represents a professor, who may teach many courses """
+    id = models.IntegerField(primary_key=True)
+    email_address = models.CharField(max_length=320, null=True)
+    name = models.CharField(max_length=64)
+
+    class Meta:
+        db_name = "instructors"


firejake308:
Also I learned so much Git while making this one commit lol

gannonprudhomme:
> Also I learned so much Git while making this one commit lol

How so? I totally feel that though I've learned a crap load about git working on this

firejake308:
I had made an instructor.py file a long time ago on a local branch, then you made one in the "added scraper models" commit, so I had to do a lot of rebasing and squashing to make a coherent commit history

autoscheduler/scraper/models/instructor.py
```diff
@@ -1,4 +1,10 @@
 from django.db import models
 
 class Instructor(models.Model):
-    """ Describe model here """
+    """ Model that represents a professor, who may teach many courses """
+    id = models.IntegerField(primary_key=True)
+    email_address = models.CharField(max_length=320, null=True)
+    name = models.CharField(max_length=64)
```
gannonprudhomme: Do you think it's necessary to include their [bannerId](https://github.com/gannonprudhomme/AutoScheduler/blob/40fb364b8d6181839a5badfffcbf0ea426276853/autoscheduler/scraper/tests/section_inputs.json#L35) as well?

firejake308: No, we only need one or the other, as far as I can tell

gannonprudhomme: I searched checked if there were any duplicate entries and it seems like both of them are unique regardless, so it really shouldn't matter which we use. Given that integers are probably always going to be faster that strings for searching/querying, I think sticking with the pidm is fine.

autoscheduler/scraper/models/instructor.py
```diff
@@ -1,4 +1,10 @@
 from django.db import models
 
 class Instructor(models.Model):
-    """ Describe model here """
+    """ Model that represents a professor, who may teach many courses """
+    id = models.IntegerField(primary_key=True)
+    email_address = models.CharField(max_length=320, null=True)
```
gannonprudhomme: Max length can be around 48, querying all of the instructors for 2019(in all departments) didn't have anyone with an email longer than that

autoscheduler/scraper/models/instructor.py
gannonprudhomme: Fun thing I found out when searching for instructors longest names: the instructor with the longest names' name is: `Kalu Arachchillage Senarath Dhananjaya Bandara Dharmasena`

firejake308: 57 < 64

gannonprudhomme: Oh no yeah this line is fine, I just thought it was funny lmao

firejake308: Also, I think pablo Picasso's full name would also throw an error, but he's not a prof here, so I think 64 is a reasonable limit

PR #43, opened 2019-10-29
Frontend tests setup

### LINKED ISSUE #14: Setup React & Typescript testing framework
Need to determine what testing framework we'd like to use for the frontend (i.e. Mocha and Chai), and set it up for use in the project

Also possibly integrate it into VSCode so that they run on save?

firejake308:
Jest and React Testing Library have been setup for the frontend. @gannonprudhomme did most of the  work, I just added the first test and confirmed that it passes/fails properly.

Closes #14


gannonprudhomme:
Any idea how we could get these running automatically in VSCode? I know from looking previously that there's a jest extension. Alternatively, we could run the tests by adding it as a task in `tasks.json`, then add a keybinding to it in `keybindings.json` that would look like

```json
{ "key": "ctrl+b", "command": "workbench.action.tasks.test" }
```

gannonprudhomme:
Requesting a review from Ryan since I made a lot of the changes

gannonprudhomme:
Everything works fo me as well, merge whenever. May want to squash "Added jest to env to ignore undef for test, etc" into the prev commitas its a pretty un-descriptive commit message

gannonprudhomme:
Actually hold that, lets move `App.test.tsx` to a `tests/` folder in `src/`, as we're going to have a lot more tests

PR #45, opened 2019-10-29
Fixed instructor table property

gannonprudhomme:
Instructor's Meta class table property was `db_name`, but should have been `db_table`.


gannonprudhomme:
Imma just merge this as it's needed for it to work, so hopefully you'll get the notification as I really just wanted you to check this out so you could see what it should've been

PR #46, opened 2019-10-29
Setup python testing framework

gannonprudhomme:
Sets up python testing to allow for running of `django.test.TestCase` and `unittest.TestCase`(default Python) tests through the use of `pytest` and `pytest-django`. 

The main thing I wanted from this is the ability to run the tests through VSCode. While VSCode has a Python test explorer built in, there's an issue with it and pytest where running `pytest` in `autoscheduler` works fine, but running the tests through the default test explorer gives me a file not found error for some reason(even though it discovers the tests fine).

You can set up the default test explorer by doing `Ctrl+Shift+P` and selecting `Python: Configure Tests`, then choosing `pytest`, then choosing `autoscheduler` as the root folder. 

To get it working, I instead installed the [Python Test Explorer for Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=LittleFoxTeam.vscode-python-test-adapter). In a few `vscode-python` extension updates we shouldn't need this, as I think the fix will be included in a future one according to [this issue](https://github.com/microsoft/vscode-python/issues/6548) on the `vscode-python` repo.

### DIFF
--- a/autoscheduler/pytest.ini
+++ b/autoscheduler/pytest.ini
@@ -2,3 +2,4 @@
 DJANGO_SETTINGS_MODULE = autoscheduler.settings
 python_files = *_test.py *_tests.py # Any file ending with _tests.py
 testpaths = scraper/tests scheduler/tests
+addopts = -ra
--- a/autoscheduler/pytest.ini
+++ b/autoscheduler/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+DJANGO_SETTINGS_MODULE = autoscheduler.settings
+python_files = *_test.py *_tests.py # Any file ending with _tests.py
+testpaths = scraper/tests scheduler/tests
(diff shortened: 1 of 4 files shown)


gannonprudhomme:
Also, do you guys think it would be a good idea to attach Brandon & Carlos to this? I'd like them to eventually be involved in code review as it will deepen their understanding of the project as well as git. Plus to test this they'd need to set up testing, which will force them to get familiar with it

gannonprudhomme:
Also before I merge I'm going to remove the `django_tests` and `unit_tests` files, just including them for now so we can ensure `pytest` is working.

gannonprudhomme:
Oops apologies for the assign+unassign lmao. This is pretty straightforward, I just attached both of you so you'd both see it

gannonprudhomme:
Rebased off `backend/master` and fixed the merge conflicts

PR #51, opened 2019-10-31
Frontend/materialui setup

### LINKED ISSUE #50: Setup MaterialUI

rconn478:
Set up material-ui (core and icons) and put some very basic testing to make sure it works.
Closes #50

### DIFF
--- a/autoscheduler/frontend/src/App.tsx
+++ b/autoscheduler/frontend/src/App.tsx
@@ -1,19 +1,26 @@
 import * as React from 'react';
 import { Router, Link } from '@reach/router';
+import { AppBar, IconButton, Toolbar } from '@material-ui/core'
+import HomeIcon from '@material-ui/icons/Home'
 
 import Empty from './components/Empty';
 
 const App: React.SFC = function App() {
-  return (
-    <div>
-      {/* Once we add material-ui, the nav bar will probably replace this link */}
-      <Link to="/">Home</Link>
-      <Router>
-        {/* One component for each page/route goes in here */}
-        <Empty path="/" />
-      </Router>
-    </div>
-  );
+    return (
+        <div>
+            <AppBar position="static">
+                <Toolbar>
+                <IconButton edge="start" color="inherit">
+                    <HomeIcon />
+                </IconButton>
+                </Toolbar>
+            </AppBar>
+            <Router>
+                {/* One component for each page/route goes in here */}
+                <Empty path="/" />
+            </Router>
+        </div>
+    );
 };
 
 export default App;
--- a/autoscheduler/frontend/src/App.tsx
+++ b/autoscheduler/frontend/src/App.tsx
@@ -1,26 +1,28 @@
 import * as React from 'react';
 import { Router, Link } from '@reach/router';
-import { AppBar, IconButton, Toolbar } from '@material-ui/core'
-import HomeIcon from '@material-ui/icons/Home'
+import { AppBar, IconButton, Toolbar } from '@material-ui/core';
+import HomeIcon from '@material-ui/icons/Home';
 
 import Empty from './components/Empty';
 
 const App: React.SFC = function App() {
-    return (
-        <div>
-            <AppBar position="static">
-                <Toolbar>
-                <IconButton edge="start" color="inherit">
-                    <HomeIcon />
-                </IconButton>
-                </Toolbar>
-            </AppBar>
-            <Router>
-                {/* One component for each page/route goes in here */}
-                <Empty path="/" />
-            </Router>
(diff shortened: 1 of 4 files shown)


autoscheduler/frontend/src/App.tsx
```diff
@@ -1,13 +1,24 @@
 import * as React from 'react';
 import { Router, Link } from '@reach/router';
+import { AppBar, IconButton, Toolbar} from '@material-ui/core'
```
gannonprudhomme:
```suggestion
import { AppBar, IconButton, Toolbar } from '@material-ui/core'
```

autoscheduler/frontend/src/App.tsx
```diff
@@ -1,13 +1,24 @@
...
+/*import AppBar from '@material-ui/core/AppBar';
+import IconButton from '@material-ui/core/IconButton';
+import Toolbar from '@material-ui/core/Toolbar';
+*/
```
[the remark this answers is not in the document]
rconn478: Yes, it is: importing multiple exports adds ~200kB when compiled with `npm run build` over importing them separately, and I meant to make a comment about it: I think we should use babel to get around this, since it's a pretty significant difference.

rconn478:
> 
> 
> Not sure how I didn't notice it in Adel's original Frontend testing PR, but linting `App.test.tsx` gives me:
> 
> ```
>   1:1  error  '@testing-library/react' should be listed in the project's dependencies, not devDependencies     import/no-extraneous-dependencies
>   2:1  error  '@testing-library/jest-dom' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies
> ```
> 
> Which of course isn't correct, since those should definitely be in devDependencies.
> 
Right now, react is listed in both devDependencies and dependencies, and it should just be in dependencies, which is why the linter is complaining. Apparently jest needs to be added as a plugin for ESLint (we can do this by adding `jest: true,` to `env` in our eslintrc.
> Also, the indention for the test uses 4 spaces, not 2. I think it's fine for now as I don't think it's necessary to add a commit just to fix it per se (unless y'all do), as we can just fix it whenever we add a test to that file next.
It had 2 space indentation before, so I left it like that, but I agree we should change it at some point.

rconn478:
Rebased, everything should be fixed now

firejake308:
I added a commit to clean up the linting. As it turns out, the manual `npm run lint` command wasn't actually testing anything because it was only looking for `.js` files, so I had to tell it to lint TS files in `package.json`. I went ahead and fixed indentation at the same time. Also, I got rid of the warning about `no-unused-dependencies` by telling ESLint that test files are allowed to use dev dependencies.

gannonprudhomme:
> Apparently jest needs to be added as a plugin for ESLint (we should be able to do this by just adding jest: true, to env in our eslintrc)

I believe that's [already the case](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/blob/7360fb2d5c39c9da02203f12df9ae1612be75d95/autoscheduler/frontend/src/.eslintrc.js#L48)

autoscheduler/frontend/src/App.tsx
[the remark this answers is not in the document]
gannonprudhomme: Yeah that is pretty significant, should we make this a task? 

Also have you used babel before @firejake308 ? I thought it just converts ES6 syntax to ES5 for old browsers. Can webpack not accomplish this?

autoscheduler/frontend/src/package.json
```diff
@@ -7,7 +7,7 @@
     "test": "jest",
     "dev": "webpack --mode development ./index.tsx --output ../static/main.js",
     "build": "webpack --mode production ./index.tsx --output ../static/main.js",
-    "lint": "eslint ."
+    "lint": "eslint --ext .js,.jsx,.ts,.tsx ."
```
gannonprudhomme: While it definitely wasn't working for me previously, even with this new change I get no output when I run `npm run lint`. Any ideas?

firejake308: Try intentionally breaking a file, then running it again. Mine succeeds silently and only produces output if there's a problem

autoscheduler/frontend/src/App.tsx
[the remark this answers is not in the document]
firejake308: Yeah, I'm in favor of using Babel to reduce our bundle size

autoscheduler/frontend/src/package.json
gannonprudhomme: Yeah this ended up working, all good

Issue #38, opened 2019-10-28
Implement Banner Requests class


gannonprudhomme:
Need to implement the Banner Requests class so we can implement course & department scraping/parsing

gannonprudhomme:
Also should have tests for each of them

gannonprudhomme:
Need to finish #39 before I finalize this, as I'd like the BannerRequests to have according tests per the above comment

gannonprudhomme:
Been going back and forth between decisions for the design of it, so this'll take a bit longer than expected

PR #55, opened 2019-11-02
Implemented BannerRequests

gannonprudhomme:
Implements #38 

Includes asynchronous/concurrent scraping for retrieving all courses for a given department, as well as retrieving all departments for a given term.

I would like `get_departments` to eventually use `aiohttp.ClientSession` rather than use the `requests` library, but for some reason the same `self.depts_url` url retrieves the departments when using `requests`, but not `aiohttp.ClientSession` - instead, it returns the cookies that should be attaching to the session.

The limit for concurrent `ClientSession`s *should be* 100, but it seemed to create ~200 sessions on the previous repo, I just frequently got `Unclosed session` warnings. As such, if this ends up being a problem we'll need to change the design to use a blocking queue / bounded buffer hold a fix number of sessions that the processes have to share. 

Apologies for this being so long, I would have liked to split it up between get_departments first then get_courses & the rest. Feel free to take your time on it.

### DIFF
--- a/autoscheduler/scraper/models/section.py
+++ b/autoscheduler/scraper/models/section.py
@@ -13,8 +13,10 @@ class Section(models.Model):
     course_num = models.IntegerField(db_index=True)
     section_num = models.IntegerField(db_index=True)
     term_code = models.IntegerField(db_index=True)
-    min_credits = models.IntegerField()
-    max_credits = models.IntegerField(null=True) # min_credits is never null, but max_credits can be
+
+    min_credits = models.IntegerField() # Will never be null
+    max_credits = models.IntegerField(null=True) # Will be null in most cases
+
     max_enrollment = models.IntegerField()
     current_enrollment = models.IntegerField()
     instructor = models.ForeignKey(Instructor, on_delete=models.CASCADE)
@@ -23,16 +25,18 @@ class Section(models.Model):
         db_table = "sections"
 
 class Meeting(models.Model):
-    """Describes a particular meeting time of a section.
-    Each section has one or more meetings.
+    """ Describes a particular meeting time of a section.
+        Each section has one or more meetings.
     """
     id = models.BigIntegerField(primary_key=True) # id is primary key in scraped data
-    crn = models.IntegerField(db_index=True) # We'll probably going to be searching for CRNs a lot
-    building = models.CharField(max_length=4, null=True) # Online/research meetings have no building
+    crn = models.IntegerField(db_index=True)
+
+    building = models.CharField(max_length=4, null=True) 
     meeting_days = ArrayField(models.BooleanField(), size=7)
     start_time = models.TimeField(null=True)
     end_time = models.TimeField(null=True)
-    meeting_type = models.CharField(max_length=3) # Meeting types: LEC, LAB, RES, INS
+
+    meeting_type = models.CharField(max_length=3) # Meeting types: LEC, LAB, REC, INS, etc
     section = models.ForeignKey(Section, on_delete=models.CASCADE)
 
     class Meta:


autoscheduler/scraper/tests/banner_requests_tests.py
```diff
@@ -0,0 +1,118 @@
...
+
+    async def test_search_does_retrieve_all(self):
+        """ Tests that search retrieves all of the departments it was given """
+
+        # Arrange
+        request = BannerRequests()
+
+        term = "201931" # Fall 2019
+
+        depts = ["CSCE", "MATH", "ECEN"]
```
gannonprudhomme: If you change the amount of departments to be > 5, you'll eventually get a `Unclosed session` error. The whole point of the context manager(aka the `async with ClientSession()...` in `search()`) is to close the session automatically when the object would leave scope, so this shouldn't be happening. Thus I need to figure out a way to create multiple sessions without the sessions unsuccessfully closing

autoscheduler/scraper/tests/banner_requests_tests.py
```diff
@@ -0,0 +1,118 @@
...
+        try:
+            _ = get_term_code("9", Semester.fall, Location.college_station)
+
+            assert False # If it doesn't raise an error, then didn't pass
+        except ValueError:
+            assert True
+            return
+
+class BannerRequestsTests(AioTestCase):
+    """ Tests BannerRequests for functionality. These access the network for communicating
```
gannonprudhomme: I'd like to have not only more tests, but for the current ones to be more descriptive/intentional in their functionality. The tests should really be examples for how they need to function, rather than basically testing that they don't throw errors.

For instance, testing that the first result of Fall 2019's departments isn't a good example of how the functions being used, as that fact isn't really that pertinent to the purpose of the function

gannonprudhomme:
Also, I would like to change [indent-after-paren](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/blob/756dedeb64c544282ca01b1a2ec8ca60c1d6264a/.pylintrc#L17) back to 4, as it wants the inner two lines below to be only indented 2 rather than 4 spaces. You can run pylint to see what I mean

```python
        data = {
            'uniqueSessionId': session_id,
            'term': term,
        }
```

autoscheduler/scraper/banner_requests.py
```diff
@@ -0,0 +1,184 @@
...
+
+        url = self.depts_url.format(**data)
+
+        response = requests.get(url) # Blocking, stops the thread until retrieves response
+
+        depts = response.json() # Also blocking
+
+        return depts
+
+    async def search(self, depts: List[str], term: str, amount: int = 750):
```
rconn478: If my understanding is correct (that search will synchronously call `get_courses` with each department in depts), I don't see a situation where doing this will be faster than making a thread for each department individually, so this function should probably just take in a single department.

autoscheduler/scraper/banner_requests.py
```diff
@@ -0,0 +1,184 @@
...
+        data = []
+        async with asyncio.gather(*data) as resp:
+            data.append(resp)
+
+    Department retrieval example:
+        term = "201931"
+        var banner = BannerRequests(term)
+
+        amount = 100
+        depts = banner.get_departments(term, amount) # Amount is optional, defaults to 300
```
rconn478: 1. BannerRequests constructor doesn't take term code as an argument (right now it's passed as an argument to `search()`), but I think it probably should be.
2. I think we should abstract away using `asyncio` (and maybe even making BannerRequests objects), we could make a top-level function `async_search(term, depts)` function to do this, and do the same with departments to avoid having to make a BannerRequests object.

autoscheduler/scraper/banner_requests.py
```diff
@@ -0,0 +1,184 @@
...
+        data = {
+            'uniqueSessionId': session_id,
+            'term': term,
+        }
+
+        await session.post(self.create_session_url, data=data)
+
+        return session
+
+    async def get_courses(self, session: ClientSession, session_id: str, dept: str,
```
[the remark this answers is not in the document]
gannonprudhomme: Nah that's completely fine to mention you good. GitHub Actions should be able to catch these in the future

autoscheduler/scraper/banner_requests.py
gannonprudhomme: > 1. BannerRequests constructor doesn't take term code as an argument (right now it's passed as an argument to search()), but I think it probably should.

Yeah I agree. I was going back and forth about it, but if they want to change the search to another term they can just change it by doing `banner.term_code = newterm`

> 2. I think we should abstract away using asyncio (and maybe even making BannerRequests objects), we could make a top-level function async_search(term, depts) function to do this, and do the same with departments to avoid having to make a BannerRequests object.

Yeah honestly the Course Search example is incorrect, as with the `search` function all you have to do is.

```python
data = banner.search(depts, term)
```

Forgot to update it after I changed the designs, will now.

autoscheduler/scraper/banner_requests.py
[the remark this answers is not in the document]
gannonprudhomme: Fixed

autoscheduler/scraper/banner_requests.py
ryan-conn: > Yeah honestly the Course Search example is incorrect, as that's the whole point of the `search` function. This is what the example should look like:
> 
> ```python
> data = await banner.search(depts, term)
> ```
> 
> Forgot to update it after I changed the designs, will now.

Since we're going to have to use these functions inside of our scraping commands, this won't work as it is now. When we start doing scraping, this how we can run search:
```
#in scrape-courses.py (function to synchronously run async):
def run_async(function, args):
    loop = asyncio.get_event_loop()
    res = loop.run_until_complete(asyncio.wait([asyncio.ensure_future(function(args))]))
    return res
```
And in `handle(self, *args, **options)`, put the following to get the data:
`data = run_async(test, depts)`
This shouldn't be included in this pull request (I'll implement it when I do the `scrape-courses` outline), but I'm putting this here just as a note.

autoscheduler/scraper/banner_requests.py
```diff
@@ -0,0 +1,187 @@
...
+        async with session.get(url) as response:
+            json = await response.json()
+
+        await self.reset_search(session)
+
+        data = json['data']
+
+        return data
+
+    async def get_departments(self, amount: int = 300) -> List[Dict]:
```
ryan-conn:
```suggestion
    def get_departments(self, amount: int = 300) -> List[Dict]:
```
Should be made synchronous

autoscheduler/scraper/banner_requests.py
gannonprudhomme: > Since we're going to have to use these functions inside of our scraping commands, this won't work as it is now. When we start doing scraping, this how we can run search:

Why exactly won't this work? Is it b/c the actual `scrape_courses` will be synchronous?

What's the need for `tasks = [asyncio.ensure_future(function(args))]`? Why couldn't we just do `loop.run_until_complete(function(args))`? That's how I'm doing it in the auto scheduler on my repo and it seems to work fine.

Also since I completely forgot about `loop.run_until_complete()`, should I transition the rest of the tests(and remove aio_test_case) to use this instead, rather than calling `await`? I personally think the `await` syntax is cleaner in terms of the tests.

ryan-conn:
@gannonprudhomme nerd

autoscheduler/scraper/banner_requests.py
gannonprudhomme: Yeah this was a mistake, `requests.get()` is a blocking call so there's no need for `async`. I originally had it there as I planned on using `ClientSession`, but it wasn't necessary. Fixing

gannonprudhomme:
Undid the commit that updated the module docstring, removed `async` from `get_departments`

autoscheduler/scraper/banner_requests.py
```diff
@@ -0,0 +1,189 @@
...
+
+    Course retrieval example:
+        var banner = BannerRequests("201931")
+
+        depts = ["CSCE", "MATH", "CHEM"]
+        requests = [banner.search(dept) for dept in depts]
+
+        data = []
+        async with asyncio.gather(*data) as resp:
+            data.append(resp)
```
gannonprudhomme: Could we change this to 

```python
loop = asyncio.get_running_loop()
data = loop.run_until_complete(banner.search(depts))
```

This is how I have it in the old AutoScheduler repo and it runs fine

.pylintrc
```diff
@@ -102,7 +102,7 @@ ignore-patterns=
 
 # Python code to execute, usually for sys.path manipulation such as
 # pygtk.require().
-#init-hook=
+init-hook='import sys; sys.path.append("autoscheduler")
```
gannonprudhomme: Of course now I can't figure out how to replicate the issue, so try to remove this line and run pylint in `scraper/`. I believe it was causing an error for `from scraper.banner_requests import Semester, ...` in `banner_requests_tests.py`, but now its not throwing one for me.

gannonprudhomme: Scratch that, the actual issue was trying to import `from scraper.banner_requests import BannerRequests` in `scraper_courses.py`, which I succeeded in replicating. so I do believe this is necessary.

gannonprudhomme:
Squashed some of the commits together

gannonprudhomme:
Oops messed up somehow lmao

gannonprudhomme:
I consider this a casualty of war

Issue #54, opened 2019-11-02
Implement MeetingCard


firejake308:
Hello, my frontend friends! I have written some tests for your first React component, `<MeetingCard />` and created a skeleton file for you to fill. Basically, I want you to learn about functional components in React, then make one that creates a MeetingCard. The component takes in a meeting as its props and it should render a card-like shape that has the subject and course number of that meeting. Once you've written your component, you can see the result for yourself by replacing `<Empty />` with `<MeetingCard meeting={insert dummy meeting variable here} />` in `App.tsx`, and if it looks good to you, run `npm run lint` to clean up your code, then `npm run test` to see if it works. If the tests pass, go ahead and commit it, and i'll check over your code. We'll be working on the `frontend/meeting` branch, so to get started, just assign yourself in GitHub and drag this issue into the Todo section in ZenHub. GLHF!

gannonprudhomme:
If this is already implemented for them to work on and assumingly doesn't break anything, I you might as well make a PR to merge `frontend/meeting` to `frontend/master`, that way we have less to review when they do the PR with their changes.

firejake308:
Okay, I'll make that PR after we go over `types.ts`

gannonprudhomme:
See [Material-UI](https://material-ui.com/) for what components we have to work with.

firejake308:
Good news, everyone! I discovered that if you extend `React.PureComponent`, you can write a stateless class component. This will hopefully make it easier for @scott-carrion because the class component syntax is more similar to the OOP that most people have experience with. If you have any questions, Scott, feel free to message me on Slack or comment on this issue.

firejake308:
By the way, I implemented the changes y'all requested in #57 and merged the tests into your branch for this

firejake308:
The meeting card has been merged into `frontend/schedule`. Marking as closed.

Issue #47, opened 2019-10-30
Outline graphic schedule components


gannonprudhomme:
Would incorporate the outlines for each of the basic components, i.e. the schedule table with all of the lines, the days bar at the top, an individual Meeting display, etc.

This would allow us to better divide the work on the schedule, so one person doesn't have to do all of it. Plus just to track progress on it better.

Also might want to outline Redux? Or at least give them a reference to it like #19 / the `frontend/redux-setup` branch

firejake308:
I think the overall design process should look like:
1. Get components that look and behave the way we want them to, populated with dummy data.
2. Connect components to redux, and populate the redux store with dummy data
3. Connect redux to backend

That way we can focus on one thing at a time

firejake308:
Setting up types would be a good first task, just to get used to TypeScript classes, but Redux itself is kinda complicated imo. I would have them learn React first, since Redux's action system is essentially a more complicated version of React's `setState`. So fill out the types first, then basic schedule, then Redux

firejake308:
Also, in my opinion, the only components should be the `<Calendar />` and `<Meeting />`, since no other parts of `Calendar` actually have any interactivity, and thus it is difficult to justify having reactive components for something like the header or the time bar on the side. `Meeting` is a pretty good first component since it takes in props and renders them, but `Calendar` is a little more involved, so that should be done later. I think one possible strategy that would be good for learning is if I write the tests for each component and the other members write components that pass those tests.

gannonprudhomme:
> Also, in my opinion, the only components should be the <Calendar /> and <Meeting />, since no other parts of Calendar actually have any interactivity, and thus it is difficult to justify having reactive components for something like the header or the time bar on the side

I definitely agree

> I think one possible strategy that would be good for learning is if I write the tests for each component and the other members write components that pass those tests.

That's a great idea I'm definitely down for you to do that.

Another thing, should this also include adding the MaterialUI components to `package.json` and making sure they work? Mainly b/c I had a bit of issue getting them set up on the previous repo somehow

rconn478:
> Another thing, should this also include adding the MaterialUI components to `package.json` and making sure they work? Mainly b/c I had a bit of issue getting them set up on the previous repo somehow

I think we should make an issue for that, I could try setting it up tomorrow morning

firejake308:
@rconn478 Sounds good, I'll try to write the tests tomorrow night after you set up Material-UI

[...]

PR #57, opened 2019-11-03
Meeting Card Skeleton

firejake308:
I have written a skeleton for the meeting card and appropriate tests. The tests are pretty simple, as the `MeetingCard` is a pretty simple component, and although I am aware that certain features will have to be added later, such as coloring and positioning, I would like to keep it simple enough to be someone's first React component, if they've never seen React before. I am aware that class components are generally more intuitive, but ESLint demands that stateless components be functional, and it isn't too hard in my opinion to get used to the idea of calling a function with certain arguments that returns a HTML element.


autoscheduler/frontend/src/types.ts
```diff
@@ -1,3 +1,32 @@
...
+export interface Section {
+  id: number;
+  subject: string;
+  courseNum: number;
+  sectionNum: number;
+  minCredits: number;
+  maxCredits: number | null;
+  currentEnrollment: number;
+  instructor: number;
+}
```
gannonprudhomme: Since this is basically the backbone of the front-end, I think this deserves its own PR so we can discuss what all we actually need.

firejake308: Do you want all of the types from `scraper/models.py` implemented? I only made the two that I needed to get the Meeting testing working, but it's trivial to copy the types from the backend branch if you want to get it all done at once

gannonprudhomme: . Only downside is there may be things we can't forsee that we'll need when actually using them, but I think it'll be fine if we change them in the future

gannonprudhomme: I don't expect it to be a 1:1 to the models(i.e. the department probably won't need its term)

gannonprudhomme:
Attached to #47 as there isn't an issue for this, and there really won't be any other outline for the schedule components(as I assume Adel will make the schedule component on his own)

gannonprudhomme:
Can this be remade(or rebased) with the up to date changes that are now on `frontend/master` from #58 ?

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -45,5 +45,6 @@ module.exports = {
   rules: {
     "react/jsx-filename-extension": [1, {"extensions": [".jsx", ".tsx"]}],
     "import/no-extraneous-dependencies": ["error", {"devDependencies": ["**/*.test.ts", "**/*.test.tsx"]}],
+    "react/prop-types": 0
```
gannonprudhomme: What does this rule specify? Is this related to #53 ?

firejake308: Yes.

autoscheduler/frontend/src/tests/Meeting.test.tsx
```diff
@@ -0,0 +1,64 @@
+import { render } from '@testing-library/react';
+import '@testing-library/jest-dom/extend-expect';
+import * as React from 'react';
+
+import MeetingCard from '../components/MeetingCard';
+import { Meeting, MeetingType, Section } from '../types';
```
gannonprudhomme: Does this not throw an error for you? I get a `Cannot find module '../types'` error when I try to run the tests.

autoscheduler/frontend/src/components/MeetingCard.tsx
```diff
@@ -0,0 +1,8 @@
+/* eslint-disable react/prop-types */
```
gannonprudhomme: Can we remove this line, per your change in `.eslintrc.js`

autoscheduler/frontend/src/tests/Meeting.test.tsx
firejake308: It does, I'll have to refactor it to reflect the latest commits to frontend/master

autoscheduler/frontend/src/components/MeetingCard.tsx
firejake308: Yes

autoscheduler/frontend/src/tests/Meeting.test.tsx
```diff
@@ -0,0 +1,64 @@
...
+
+test('displays subject and course number', () => {
+  const { getByText } = render(<MeetingCard meeting={testMeeting} />);
+  expect(getByText(/CSCE/)).toBeTruthy();
+  expect(getByText(/121/)).toBeTruthy();
+});
+
+test('subject and course number can adapt to random values', () => {
+  // arrange
+  const randSubject = new Array(5).reduce((acc) => acc + String.fromCharCode(
```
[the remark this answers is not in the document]
gannonprudhomme: > I'm not sure I like using random values for unit testing, unless we give a seed to the random function (we'd have to use seedrandom in npm or something similar), since it makes tests unpredictable. There isn't a reason that I can think of that using random values would change anything versus just giving it values we made manually, either, since it's just a static length string and a number 100-599

Yeah I agree, there really isn't a logical reason to have it be a random value, since there isn't logic related to the actual value of the subject, as long as it's a certain length. See [Unit Testing Standards](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/wiki/Unit-Testing-Standards), specifically the "Readable" section.

gannonprudhomme:
Given our discussion @firejake308 (which I think you were at my place for @ryan-conn ) about displaying the meeting type on the MeetingCard, should there be a test for it? Or not since we haven't determined exactly how it should be displayed.

ryan-conn:
I'll write some tests to make sure it displays correctly while I'm making the actual MeetingCard component.

gannonprudhomme:
> I'll write some tests to make sure it displays correctly while I'm making the actual MeetingCard component.

In that case, should we even merge this into `frontend/master`? You(ryan) could just use it when you're working on it.

That being said, the benefit of merging it to master is that there's less changes to look at when Ryan makes a PR for #54

ryan-conn:
> In that case, should we even merge this into `frontend/master`? You(ryan) could just use it when you're working on it.
> 
> That being said, the benefit of merging it to master is that there's less changes to look at when Ryan makes a PR for #54

Yeah I don't think this should get merged, I think @firejake308 should push `Meeting.test.tsx` (rename to `MeetingCard.test.tsx` and move to `frontend/src/tests/components`) as well as the `eslintrc` change to the `frontend/meeting-card` branch I made a little bit ago and we could just make one pull request for all the changes.

PR #58, opened 2019-11-05
Added TypeScript Types

firejake308:
This PR currently has the Meeting and Section type definitions for the frontend, which are the bare minimum to implement `<MeetingCard />`. If desired, the remaining types can be added, and the interfaces can be converted to classes with custom constructors to offer runtime type-checking.


autoscheduler/frontend/src/types.ts
```diff
@@ -1,3 +1,48 @@
+/* eslint-disable import/prefer-default-export */
+
 /**
  * This file is where we will declare the types of each data-holding class we will be using.
  */
+export enum MeetingType {
+  LEC, LAB, RES, INS
```
gannonprudhomme: Should include all of the possible meeting types, see: [Meeting Types](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/wiki/Meeting-Types)

autoscheduler/frontend/src/types.ts
```diff
@@ -1,3 +1,48 @@
...
+  instructor: number;
+}
+
+export interface Course {
+  id: string;
+  subject: string;
+  courseNum: number;
+  title: string;
+  description: string;
+  coreCurriculum: string;
```
gannonprudhomme: Will we need `coreCurriculum`? We don't have any plans for incorporating it at the moment

autoscheduler/frontend/src/types.ts
```diff
@@ -1,3 +1,48 @@
...
+
+export interface Section {
+  id: number;
+  subject: string;
+  courseNum: number;
+  sectionNum: number;
+  minCredits: number;
+  maxCredits: number | null;
+  currentEnrollment: number;
+  instructor: number;
```
gannonprudhomme: Shouldn't we have a reference to an `Instructor` type instead?

autoscheduler/frontend/src/types.ts
```diff
@@ -1,3 +1,48 @@
...
+
+export interface Meeting {
+  id: number;
+  crn: number;
+  building: string | null;
+  meetingDays: boolean[];
+  startTimeHours: number;
+  startTimeMinutes: number;
+  endTimeHours: number;
+  endTimeMinutes: number;
```
gannonprudhomme: Is there any benefit to storing these as separate values rather than two `date` objects(or rather, only the time within the date object)?

firejake308: The JS `Date` object will have some extraneous data for the year, month, date, seconds, and milliseconds, all of which will be filled with BS data. The space differences for these approaches are negligible, so whichever syntax produces the cleaner syntax should be preferred. I chose to prioritize clearly defining which data we are using over minimizing the number of fields. If this leads to messy syntax in your opinion, we can change it.

firejake308:
Added requested changes and runtime type-checking. Tests coming soon.

firejake308:
I've added the first suite of tests, and I just wanted one of y'all to look over it before I write the rest

autoscheduler/frontend/src/tests/Instructor.test.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
```
gannonprudhomme: Do we want this rule at all? From the related TSLint issue from [the typescript-eslint docs](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-var-requires.md):
```
ES2015-style imports are part of the JavaScript language specfication and recommended as the path going forward
```

gannonprudhomme:
Also, how come `Instructor.test.js` is a `.js` file and not a `.ts` file?

autoscheduler/frontend/src/tests/Instructor.test.js
```diff
@@ -0,0 +1,44 @@
...
+nonNullableProps.forEach((prop) => {
+  test(`Instructor rejects null ${prop}`, () => {
+    // arrange
+    const badArgs = { ...correctArgs };
+    badArgs[prop] = null;
+
+    // act and assert
+    expect(() => createInstructor(badArgs)).toThrow();
+  });
+});
```
gannonprudhomme: I think jest's [test.each](https://jestjs.io/docs/en/api#1-testeachtable-name-fn-timeout) would be cleaner

autoscheduler/frontend/src/types.ts
```diff
@@ -1,3 +1,132 @@
+/* eslint-disable max-classes-per-file */
```
gannonprudhomme: Honestly, I think these classes should be separated into different files, especially since they have their own logic in them now(with `Meeting` and `MeetingType` being in one).

While I do think `max-classes-per-file`'s default of 1 is a bit too strict(which I propose to change to 2), I do think it's correct in indicating that this file is falling outside of the guideline for giving a file a single responsibility.

autoscheduler/frontend/src/tests/Instructor.test.js
firejake308: Agreed. Fixing on next commit.

autoscheduler/frontend/src/types.ts
firejake308: 1 class per file should be fine, since `MeetingType` is an enum and doesn't count as a class

autoscheduler/frontend/src/tests/Instructor.test.js
firejake308: TypeScript won't let me use the bracket notation that lets me iterate through the object's non-nullable properties, so I just switched it to JavaScript. Hacky, I admit, but it's better than writing each test individually. Consequently, whatever compiler/engine Jest uses doesn't support ES2015-style imports in plain JS files, so I switched to the `require` syntax.

I looked into it more, and there is a workaround that I will implement in the next commit, but just a heads up, it requires disabling the "no-explicit-any" rule

firejake308:
Okay, all done. Only thing to note is that all arguments are required, and if no data is available, `null` should be specified when instantiating one of these types.

autoscheduler/frontend/src/types/Course.ts
```diff
@@ -0,0 +1,25 @@
...
+  constructor(src: {
+      id: string;
+      subject: string;
+      courseNum: number;
+      title: string;
+      description: string | null;
+      creditHours: number | null;
+    }) {
+    if (src.id == null) { throw Error(`Course.id is invalid: ${src.id}`); }
+    if (src.subject == null) { throw Error(`Course.id is invalid: ${src.subject}`); }
```
ryan-conn:
```suggestion
    if (src.subject == null) { throw Error(`Course.subject is invalid: ${src.subject}`); }
```

autoscheduler/frontend/src/types/Course.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (!Number.isInteger(src.courseNum)) {
+      throw Error(`Course.courseNum is invalid: ${src.courseNum}`);
+    }
+    if (src.title == null) { throw Error(`Course.title is invalid: ${src.title}`); }
+    if (src.description === undefined) {
+      throw Error(`Course.courseNum is invalid: ${src.description}`);
```
ryan-conn:
```suggestion
      throw Error(`Course.description is invalid: ${src.description}`);
```

PR #67, opened 2019-11-09
Implemented department scraping #41

### LINKED ISSUE #41: Implement Department Scraping
Need to implement department scraping, which would include scraping all of the departments for the university.

I (Gannon) will implement #38 shortly, which will provide a `get_departments` function that will return a JSON/dictionary with all of the departments for a section.

The departments change each semester. For instance, there are 2 depts in fall 2019 that aren't in fall 2018 (AREN, EREN), and 6 depts in fall 2018 not in fall 2019. As such, we need to scrape all of the departments from a certain term to the next term(spring 2020?)

This is an example of what you'll retrieve as input. The rest that needs to be done is parsing this and saving it as a Django model
```json
[
    {
        "code": "ACCT",
        "description": "ACCT - Accounting"
    },
    {
        "code": "AEGD",
        "description": "AEGD - Adv Ed Gen Dentistry"
    },
]
```

btgathright:
Closes #41

### DIFF
--- a/autoscheduler/scraper/management/commands/scrape_depts.py
+++ b/autoscheduler/scraper/management/commands/scrape_depts.py
@@ -1,0 +1,71 @@
+# Called to execute the beginning of scaping courses
+
+from django.core.management import base
+from scraper.models import Department
+
+import requests
+import json
+import time
+import datetime
+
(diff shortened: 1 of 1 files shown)


ryan-conn:
Please use the `backend/scraper/department-scraping` branch as a base for this instead of using the code from the old repository, since it uses banner requests instead of get requests. I'm going to close this pull request because we can't use it, but please let me know if you have any questions.

PR #69, opened 2019-11-13
Hot reload for npm run dev

### LINKED ISSUE #16: Setup VSCode to run webpack on file save

firejake308:
I was working on the Schedule, and having to type `npm run dev` over and over got annoying, so I added the HotModuleReplacement plugin for webpack. Now, you type `npm run dev` once, and when you make a change, webpack will rebuild main.js. It'll take a second to build, but you should be able to reload the page and see your changes without having to enter any additional commands after you save.

Obligatory `"Nice."` for the PR number.

Closes #16


firejake308:
Will be squashing all of these commits btw

firejake308:
I went back to Live Reload, which rebuilds the entire project, instead of Hot Module Replacement, which theoretically reloads only the modules that have changed because the reload times were about the same, and HMR made a bunch of hot-update.js files in static that would be annoying to delete every time.

autoscheduler/frontend/src/package.json
```diff
@@ -37,7 +37,8 @@
     "typescript": "^3.6.4",
     "typescript-eslint": "0.0.1-alpha.0",
     "webpack": "^4.41.0",
-    "webpack-cli": "^3.3.9"
+    "webpack-cli": "^3.3.9",
+    "webpack-livereload-plugin": "^2.2.0"
```
gannonprudhomme: Since you updated `package.json`, shouldn't there be an according `package-lock.json` change with it?

firejake308: Apparently, it's [supposed to be included](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5), but for some reason, it didn't get added with `git commit -am`. I wonder if git stopped tracking changes because it got too long and we need to add it manually or something.

firejake308:
Strangely enough, I ran `git add package-lock.json` and then `git status` and it said, "Nothing to commit, working tree clean." If you can run it without an updated package-lock, I think we can just go ahead and merge this and continue leaving off the package-lock

gannonprudhomme:
So we don't want to include `package-lock.json` going forward? If so, we should remove it from being tracked and add it to the `.gitignore`

ryan-conn:
> 
> 
> So we don't want to include `package-lock.json` going forward? If so, we should remove it from being tracked and add it to the `.gitignore`

I'm fine with us doing this

firejake308:
Y'all I tink I figured it out. My WSL has npm version 3.5.2 installed, but package locks were added in npm 5. That's why `npm install` wasn't writing to my package lock. Nice catch, @gannonprudhomme! In that case, we'll keep the package-lock, because it should theoretically help us avoid version conflicts later, and I'll commit the fixed lock.

firejake308:
Also, the fact that npm can install itself is as crazy to me as the fact that C compilers are written in older versions of C. The level of recursiveness  is up there.

firejake308:
One hour of `sudo` and `chown` later, I've finished the commit. I also took the liberty to fix some other things to get it to pass all GitHub Actions for frontend. Re-requesting review.

autoscheduler/frontend/src/package.json
```diff
@@ -20,6 +21,7 @@
     "@types/react-dom": "^16.9.1",
     "@typescript-eslint/eslint-plugin": "^2.4.0",
     "@typescript-eslint/parser": "^2.4.0",
+    "continuable-cache": "^0.3.1",
```
gannonprudhomme: Since `webpack-livereload-plugin` eventually ends up depending on `continuable-cache` (through `tiny-rl -> body -> continuable-cache`) which it has specified in the `package-lock.json`, shouldn't we not need to need to manually include this (assuming you did it to pass [this build failure](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/runs/305750606#step:5:8))?

firejake308: Theoretically, yes, that's how it should work, and in practice, it does work that way on my laptop, but not on GitHub Actions. I tried to see if I could replicate the problem in my own Docker container, but I got an `invalid volume specification .../postgres_username:/run/secrets/SECRET_POSTGRES_USER:ro` error. So at this point, it looks like we either just manually add it as a top-level dependency or you can try to run it without the top-level dependency in a clean Docker container and try to debug it yourself. My current theory is that Docker is using an older version of npm that ignores the package-lock and only looks at `package.json`

gannonprudhomme: So the Docker files we have inside of the repo are for the `Backend Testing` step, and aren't related to the other ones at all(which is why the error is related to Postgres). 

The step downloads Node(version 10.x) just runs `npm install`, so I'm really not sure why it's not correct. Likewise, I made a `frontend/live-reload-actionstest` branch just for messing with this and run [npm version](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/runs/306324728#step:5:1) on it, which has an npm version of 6.11.3, which is definitely up to date.

On that branch I reset to the `Merge branch 'frontend/master' into frontend/live-reload` commit before you added the `continuable-cache` package, ran `npm install` and committed the updated `package-lock.json` and it [passed the build](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/runs/306324728#step:6:1), so I think updating the `package-lock.json`, rather than specifying the `continuable-cache` package directly should make it work.

firejake308: Oof, I was overthinking it. Fixed by removing the unnecessary dependency.

autoscheduler/frontend/src/package.json
```diff
@@ -48,7 +46,6 @@
     "@reach/router": "^1.2.1",
     "@types/reach__router": "^1.2.6",
     "@types/react-redux": "^7.1.5",
-    "prop-types": "^15.7.2",
```
gannonprudhomme: Do we not have a need for `prop-types` anymore? Is it just needed by router as a dependency? I know you removed `react, react-dom`, and `prop-types` from `devDependencies` to silence the `included as both a dev and production dependency` warnings, but this is removing `prop-types` from both.

firejake308: Yeah, it was intentional, since we aren't really using prop-types as a top-level dependency in either development or production.

autoscheduler/frontend/src/package-lock.json
```diff
@@ -7,7 +7,7 @@
     "@babel/code-frame": {
       "version": "7.5.5",
       "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.5.5.tgz",
-      "integrity": "sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==",
+      "integrity": "sha1-vAeC9tafe31JUxIZaZuYj2aaj50=",
```
gannonprudhomme: This looks like it's reverting back to the old type of `npm` to me. When I run `npm install`, it changes the `integrity` to look like the line that was replaced.

Also, can you run `npm audit fix` and commit the changes? When I ran it locally it fixed the vulnerability just using that command. Over than that this should be good.

firejake308: Yeah, I noticed the messed-up integrities over in `frontend/css-modules`, so I'll just go ahead and apply the fix from that branch here before we merge into `frontend/master`. The end goal is to choose sha512 over sha1 whenever there's a conflict, and what I'm doing is `rm -rf node_modules && rm package-lock.json` followed by a fresh `npm install`. It worked in the other branch, but be forewarned, many packages will be updated in this process.

firejake308: Incidentally, Jest was also updated in this process, and `npm audit` now shows no issues

ryan-conn:
Hey I don't know if you realized but this branch has a few commits duplicated from `frontend/master`

firejake308:
@ryan-conn Yeah, I know, I had to merge frontend/master into frontend/live-reload in order to merge frontend/live-reload into frontend/master. I'm hoping it will be fine if I just squash and merge though. What do you think?

ryan-conn:
I talked about it with Gannon, I don't think that would cause any issues but we'd prefer if you could revert to before the commits from frontend/master and force push over them just to be safe

firejake308:
This is gonna be the best I can do.

firejake308:
Well, I figured out that I could have made this cleaner using `git rebase frontend/master` instead of `git merge frontend/master`. I don't think it's worth re-opening at this point, but I used rebase to fix the merge conflicts in the `frontend/css-modules` PR

PR #72, opened 2019-11-14
Added GitHub Actions

gannonprudhomme:
#### Added a GitHub Actions workflow.  This includes:
- Running `webpack` (`npm run build`)
- Running `jest` for the frontend (`npm test`)
  - Currently failing since jest isn't set up on the `master` branch, but I confirmed it works in the `frontend/github-actions` temp branch I made. 
- Running `pytest` for the backend
- Running `pylint`
  - Currently failing since there are so many errors in the files that I don't want to fix in this PR
  - Also, a lot of the failures are `missing-module-docstring`, which we have disabled in `backend/master`.
- Running `eslint`

#### There's still a few things I would like to implement(or am considering):
- [Danger.js](https://danger.systems/js/). Had a bit of trouble getting this set up, so I'll probably do it in a separate PR. Danger is useful for various things, such as:
  - Adding a warning/error PR comment if a PR is over 500 lines
  - Adding a warning if a PR doesn't have an assignee
  - Adding a error if any of the tests fail(should also point to what failed)
  - Adding warnings if files(such as React components) were added without new tests/a modification to an existing one.
- Maybe use a specific eslint Action or [linty-fresh]() for only linting the files that have changed(in the frontend)
  - Don't think this is necessary though, since `eslint` already passes in all of the files currently so any failing change will fail the CI action, and thus we'll know one of the changes one of us made is the reason it's failing.

#### Also
Also note that pylint is failing in all of these builds: we can either only lint the files that have changed in a commit/PR, or we need to fix all of these errors.
- If we do want to fix all of these errors, I'll do it in a separate PR off of `backend/master` once this is merged.
- Could also just ignore all of the Django files that are failing `pylint`, since that's where a majority of the errors are. This could be annoying to do though.


requirements.txt
```diff
@@ -1,6 +1,6 @@
 Django~=2.2.6
 djangorestframework~=3.9.2
-psycopg2==2.8.3
+psycopg2-binary==2.8.4
```
gannonprudhomme: Per the pip package:

> The binary package is a practical choice for development and testing but in production it is advised to use the package built from sources.

So this isn't the best long term choice, but it's required for getting Django to run within Docker.

.github/lint-requirements.txt
```diff
@@ -0,0 +1,2 @@
+Django~=2.2.6
+pylint-django==2.0.11
```
gannonprudhomme: This file is for `pylint` to install, as it doesn't need all of the dependencies in `requirements.txt`

.github/workflows/ci-workflow.yml
```diff
@@ -0,0 +1,72 @@
+name: Continuous-Integration
+on: [push] # This workflow triggers anytime a commit is pushed
```
gannonprudhomme: Should we have it run on pull requests as well? Not sure if there's really a benefit to it in this case.

firejake308: Sometimes I'll commit some changes just to save my spot  or in case I need to revert. With the new workflow, I just won't be able to push these changes to origin, right?

firejake308:
To get the frontend test working, you should probably just remove the Meeting tests, since those are designed to fail until MeetingCard is implemented

requirements.txt
firejake308: Does GitHub Actions run the tests in a Docker environment? Or are you just preparing for the future?

.github/workflows/ci-workflow.yml
gannonprudhomme: If you do push them and rebase them later by squashing or whatever it will re run the CI steps since they're new commits(since rebasing makes copies of the commits you had). So if you squash some commits that were previously failing(which ends up making them pass) and force push, it will run the CI again on those commits and update the check mark/X.

This is why the most recent commit in this branch is the only one with a check/X next to it.

gannonprudhomme:
> To get the frontend test working, you should probably just remove the Meeting tests, since those are designed to fail until MeetingCard is implemented

I think it's fine if they're the only one failing, since we can just look at the CI job for frontend testing and make sure no other test is failing it other than the MeetingCard one. Alternatively, we could not merge #57 into `frontend/master` and just leave it in its branch until Ryan works on #54

requirements.txt
gannonprudhomme: So the Backend Testing step is basically required to do in Docker, since we need 2 containers, one for the Postgres DB, and one for the Django server. You may be able to do it just using GitHub Actions, but doing the set up in Docker greatly reduces how long it takes to get it up and running, since if it works in a local Docker container on our PC's, then it will most likely work in the GitHub Actions CI environment(which is one of the major benefits of containerization).

gannonprudhomme:
If y'all'd like to get Docker/the Backend Testing set up working locally, you can download Docker Desktop, then run `docker-compose up` in the root directory of the project. This will use the `docker-compose.yml` file to determine how to build the containers, and the Django container will use the `Dockerfile` to determine how to build itself further. Then, the `Django` container(through the `Dockerfile`) runs `docker-entrypoint.sh` to make migrations and run the tests.

Y'all shouldn't need to get it working since you can see what it does in the Backend Tesitng step, but it doesn't hurt to try to look at what everything does to understand it.

gannonprudhomme:
For the backend, do we want to fix all of the errors in the Django files that are listed in the failed job [here](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/pull/72/checks?check_run_id=302173198)(in the pylint step)? Or do we want to just only lint the files that were changed in the commit?

requirements.txt
gannonprudhomme: Likewise, the Backend Linting step uses [this action](https://github.com/cclauss/GitHub-Action-for-pylint), which also runs Docker.

.github/workflows/ci-workflow.yml
ryan-conn: I don't think there will be any benefit from running on pull requests either, considering the only time re-running would change the output is when files are changed, and that's already covered by running on push.

gannonprudhomme: Yeah I agree. Only thing that would need to be run only on pull requests is Danger, but that'll be covered in a different PR.

ryan-conn:
Linting only changed files would be nice if we set it up, but the main issue here is that our linting isn't set up like we want it, which will continue to be a problem even if we lint only changed files, since most of the issues pylint points out are django files like views/apps/tests, pointing out nonissues, so I think we should consider improving our pylintrc so we don't have to deal with it in the future.

gannonprudhomme:
> I'm pretty sure backend testing is running incorrectly, since the output of all the django commands is `Please configure /app/autoscheduler/config/postgres-info.json to include your postgres user and password.`
> I tried copying a pre-configured `postgres-info` file to Docker in the Dockerfile, but it still gave the same output, so I'm not sure exactly what to do to fix this.

It shouldn't matter, since the `docker.py` settings file overrides the values that the `postgres-info.json` are needed for. I don't think we'll be able to make it not have that output every time, but I can confirm that the tests correctly fail in Docker if they fail locally(and thus the job fails correctly)

gannonprudhomme:
> so I think we should consider improving our pylintrc so we don't have to deal with it in the future.

Improve in what way? By ignoring those Django files? Stuff like `views.py` will be used in the future once we start working on actual API requests. Likewise `scheduler/tests.py` will be replaced by a tests folder like `scraper` has once we start working on the actual scheduler.

ryan-conn:
> It shouldn't matter, since the `docker.py` settings file overrides the values that the `postgres-info.json` are needed for. I don't think we'll be able to make it not have that output every time, but I can confirm that the tests correctly fail in Docker if they fail locally(and thus the job fails correctly)

Alright, I was under the impression that it just wasn't running them at all since there isn't any output besides that for any of the django commands. If it fails properly, it should be fine, but if there's a way that we can see what the actual output of `manage.py test` is, it would be useful (right now, does it show which tests failed if it does fail, or not show any information about tests?).

> Improve in what way? By ignoring those Django files? Stuff like `views.py` will be used in the future once we start working on actual API requests. Likewise `scheduler/tests.py` will be replaced by a tests folder like `scraper` has once we start working on the actual scheduler.

Even when `views.py` is filled out, pylint will still give errors like ` autoscheduler/frontend/views.py:4:0: C0116: Missing function or method docstring (missing-function-docstring)` for not having a docstring for `index(request)`. The easiest way to go would be to just stop requiring docstrings for everything and leave it up to whoever's writing the code to determine what actually needs one, since there are a quite a few cases where pylint wants docstrings for simple classes/methods. For the other issues, it wouldn't take that long to clean up the code we have (just removing imports), and making sure that any code we push doesn't have unused imports either.

gannonprudhomme:
> right now, does it show which tests failed if it does fail, or not show any information about tests

It does indeed, however we don't have any failing tests so it's not right now. See [here](https://github.com/gannonprudhomme/Automatic-Aggie-Scheduler/runs/284913681). It's on the fork, but I used the same code from the fork in this PR/branch. The first error message I'm not sure what it's referring to, but you can see a failed test below it.

> . The easiest way to go would be to just stop requiring docstrings for everything and leave it up to whoever's writing the code to determine what actually needs one, since there are a quite a few cases where pylint wants docstrings for simple classes/methods.

We have `missing-module-docstrings` disabled in `backend/master`, which I mention in the PR description 😛

ryan-conn:
About your comments in the PR message: I agree that setting up Danger JS could be useful (but obviously in a different pull request, we could create an issue for it), and that setting up linty-fresh shouldn't be necessary as long as we fix all our linting issues before merging into master branches.

gannonprudhomme:
Just as a note, you can see what danger would look like [here](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/pull/76)

ryan-conn:
> We have `missing-module-docstrings` disabled in `backend/master`, which I mention in the PR description 😛

I meant to disable missing class and function docstrings as well (like the `index(request)` example I gave in each `views.py` file), since they're all separate rules in pylint.

gannonprudhomme:
> > We have `missing-module-docstrings` disabled in `backend/master`, which I mention in the PR description 😛
> 
> I meant to disable missing class and function docstrings as well (like the `index(request)` example I gave in each `views.py` file), since they're all separate rules in pylint.

Oh I totally beaned you did say `missing-function-docstring`. That being said, I do want to enforce function docstrings so we just gotta satisfy it in that case

gannonprudhomme:
Alright im going to merge this. Will probably make a PR for moving the docker files into a separate folder, as well as one for Danger.

PR #77, opened 2019-11-15
Fixed types to correctly declare class members

ryan-conn:
Currently, each of the classes we made for types call `Object.assign()` in the constructor to assign class members. However, when I was working on MeetingCard, I realized that this doesn't actually add it to the properties typescript gives to its class. Declaring all members up-front fixes this.


gannonprudhomme:
Looks fine to me, but gonna leave it up to Adel to approve it.

PR #82, opened 2019-11-15
Frontend/master

f4alt:
welcome page skeleton is done.
Still needs:
1) google login redirect and logo
2) text to be centered

later:
1) linking semester dropdown text to following page


firejake308:
It looks like #83 was what you meant to do. Can we close this one?

gannonprudhomme:
yeah we can, closing

PR #84, opened 2019-11-16
Split actions workflow into separate front/backend parts

gannonprudhomme:
There's really no need for the frontend steps to be checked on the backend branches and vice-versa. As such, this splits up the action into separate parts, which will only be run on their respective branches on push. 

This way, we won't have failing jobs on  branches that are caused by problems in the backend, such as the Django files failing pylint.

Also, I won't be merging these changes into `frontend/master` and `backend/master` until I also make a PR for Danger, which I'll make after this gets merged.

### DIFF
--- a/.github/workflows/backend-workflow.yml
+++ b/.github/workflows/backend-workflow.yml
@@ -23,4 +23,4 @@ jobs:
         uses: cclauss/GitHub-Action-for-pylint@master
         with:
         # Installs Django and pylint-django, then runs pylint on all of the files in autoscheduler/
-          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/**/*.py --load-plugins pylint_django
+          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/** --load-plugins pylint_django
--- a/.github/workflows/backend-workflow.yml
+++ b/.github/workflows/backend-workflow.yml
@@ -1,0 +1,26 @@
+name: Backend CI
+on: 
+  push:
+    branches-ignore:
+      - 'frontend/**' # Don't run on any frontend branches
+
+jobs:
+  test-backend:
+    name: 'Testing'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: docker-build
+      # This option makes the Postgres container exit when the autoscheduler container does
+        run: docker-compose up --abort-on-container-exit
+
+  lint-backend:
+    name: 'Linting'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: pylint
+        uses: cclauss/GitHub-Action-for-pylint@master
+        with:
+        # Installs Django and pylint-django, then runs pylint on all of the files in autoscheduler/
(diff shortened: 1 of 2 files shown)


.github/workflows/backend-workflow.yml
```diff
@@ -0,0 +1,26 @@
+name: Backend CI
+on: 
+  push:
+    branches-ignore:
+      - 'frontend/**' # Don't run on any frontend branches
```
gannonprudhomme: Rather than doing `branches: - 'backend/**`, we just exclude frontend branches from running it. This way pushes to the master or branch or small direct-to-master fixes (such as this one) have the actions run on their commits.

.github/workflows/backend-workflow.yml
```diff
@@ -0,0 +1,26 @@
...
+  lint-backend:
+    name: 'Linting'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: pylint
+        uses: cclauss/GitHub-Action-for-pylint@master
+        with:
+        # Installs Django and pylint-django, then runs pylint on all of the files in autoscheduler/
+          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/**/*.py --load-plugins pylint_django
```
gannonprudhomme: It turns out that this doesn't actually get `pylint` to lint *all* of the files, I think it just does files that are one directory down(such as `scraper/banner_requests.py`. For instance, right now it's not linting `autoscheduler/autoscheduler/settings/docker.py`, which has a bunch of linting errors. As such, I think it should be replaced with:

```suggestion
          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/** --load-plugins pylint_django
```

We should be able to just try these locally by running the command (just the second half after the semi-colon)

.github/workflows/frontend-workflow.yml
```diff
@@ -1,29 +1,12 @@
...
-    steps:
-      - uses: actions/checkout@master
-      - name: pylint
-        uses: cclauss/GitHub-Action-for-pylint@master
-        with:
-        # Installs Django and pylint-django, then runs pylint on all of the files in autoscheduler/
-          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/**/*.py --load-plugins pylint_django
-
   test-frontend:
-    name: 'Frontend Testing'
```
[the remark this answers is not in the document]
gannonprudhomme: If we did it this way, then in the checks (like at the bottom of this PR), it would say `Frontend CI / Frontend Testing (push)`, which I thought was redundant since it always includes the workflow name (`Frontend CI` in this case) when listing the checks. Do you think doing it like that helps the clarity regardless?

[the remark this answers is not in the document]
firejake308: Oh no, it's fine in that case. For whatever reason, I was blind and didn't see the Frontend CI prefix. Looks good to me!

.github/workflows/backend-workflow.yml
firejake308: That seems to be how it works on my computer as well

gannonprudhomme:
Gonna squash & merge with master, then will create the Danger PR. After it's merged into `master` as well, then I'll merge the changes into `backend/master` and `frontend/master`

PR #86, opened 2019-11-16
Implement Department Scraping

### LINKED ISSUE #41: Implement Department Scraping

btgathright:
Closes #41. I'm not sure if the pass keyword at the end of the handle should be removed or not. Also the file I uploaded had a different name "scrape_depts.py" instead of "scrape_departments.py" could someone with authority to delete files remove the "scrape_departments.py" file or rename the "scrape_depts.py" file.


autoscheduler/scraper/management/commands/scrape_depts.py
```diff
@@ -0,0 +1,27 @@
...
+        term = add_arguments()
+        var request = BannerRequests(term)
+        depts = request.get_departments(term,amount)
+
+    def add_arguments(self, parser):
+        parser.add_argument('--term', type=str, required=False, default="201931")
+
+    def handle(self, *args, **options):
+        start = time.time()
+        scrape_departments()
```
gannonprudhomme: To actually access the term from `scrape_departments`, you need to change it to accept a `term: str` argument, and then change this line in handle to 
```suggestion
        scrape_departments(options['--term'])
```

autoscheduler/scraper/management/commands/scrape_departments.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class Command(base.BaseCommand):
+    """ Gets all departments from banner and adds them to the database """
+
+    def add_arguments(self, parser):
+        parser.add_argument('--term', type=str, required=False, default="201931")
+
+    def handle(self, *args, **options):
+        pass
```
gannonprudhomme: While I do prefer the file to be called `scrape_depts` rather than `scrape_departments` since it's less to type, why make a completely different file instead of either using this one or renaming it? 

Either delete this file, or move your changes from `scrape_depts` into it and delete it

autoscheduler/scraper/management/commands/scrape_depts.py
```diff
@@ -0,0 +1,27 @@
...
+        end = time.time()
+        seconds_elapsed = int(end - start)
+        time_delta = datetime.timedelta(seconds=seconds_elapsed)
+        print(f"Finished scraping departments in {time_delta}")
+        pass
```
gannonprudhomme: Needs a new line at the end of the file.

btgathright:
I updated the changes request. Can you rereview this?

gannonprudhomme:
Sorry I must have missed the description the first time. What do you mean by

>someone with authority to delete files remove the "scrape_departments.py" file or rename the "scrape_depts.py" file.

You should be able to delete the files just fine, are you uploading the file changes to GitHub directly? Message me on slack as we need to set up Git for you if so.

autoscheduler/scraper/management/commands/scrape_departments.py
```diff
@@ -0,0 +1,28 @@
...
+from scraper.models import Department
+
+import time
+import datetime
+
+class Command(base.BaseCommand):
+    """ Gets all departments from banner and adds them to the database """
+
+    def scrape_departments():
+        term = add_arguments()x
```
gannonprudhomme: Are you running these changes before you commit them? This `x` should have been a quick find. Likewise, the `scrape_departments` function doesn't have a `term` parameter, and the `term = add_argument()` line shouldn't be there, as `add_argument()` isn't a function for us to call(it's for `Django` to use), nor does it return a string.

autoscheduler/scraper/management/commands/scrape_departments.py
```diff
@@ -0,0 +1,28 @@
...
+        var request = BannerRequests(term)
+        depts = request.get_departments(term,amount)
+
```
gannonprudhomme: Likewise, you're still missing the bulk of the feature, as right now all this command does is retrieve the data from Banner and nothing else. You need to iterate through `depts` and call some `parse_depts` function, which would then save each department to the database with Django.

It would look something like

```python
def parse_dept(data, term: str):
    dept = Department(code=data['code'], desc=data['desc'], term=term)
    dept.save() # Saves the department to the database
```

Not sure about the actual values(i.e. `data['desc']`) but it should be something along those lines. To check, you should be able to just `print(depts)` in the `scrape_departments` function, then actually run it by typing `python manage.py scrape_departments` when you're in the first `autoscheduler` folder(that has `frontend/`, `scraper/` in it).

gannonprudhomme:
Also, this branch still is not up to date with `backend/master`, so we need to merge it into this branch.

gannonprudhomme:
Merged with `backend/master` so that the GitHub Actions checks could run. Should squash this into one commit upon approval

gannonprudhomme:
Added the tests per #93 . Had to name the file `test_scrape_depts.py` rather than `scrape_depts_tests.py` (like I would like) so that the Django test runner would actually pick it up (but `pytest` picks it up regardless of the name). 

Once `master` gets merged into `backend/master`, I'll work on #87 , which will include changing the `Backend Testing` step to use `pytest` rather than Django's test runner, I'll rename it back and remove the `__init__.py` that I had to include in the `commands/` folder.

btgathright:
When I run the checks the only errors are in frontend testing and the backend linting, however the backend linting errors don't show under scraper. How do I see where my linting errors are?

gannonprudhomme:
> When I run the checks the only errors are in frontend testing and the backend linting, however the backend linting errors don't show under scraper. 

Yeah this needs to be updated with the changes on `backend/master`, which include fixing the checks so that they run `pylint` in the correct directories, as it's not running in sub-folders right now (such as scraper/management)

>How do I see where my linting errors are?

You can lint the specific file by running, `pylint --load-plugins pylint_django autoscheduler/scraper/management/commands/scrape_depts.py`

btgathright:
I get an error for no module pylint_django

btgathright:
Completed with no linting errors on my end.

autoscheduler/scraper/management/commands/scrape_depts.py
```diff
@@ -1,23 +1,29 @@
...
 def scrape_departments(term):
-        request = BannerRequests(term)
-        json = request.get_departments()
-        print(json)
-        parse_departments(json, term)
+    """ Takes term input and collects json object of departments """
+
+    request = BannerRequests(term)
+    json = request.get_departments()
+    print(json)
```
gannonprudhomme: Need to remove the print statement, as it'll clutter up the output when we run it

gannonprudhomme:
Looks good. I'm going to squash some of these commits and fix some linting errors in `tests_scrape_depts.py`, then we'll be good.

PR #83, opened 2019-11-16
Added landing page

f4alt:
welcome page skeleton is done.
Still needs:
- google login redirect and logo
- text to be centered

later:
- linking semester dropdown text to following page

PR #88, opened 2019-11-18
Added CSS modules

### LINKED ISSUE #73: Implement css and font loading for Webpack/React
Need to look into and implement how to do css and font loading for Webpack.

[This article](https://medium.com/@dmitrynozhenko/9-ways-to-implement-css-in-react-js-ccea4d543aa3) outlines a few ways for how to go about it. I was mainly looking at the first one(Standard css) and the third(css modules), but wasn't sure which was the best. 

Likewise for fonts, we'll either need to use [url-loader](https://www.npmjs.com/package/url-loader) or [file-loader](https://www.npmjs.com/package/file-loader), depending on if we want to use fonts from a CDN, or if we want to download the fonts locally. If we do want to download the fonts locally, we can use [typefaces](https://github.com/KyleAMathews/typefaces). 

This should also include setting the fonts in the Theme for Material-UI, which you can read more about [here](https://material-ui.com/customization/typography/).

firejake308:
CSS Modules have been implemented. Now, we can write all of our CSS in a regular CSS file adjacent to the component, `import * as styles from 'Component.css'`, and then set `className={styles.myclass}` to apply custom CSS to our components. The compiler will automatically generate a cryptic hash in front of the class name to ensure that the styles from one component don't conflict with those from another. If you wish to dynamically determine a particular style rule, then continue to use the `style` attribute in JS.

Closes #73


firejake308:
One thing I will point out is that I had to update `package-lock.json` because my old version of NPM had been using SHA1 integrities, but the new version of NPM used SHA512 integrities. To clear up the conflict, I elected to `rm -rf node_modules` and re-install all of the packages, which resulted in the conversion of everything to SHA512 integrity, as desired. It also updated some packages, resulting in the many changes to `package-lock.json` that you see.

gannonprudhomme:
Should we merge this before #83 and use CSS modules within the landing page?

firejake308:
Yes, that would be ideal

firejake308:
Oops, just realized that Jest isn't playing nicely with CSS modules; I'll be fixing that before we merge

firejake308:
I think this can go ahead and be merged; the linting error is from Ryan's type declaration fix and will be fixed when live-reload is merged

gannonprudhomme:
Also just btw, I think you commented on the commits themselves rather than from the PR, so we can't see the comments in the Files Changed menu or the lines that you commented on in the PR view (at least for me).

gannonprudhomme:
If `.d.ts` files are auto-generated, should they be committed in the first place? Likewise, when I remove the `.d.ts` file, it takes 2 builds for it to run correctly, the first one auto generates it then throws an error, then the next build it runs without error. Is this how it runs for y'all?

gannonprudhomme:
Also, can you merge this with the up-to-date `frontend/master`

firejake308:
> If `.d.ts` files are auto-generated, should they be committed in the first place? Likewise, when I remove the `.d.ts` file, it takes 2 builds for it to run correctly, the first one auto generates it then throws an error, then the next build it runs without error. Is this how it runs for y'all?

I believe they should be committed for the exact reasons that you specified; so that you can run it in one build. I believe that what you described is the expected behavior, since `Component.tsx` will first try to import `Component.css.d.ts`, find none, and throw an error. During that build, however, the `typings-for-css-modules-loader` plugin will generate a `Component.css.d.ts` from `Component.css`. This will be available for the second build, which will pass. When in development, webpack will be watching for the change and update when the new file is generated. When building for production, you should have already generated the `d.ts` during dev testing, and one build should be fine. If you didn't, just run it again.

Issue #75, opened 2019-11-14
Implement Danger.js


gannonprudhomme:
[Danger.js](https://danger.systems/js/) is useful for various things, such as:
  - Adding a warning/error PR comment if a PR is over 500 lines
  - Adding a warning if a PR doesn't have an assignee
  - Adding a error if any of the tests fail(should also point to what failed)
  - Adding warnings if files(such as React components) were added without new tests/a modification to an existing one.

Probably gonna add more to this

Issue #87, opened 2019-11-17
Fix failing pylint files


gannonprudhomme:
There are a few Django related files that are failing the `Backend Linting` step, so we need to fix them so that the step will pass

gannonprudhomme:
Also need to fix `pytest`

gannonprudhomme:
Closing per #100

PR #91, opened 2019-11-19
Added Danger.js

gannonprudhomme:
See #75

### DIFF
--- a/.github/dangerfile.ts
+++ b/.github/dangerfile.ts
@@ -66,7 +66,7 @@ if (hasBackendChanges && !hasBackendTestChanges) {
         + ' refactoring existing code.');
 }
 
-// Prevent frontend changes in the backend
+// Add a failure if there are backend changes on a frontend branch
 if (pr.head.label.includes('frontend/') && backendTouchedFiles.length > 0) {
   // Add a warning if there's touched files in the backend
   fail(':exclamation: No Python/Django related files should be touched on a frontend-'
--- a/.github/dangerfile.ts
+++ b/.github/dangerfile.ts
@@ -28,6 +28,11 @@ if (pr.body.length === 0) {
   warn('Please add a PR description');
 }
 
+// Add a warning if there aren't any labels on this PR
+if (github.issue.labels.length === 0) {
+  warn('Please add a label to this PR');
+}
+
 /* Code-specific warnings */
 
 let frontendTouchedFiles = touchedFiles.filter(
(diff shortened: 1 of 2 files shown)


gannonprudhomme:
I intentionally didn't add a description or added assignee's/a label so that it could give an example of errors

gannonprudhomme:
Given that `package-lock.json` file updates can be thousands of lines long, should I make it so that the bigPRCount check includes all additions/removals excluding the changes in `package-lock.json`?

firejake308:
If that is possible, then that would be ideal, because I looked at the commit where we added Jest, and that one library added like 7000 lines to the package-lock.

gannonprudhomme:
Added the `package-lock.json` big PR check, re-requesting review. 

Also, note that the comment from the GitHub Actions bot was removed, since the warnings it was mentioning were fixed. However, I think you have to push again for it to update, but we may be able to add an option so that it runs on update.

gannonprudhomme:
> However, I think you have to push again for it to update, but we may be able to add an option so that it runs on update.

Turns out this was a lot easier than I thought. Simply had to add the `types` property to the workflow trigger. See [here](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/events-that-trigger-workflows#pull-request-event-pull_request) for more info.

.github/dangerfile.ts
```diff
@@ -0,0 +1,108 @@
...
+    commitContainsPackageLock = true;
+  }
+});
+
+// Check if the PR line count(other than package-lock.json?) is over 500 lines
+if (commitContainsPackageLock) {
+  // Counts the changes in package-lock.json, so that the big PR check doesn't include them
+  danger.git.diffForFile(packageLockPath).then((diff) => {
+    const changed = diff.added.split('\n').length + diff.removed.split('\n').length;
+    checkForBigPR(changed);
```
firejake308: Looks like it should work to me

.github/workflows/danger-workflow.yml
```diff
@@ -0,0 +1,24 @@
+name: Danger
+on: 
+  pull_request:
+    # By default, the pull_request event only triggers on [opened, synchronize, reopened]
+    types: [opened, synchronize, reoponed, assigned, unassigned, labeled, unlabeled]
```
firejake308: So I added the label, the Danger/Danger CI check ran for a third time, and the comment disappeared. Looks like it works. And I agree that these are the events we need to re-run on, since the other events listed in the reference would not affect the pass/fail state of Danger

gannonprudhomme:
Squashed & merged. Going to merge the branches into `backend/master` and `frontend/master`, then will make a PR for #87 on the backend

Issue #44, opened 2019-10-29
Implement course scraping


gannonprudhomme:
Need to implement course scraping, which will also include scraping/parsing for `Instructor`, `Section` and `Meeting`. 

To begin working on one of the issues(`parse_section`, `parse_course`, etc), first checkout `backend/course-scraping` then make a branch off of it. Then add commits to it, and once you're done make a pull request to merge your branch into `backend/course-scraping`.

ryan-conn:
When running async functions in commands (such as `scrape-courses`), we'll need to do the following in order to run them synchronously:
```
# in scrape-courses.py (function to synchronously run async):
def run_async(function, args):
    loop = asyncio.get_event_loop()
    tasks = [asyncio.ensure_future(function(args))]
    return loop.run_until_complete(asyncio.wait(tasks))
```
And in `handle(self, *args, **options)`, put the following to get the data:
```
depts = ["CSCE", "MATH", "CHEM"]
banner = BannerRequests("201931")
data = run_async(banner.search, depts)
```

gannonprudhomme:
Just by the way @ThatJuanGuy, I updated the description of this issue to reflect what we're doing instead of that `JSESSIONID` cookie stuff. It turns out I was completely wrong about needing to directly deal with cookies like that, as told to me by this random internet god on my other repo [here](https://github.com/gannonprudhomme/AutoScheduler/issues/10)

PR #94, opened 2019-11-19
Implemented course scraping

ThatJuanGuy:
…_instructor functions

There were some data fields I couldn't find in the JSON. I just left these as empty strings or False for now. Will add comments over the code to show where these are.


autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+from course import generate_course_id
 
-def parse_section(section):
-    " parse_course() must be done first, and pass the data for each section here "
+def parse_section(course):
+    #gets values to create object
+    section_id = ''
+    section_subject = course['subject']
+    course_number = course['courseNumber']
+    section_number = ''
```
ThatJuanGuy: Could not find. Left '' as a place holder

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
 from django.core.management import base
 from scraper.banner_requests import BannerRequests
 from scraper.models import Course, Instructor, Section, Meeting
+from course import generate_course_id
 
-def parse_section(section):
-    " parse_course() must be done first, and pass the data for each section here "
+def parse_section(course):
+    #gets values to create object
+    section_id = ''
```
ThatJuanGuy: Couldn't find a section id labeled in the data. It might be id but I'm not sure. In that case, section_id = course['id']

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    course_number = course['courseNumber']
+    course_title = course['courseTitle']
+    course_term_code = course['term']
+    course_id = generate_course_id(course_dept, course_number, course_term_code)#primary key
+
+    course_desc = ''#long description-could not find in data
+
+    course_prereqs = ''
+    course_coreqs = ''
+    course_cd = False #to be implented later. cuurrently set as false
```
ThatJuanGuy: Left with course_cd and course_icd with False as a placeholder because of the meeting Sunday when it was decided we didn't actually want to implement core curriculum tags yet. These fields are just in here to match with the model. course_core_curriculum is also '' as a placeholder for the same reason.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
-    " Parse a course and save all of its relevant information to the database "
+    course_dept = course['subject']
+    course_number = course['courseNumber']
+    course_title = course['courseTitle']
+    course_term_code = course['term']
+    course_id = generate_course_id(course_dept, course_number, course_term_code)#primary key
+
+    course_desc = ''#long description-could not find in data
+
+    course_prereqs = ''
```
ThatJuanGuy: Could not find prereqs or coreqs in data. I left them as '' for now so that I could match the model. Honestly, I don't even really think these are necessary though. Aside from checking coreqs to make sure a user registers all required corequistes and giving them a warning if they don't, I don't see any reason why we would need this data. It also seems like a feature that should be implemented later and not prioritized right now.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    meeting_section = ''
+
+    #creates and saves meeting object
+    m = Meeting(id=meeting_id, crn=meeting_crn, building=meeting_building, meeting_days=attend_days, start_time=meeting_start_time, end_time=meeting_end_time, meeting_type=meeting_class_type, section=meeting_section)
+    m.save()
+
+def parse_instructor(course):
+    #gets values to create object
+    for i in course['faculty']:
+        instructor_id = ''
```
ThatJuanGuy: I was comparing the data to the JSON file on Gannon's old repository I saw there was a 'instructorPidm' field which seemed like an id and changed only when the instructor changed. Left as blank for now because I couldn't find the equivalent in the python dictionary.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    meeting_class_days = ['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday']
+    attend_days = []
+    for j in meeting_class_days:
+        if i['meetingTime'][j] == True:
+            attend_days.append(j)
+    meeting_start_time = i['meetingTime']['beginTime']
+    meeting_end_time = i['meetingTime']['endTime']
+
+    meeting_class_type = i['meetingTime']['meetingType']
+    meeting_section = ''
```
ThatJuanGuy: This is just the section number. However, I couldn't find the section number in the data. Thus I left it as '' for now as well.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    #calls parse meeting for each meeting time in the section
+    for i in course['meetingsFaculty']:
+        parse_meeting(i)
+
+def parse_meeting(i):
+    #gets values to create object
+    meeting_id = ''
+    meeting_crn = i['meetingTime']['courseReferenceNumber']
+
+    meeting_building = i['meetingTime']['building']
```
ThatJuanGuy: Should we add room number as a field to the meeting model? We have the data.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    s = Section(id=section_id, subject=section_subject, course_num=course_number, section_num=section_number, term_code=section_term_code, min_credits=section_min_credits, max_credits=section_max_credits, max_enrollment=section_max_enrollment, current_enrollment=section_current_enrollment, instructor=section_instructor)
+    s.save()
+
+    #calls parse meeting for each meeting time in the section
+    for i in course['meetingsFaculty']:
+        parse_meeting(i)
+
+def parse_meeting(i):
+    #gets values to create object
+    meeting_id = ''
```
ThatJuanGuy: Could not find some sort of reliable meeting id in data. Left as '' for now.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    section_term_code = course['term']
+    
+    section_min_credits = course['creditHourLow']
+    section_max_credits = course['creditHourHigh']
 
-def parse_meeting(meeting):
-    " parse_section() must be done first, and pass the data for each meeting here"
+    section_max_enrollment = course['maximumEnrollment']
+    section_current_enrollment = course['enrollment']
+    section_instructor = ''
```
ThatJuanGuy: Unsure of what to do in this case because faculty is a list and this is a string field. In some courses, there might be more than one faculty member listed. For now I just made it so it puts all the instructor names into a list and outputs that as a string. Also, I just realized this results in an empty string sometimes. I'ma fix it real quick.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    section_number = ''
+    section_term_code = course['term']
+    
+    section_min_credits = course['creditHourLow']
+    section_max_credits = course['creditHourHigh']
 
-def parse_meeting(meeting):
-    " parse_section() must be done first, and pass the data for each meeting here"
+    section_max_enrollment = course['maximumEnrollment']
+    section_current_enrollment = course['enrollment']
```
ThatJuanGuy: Are we actually going to use the current and max enrollment for anything? I don't see how this is really going to help create schedule unless we're doing prioritize class size or something.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+def parse_section(course):
+    #gets values to create object
+    section_id = ''
+    section_subject = course['subject']
+    course_number = course['courseNumber']
+    section_number = ''
+    section_term_code = course['term']
+    
+    section_min_credits = course['creditHourLow']
+    section_max_credits = course['creditHourHigh']
```
ThatJuanGuy: This is different between the course and section models. Course just has one field for credit hours but section has 2. Should this be consistent across both?

autoscheduler/scraper/management/commands/scrape_courses.py
ThatJuanGuy: Fixed the empty string bug. Is there a way to add that to this pull request or do I have to make a new one? If I have to make a new one, is it possible to keep these comments or do I have to copy paste them to the new request?

autoscheduler/scraper/management/commands/scrape_courses.py
ThatJuanGuy: Couldn't find a section id labeled in the data. It might be id but I'm not sure. In that case, section_id = course['id']

autoscheduler/scraper/management/commands/scrape_courses.py
firejake308: Just add a commit to the same branch and the PR will update automatically

autoscheduler/scraper/management/commands/scrape_courses.py
firejake308: Since `instructor.id` is a primary key, we're going to need unique values for this at some point

autoscheduler/scraper/management/commands/scrape_courses.py
ryan-conn: We won't want to create schedules with sections that are full in them, so enrollment will definitely be useful. We might want to display that and the max enrollment to the user too, but if we're not I agree that I don't see a point in storing it.

gannonprudhomme:
Honestly I had no idea you were doing the entire thing, but great work! In that case, this should be merging into `backend/master` rather than `backend/scraper/course-scheduling`, since this completes the entirety of #44

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
...
+    section_current_enrollment = course['enrollment']
+    section_instructor = ''
+    for i in course['faculty']:
+        instructor_name = i['displayName']
+        section_instructor += instructor_name + ','
 
-def parse_instructor(instructor):
-    " parse_course() must be done first, and pass the data for the instructor here "
+    #creates and saves section object
+    s = Section(id=section_id, subject=section_subject, course_num=course_number, section_num=section_number, term_code=section_term_code, min_credits=section_min_credits, max_credits=section_max_credits, max_enrollment=section_max_enrollment, current_enrollment=section_current_enrollment, instructor=section_instructor)
```
ryan-conn: You'll want to run pylint on this file (message me on Slack if you need help setting it up), as there are a few lines that it wants fixed (for example, this one gives the message `Line too long (216/90)`)

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -4,18 +4,89 @@
 from django.core.management import base
 from scraper.banner_requests import BannerRequests
 from scraper.models import Course, Instructor, Section, Meeting
+from course import generate_course_id
```
ryan-conn:
```suggestion
from scraper.models.course import generate_course_id
```

autoscheduler/scraper/management/commands/scrape_courses.py
ryan-conn: `prequisites` should be changed to `prerequisites`

autoscheduler/scraper/management/commands/scrape_courses.py
ThatJuanGuy: I don't know how to run it on VSC but it worked when I copy pasted the output for scrape courses into Spyder and ran the code using that data.

gannonprudhomme:
```suggestion
from scraper.models.course import generate_course_id
```

This was causing an error, but should always use absolute imports regardless

ryan-conn: You can run `python manage.py scrape_courses` in the terminal to run it in VSCode. Let me know on Slack if you have any questions about setting anything up

gannonprudhomme:
Also, since we have no immediate plans to use `prerequisites`, `corequisites`, `cd`, or `icd`, I say we can remove them from the `Course` model. If we need them in the future we can always add them.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,118 @@
...
+
+    section_max_enrollment = course['maximumEnrollment']
+    section_current_enrollment = course['enrollment']
+    section_instructor = ''
+    for i in course['faculty']:
+        instructor_name = i['displayName']
+        section_instructor += instructor_name + ','
+
+    #creates and saves section object
+    s = Section(id=section_id, subject=section_subject, course_num=course_number, section_num=section_number, term_code=section_term_code, min_credits=section_min_credits, max_credits=section_max_credits, max_enrollment=section_max_enrollment, current_enrollment=section_current_enrollment, instructor=section_instructor)
```
gannonprudhomme: While we may be able to pass in the primary key for the Instructor, instead I think `parse_instructor` should return an `Instructor` model (that's already saved to the DB), which you can then pass into as an argument to `parse_section`. Then when you create the Section object, you simply set the `instructor` field to the `Instructor` model that was passed into the function.

gannonprudhomme: While we may be able to pass in the primary key for the Instructor, instead I think `parse_instructor` should return an `Instructor` model (that's already saved to the DB), which you can then pass into as an argument to `parse_section`. Then when you create the Section object, you simply set the `instructor` field to the `Instructor` model that was passed into the function.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,118 @@
...
+    s.save()
+
+    #calls parse meeting for each meeting time in the section
+    for i in course['meetingsFaculty']:
+        parse_meeting(i)
+
+def parse_meeting(i):
+    #gets values to create object
+    meeting_id = ''
```
gannonprudhomme: Instead of the parameter and variable being `i`, I'd rather it be something like `meetingJson` or `meetingData`

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Since the section number isn't in the meeting data, you need to pass it in as an argument to the `parse_meeting` function. However, it's not actually the section number, but rather the `Section` model itself. As such, you should pass in the `Section` model that you create in `parse_section` to it.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,118 @@
...
+
+    course_prereqs = ''
+    course_coreqs = ''
+    course_cd = False #to be implented later. cuurrently set as false
+    course_icd = False #to be implented later
+    course_core_curriculum = ''#to be implented later
+    course_credit_hours = course['creditHourLow']
+
+    #creates and saves course object
+    c = Course(id=course_id, dept=course_dept, course_num=course_number, title=course_title, description=course_desc, prequisites=course_prereqs, corequisites=course_coreqs, cd=course_cd, icd=course_icd, core_curriculum=course_core_curriculum, credit_hours=course_credit_hours)
```
gannonprudhomme: Meant to say it before, but can you add a `term` field to the `Course` model, that way we can determine which courses in the `courses` table correspond to which term (otherwise we could show the users courses that aren't actually available that term).

gannonprudhomme:
### Note:
I rebased so make sure you run `git reset --hard origin/backend/scraper/course-scraping-parse-course` before pushing or making any local changes.

 Honestly I hate that it says I authored *all* the commits, so I'll get the main branch author to do it next time

gannonprudhomme:
Also, make sure to run through all the comments and resolve/hide any comments that aren't relevant anymore 😄

autoscheduler/scraper/management/commands/scrape_courses.py
ThatJuanGuy: Used course['id'] for section id

.vs/VSWorkspaceState.json
```diff
@@ -0,0 +1,6 @@
+{
```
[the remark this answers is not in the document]
ThatJuanGuy: Will do

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,105 @@
...
+    #gets values to create object
+    for i in course['faculty']:
+        instructor_id = i['bannerId']
+        instructor_email = i['emailAddress']
+        instructor_name = i['displayName']
+        #creates and saves instructor object
+        i = Instructor(id=instructor_id, email_address=instructor_email, name=instructor_name)
+        i.save()
+        #calls parse_section
+        parse_section(course,i)
```
gannonprudhomme: Since the faculty block is like its own part of the section input, I think `parse_instructor` should return the Instructor model, and instead of calling parse_section in it, you can call it within `parse_course`.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,105 @@
...
+    meeting_class_type = i['meetingTime']['meetingType']
+    meeting_section = ''#this should be an object
+    #creates and saves meeting object
+    m = Meeting(id=meeting_id, crn=meeting_crn, building=meeting_building, meeting_days=attend_days, start_time=meeting_start_time, end_time=meeting_end_time, meeting_type=meeting_class_type, section=meeting_section)
+    m.save()
+
+def parse_instructor(course):
+    #gets values to create object
+    for i in course['faculty']:
+        instructor_id = i['bannerId']
```
gannonprudhomme: Since the bannerId isn't actually unique to a professor, and it looks like the `id` object is always null in the section data from Banner, I think we should change it to use the instructor's name as the primary key. We can't do the email, since some professors don't have an email listed (not sure how that makes any sense but w/e lol)

ThatJuanGuy: Ok. Got it

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,106 @@
...
+    section_current_enrollment = courseData['enrollment']
+    section_instructor = instructor_object
+    #creates and saves section object
+    section_object = Section(id=section_id, subject=section_subject, course_num=course_number, section_num=section_number, term_code=section_term_code, min_credits=section_min_credits, max_credits=section_max_credits, max_enrollment=section_max_enrollment, current_enrollment=section_current_enrollment, instructor=section_instructor)
+    section_object.save()
+    #calls parse meeting for each meeting time in the section
+    meeting_count=0
+    for i in courseData['meetingsFaculty']:
+        parse_meeting(i,section_object,meeting_count)
+        meeting_count+=1
```
gannonprudhomme: I would either change this to a for i in range loop, and have it then be `parse_meeting(courseData['meetingsFaculty'][i], section_object, i)`, or change it the loop variable to be `for meeting in courseData['meetingsFaculty']`, as i is usually used as the typical `for(int i = 0; ...)` integer loop variable (imo).

autoscheduler/scraper/models/section.py
```diff
@@ -12,29 +14,38 @@ class Section(models.Model):
...
+    return "".join((section_num, meetings_count))
+
+def generate_meeting_time(string_time: str):
+    """ Generates a meeting time from a string in format hhmm. ex) 1245 = 12:45am. 1830 = 6:30 pm."""
+    hour = int(string_time[0:2])
+    minute = int(string_time[2:4])
+    time = datetime.time(hour,minute)
+
+    return time
+
```
gannonprudhomme: While technically meetings are the only time we parse the meeting, I don't necessarily think it's necessary to have it in this file (rather, it should be in `scrape_courses.py`). Likewise, this needs to account for when the meeting times are empty, so you need to check if string_time is empty, and if it is, return `None`.

autoscheduler/scraper/models/instructor.py
```diff
@@ -2,7 +2,7 @@
 
 class Instructor(models.Model):
     """ Model that represents a professor, who may teach many courses """
-    id = models.IntegerField(primary_key=True)
+    id = models.CharField(max_length=100,primary_key=True)
     email_address = models.CharField(max_length=48, null=True)
     name = models.CharField(max_length=64)
```
gannonprudhomme: Since we're  going to use the name for the primary key for now, the id's max length should match the name's, and name should be removed.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,106 @@
...
+    meeting_crn = meetingsData['meetingTime']['courseReferenceNumber']
+    meeting_building = meetingsData['meetingTime']['building']
+    #checks which days classes are on and adds them to attend_days
+    meeting_class_days = ['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday']
+    attend_days = [False,False,False,False,False,False,False]
+    day_string = 'monday'
+    for j in range(0,len(meeting_class_days)):
+        day_string = meeting_class_days[j]
+        if meetingsData['meetingTime'][day_string] == True:
+            attend_days[j] = True
```
gannonprudhomme: First, I think this should be abstracted into its own function so we can test it.

Second, I think this can be simplified closer to how you had it previously. Instead of pre-filling` attend_days`, you can do it like:

```python
    meeting_class_days = ['monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday']
    attend_days = []
    for day in meeting_class_days:
        attend_days.append(meetingsData['meetingTime'][day])
```

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,106 @@
...
+    #creates and saves section object
+    section_object = Section(id=section_id, subject=section_subject, course_num=course_number, section_num=section_number, term_code=section_term_code, min_credits=section_min_credits, max_credits=section_max_credits, max_enrollment=section_max_enrollment, current_enrollment=section_current_enrollment, instructor=section_instructor)
+    section_object.save()
+    #calls parse meeting for each meeting time in the section
+    meeting_count=0
+    for i in courseData['meetingsFaculty']:
+        parse_meeting(i,section_object,meeting_count)
+        meeting_count+=1
+
+def parse_meeting(meetingsData,section_object,meeting_count):
```
gannonprudhomme:
```suggestion
def parse_meeting(meetings_data,section_object,meeting_count):
```

Python uses snake style naming conventions rather than camel case. Pylint should find these errors.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,106 @@
...
+        return(instructor_object)
+
+def parse_course(courseData):
+    course_dept = courseData['subject']
+    course_number = courseData['courseNumber']
+    course_title = courseData['courseTitle']
+    course_term_code = courseData['term']
+    course_id = generate_course_id(course_dept, course_number, course_term_code)#primary key
+    course_desc = ''#long description-could not find in data
+    course_core_curriculum = ''#to be implented later
```
gannonprudhomme: We have to get description from an entirely different URL (and we don't have any plans of using it rn), and we're not going to be using core curriciulum for now, so both of these can be removed from the `Course` model.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,106 @@
...
+    course_credit_hours = courseData['creditHourLow']
+    #creates and saves course object
+    course_object = Course(id=course_id, dept=course_dept, course_num=course_number, title=course_title, description=course_desc, core_curriculum=course_core_curriculum, credit_hours=course_credit_hours, term=course_term_code)
+    course_object.save()
+    #calls other parse fuctions
+    instructor_object = parse_instructor(courseData)
+    parse_section(courseData,instructor_object)
+    print(f'{courseData}\n')
```
gannonprudhomme: This print is fine for debugging, but should be removed before we merge this into `backend/master`

autoscheduler/scraper/models/course.py
```diff
@@ -8,20 +8,14 @@ def generate_course_id(dept: str, course_num: str, term: str):
 class Course(models.Model):
     """ Basic course model """
 
-    id = models.CharField(max_length=10, primary_key=True) # Could be dept+course_num
+    id = models.CharField(max_length=14, primary_key=True) # Could be dept+course_num
```
gannonprudhomme: This will now actually have to be 15, since course_num can be 5 characters long (i.e. LAW 7882S) plus the 6 character term code, plus the 4 character subject/dept.

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Just made a comment about this but didn't realize you originally said this. You're correct in that we have no current plans for using this data, so it can be removed.

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: For now I think it's fine if we don't as I don't think it would affect anyone (plus for minimizing distance between classes all we need is the building). We can always add it later if we find a need for it.

autoscheduler/scraper/models/section.py
```diff
@@ -12,29 +14,38 @@ class Section(models.Model):
...
     """
     id = models.BigIntegerField(primary_key=True) # id is primary key in scraped data
     crn = models.IntegerField(db_index=True)
-
     building = models.CharField(max_length=4, null=True)
     meeting_days = ArrayField(models.BooleanField(), size=7)
     start_time = models.TimeField(null=True)
     end_time = models.TimeField(null=True)
-
     meeting_type = models.CharField(max_length=3) # Meeting types: LEC, LAB, REC, INS, etc
```
gannonprudhomme: Also, `meeting_type` has to be 4 characters long, b/c of PRAC (practical?)

autoscheduler/scraper/management/commands/scrape_courses.py
ryan-conn: It probably should, seeing as on our website we'll probably want to display the credit hours for a course overall (maybe in the course search, definitely when viewing schedule overviews). As for how we would implement this, whenever adding a section, we could update the corresponding course's minimum credits.

gannonprudhomme:
> I think we'll want to move the functions you created in section.py into scrape_courses.py, as they won't be used except when scraping courses

@ryan-conn I agree with this, except that I think `generate_meeting_id` should be left in `section.py`, as it's directly related to the field model, since it describes what the primary key actually is

gannonprudhomme:
So I must have incorrectly rebased a bit ago as there are duplicate commits that you can see here: 
chrome_x3F6cY3i73

So I'm going to rebase off of `backend/master` then just cherry-pick the commits here into it, then force push.

gannonprudhomme:
Force pushed, so make sure you run `git fetch` and `git reset --hard origin/backend/course-scraping-parse-courses` before making any commits. If there ends up being a merge commit after this, it's because the commit author didn't `git reset`.

gannonprudhomme:
Also just realized this, but a `Section` needs a crn so we can send it to the frontend. In that same vein, `Meeting` shouldn't have a crn, since it's unique to the section rather than each meeting.

autoscheduler/scraper/models/section.py
```diff
@@ -12,29 +14,38 @@ class Section(models.Model):
...
     current_enrollment = models.IntegerField()
     instructor = models.ForeignKey('Instructor', on_delete=models.CASCADE)
 
     class Meta:
         db_table = "sections"
 
+def generate_meeting_id(section_num: str, meetings_count: str):
+    """ Generates the meeting id in the form of {section_num}{meetings_count}"""
+
+    return "".join((section_num, meetings_count))
```
gannonprudhomme: This needs to be the section's `id` (not its crn) + `meeting_count`. Since most section numbers start at 500, this will very quickly run into intersections with other data.

gannonprudhomme:
Added `__init__.py`'s so `pylint` will lint the sub-folders.

firejake308:
Is this command supposed to scrape all of the meetings for each of the sections as well? Because I see several sections where I know that the class has both a lecture and a lab, but digging around in `psql` after running `scrape_courses`, I only see either the lecture or the lab, and some of the others don't have any meetings at all

gannonprudhomme:
@ThatJuanGuy Do you have an ETA for when you'll be able to finish this up?

ThatJuanGuy:
Today

ThatJuanGuy:
Regarding the missing doc string, I've assigned one to everything except for classes and get_departments. Are doc strings necessary for the 

class Meta: 
     db_table = tablename

and if so what do I put in there?

gannonprudhomme:
> Regarding the missing doc string, I've assigned one to everything except for classes and get_departments. Are doc strings necessary for the
> 
> class Meta:
> db_table = tablename
> 
> and if so what do I put in there?

No they are not. If you run pylint with the `--load-plugins pylint_django` argument, then you won't get this error.

gannonprudhomme:
> Force pushed, so make sure you run `git fetch` and `git reset --hard origin/backend/course-scraping-parse-courses` before making any commits. If there ends up being a merge commit after this, it's because the commit author didn't `git reset`.

Per this previous comment, I'm gonna need to clean up this commit history in a bit since there are merge commits. If you press the `Commits` tab of this PR, you can see a bunch of duplicate commits, hence the need for the `git reset`. 

That being said, this is like the exact reason why we (I) shouldn't rebase a branch that others are working on, which I'm gonna try to have us all avoid in the future.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,81 +1,108 @@
...
 from scraper.banner_requests import BannerRequests
 from scraper.models import Course, Instructor, Section, Meeting
 from scraper.models.course import generate_course_id
-from scraper.models.section import generate_meeting_id, generate_meeting_time
+from scraper.models.section import generate_meeting_id
 
-def parse_section(courseData,instructor_object):
+def generate_meeting_time(string_time: str):
+    """ Generates a meeting time from a string in format hhmm.
+            ex) 1245 = 12:45am. 1830 = 6:30 pm."""
```
gannonprudhomme: I think `convert_meeting_time` would be a more applicable name, as the function is given a string and _converts_ it to a `datetime.time` object.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,81 +1,108 @@
...
+    if string_time == "":
+        return None
+    hour = int(string_time[0:2])
+    minute = int(string_time[2:4])
+    meeting_time = datetime.time(hour, minute)
+
+    return meeting_time
+
+def generate_meeting_days(meetings_data):
```
gannonprudhomme: Similar to the above comment, I think a better name for this would be `parse_meeting_days`, since it goes through all of the `day` fields and creates an object out of it (which is what the other `parse_` functions do, expect they make actual Models)

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,81 +1,108 @@
...
+        id=course_id, dept=course_dept, course_num=course_number, title=course_title,
+        credit_hours=course_credit_hours, term=course_term_code)
     course_object.save()
     #calls other parse fuctions
-    instructor_object = parse_instructor(courseData)
-    parse_section(courseData,instructor_object)
-    print(f'{courseData}\n')
+    instructor_object = parse_instructor(course_data)
+    parse_section(course_data, instructor_object)
+    print(f'{course_data}\n')
```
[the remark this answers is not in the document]
ThatJuanGuy: Will do. I thought I had already removed it. My bad.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,133 @@
...
+    def handle(self, *args, **options):
+        banner = BannerRequests(options['term'])
+        loop = asyncio.get_event_loop()
+        # the next line should be used in production, but we should only use two or three
+        # departments when testing to make the code run way faster
+        # once department scraping is done, we can replace get_department_names
+        # with just getting each of the departments for the term from our database
+        # depts = get_department_names(banner)
+        depts = ['CSCE'] # test with one department, change this if you want
+        json = loop.run_until_complete(banner.search(depts, 10)) # only get a few courses
```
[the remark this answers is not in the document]
ThatJuanGuy: Ok.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,133 @@
...
+        #start = time.time()
+        for course_list in json:
+            for course in course_list:
+                parse_course(course)
+        #finish = time.time()
+        #elapsed_time = finish - start
+        #print(f'Finished scraping in {elapsed_time:.2f} seconds')
```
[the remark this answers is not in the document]
ThatJuanGuy: I removed them with the print statement because I assumed the use was for debugging and testing stuff.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,81 +1,108 @@
...
+        min_credits=section_min_credits, max_credits=section_max_credits,
+        max_enrollment=section_max_enrollment,
+        current_enrollment=section_current_enrollment, instructor=section_instructor)
     section_object.save()
     #calls parse meeting for each meeting time in the section
-    meeting_count=0
-    for i in courseData['meetingsFaculty']:
-        parse_meeting(i,section_object,meeting_count)
-        meeting_count+=1
+    for i, meetings_data in enumerate(course_data['meetingsFaculty']):
```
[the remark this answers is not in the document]
ThatJuanGuy: Thanks. I didn't either. I learned about that and list comprehension from Ryan's comment a while back so credit goes to him for that.

autoscheduler/scraper/management/commands/scrape_courses.py
ThatJuanGuy: That makes sense. Because there is no new data being created, I can see why generate instead of convert could be misleading. Will change.

autoscheduler/scraper/management/commands/scrape_courses.py
ThatJuanGuy: See above comment.

autoscheduler/scraper/management/commands/scrape_courses.py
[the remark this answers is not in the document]
gannonprudhomme: Nah, I like to have it for comparing speeds just in general, as the end-result for scrape courses (and really scraping in general) should have it so that it scrapes _all_ of the courses going back a few years (needed for grade distributions), which will end up taking a few minutes.

autoscheduler/scraper/management/commands/scrape_courses.py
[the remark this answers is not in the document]
ThatJuanGuy: So some courses don't have an instructor name. Since instructor name is the primary key for instructor, it can't be null. This causes the program to crash. I was thinking we could either A) find some other primary key for instructor or B) (probably a better idea) change the parse course function to account for there being no instructor if there is none. Any advice on how to proceed?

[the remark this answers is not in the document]
gannonprudhomme: You can just make the instructor field in `Section` to be able to be null. Since `parse_instructor` only returns in the for-loop, and in the case that there are no instructors the for loop won't ever run, and thus the object it returns will be `None`.

gannonprudhomme:
Seems to be working completely for the Fall 2019 semester, can you double check it? @ryan-conn

gannonprudhomme:
Likewise, hopefully either tomorrow or Tuesday I'll fully implement #102 / #113  then will make it a full PR

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,133 @@
...
+        parser.add_argument('--term', type=str, required=False, default="201931")
+
+    def handle(self, *args, **options):
+        banner = BannerRequests(options['term'])
+        loop = asyncio.get_event_loop()
+        # the next line should be used in production, but we should only use two or three
+        # departments when testing to make the code run way faster
+        # once department scraping is done, we can replace get_department_names
+        # with just getting each of the departments for the term from our database
+        depts = get_department_names(banner)
```
firejake308: Correct me if I'm wrong, but in order to get this to work, I had to change this line to `depts = get_department_names(options['term'])` Since the argument to `get_department_names` should be a term code, not a `BannerRequests` object.

firejake308: Correct me if I'm wrong, but in order to get this to work, I had to change this line to `depts = get_department_names(options['term'])` Since the argument to `get_department_names` should be a term code, not a `BannerRequests` object.

gannonprudhomme:
Is this ready to be reviewed? If so, request us for review again

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,135 @@
...
+    return meeting_time
+
+def parse_meeting_days(meetings_data):
+    """generates a list of seven booleans where each corresponds to a day in the week.
+    The first is Monday and the last is Sunday. If true, there is class that day"""
+    meeting_class_days = [
+        'monday', 'tuesday', 'wednesday', 'thursday', 'friday', 'saturday', 'sunday']
+    attend_days = [meetings_data['meetingTime'][day] for day in meeting_class_days]
+
+    return attend_days
```
ryan-conn: You can just immediately return the list comprehension instead of assigning it to `attend_days` then returning

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -1,6 +1,135 @@
...
+
+def parse_section(course_data, instructor_object):
+    """called from parse_course. puts section data in database. calls parse meeting"""
```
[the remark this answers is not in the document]
gannonprudhomme: I'm gonna make a PR fixing stuff like this after this is approved (and #124 is merged into this) just btw

gannonprudhomme:
@ThatJuanGuy Will you be able make these changes anytime soon? If not let me know and I'll make a PR with these additions to merge into this branch. I plan on making one anyways for some nit picks / stylistic changes that would be a bother to manually request changes for.

gannonprudhomme:
After #124 is merged, I'll fix the commits in this branch (probs by cherry-picking), re-request your review @ryan-conn, then merge this into `backend/master` then into `master`

gannonprudhomme:
First, I rebased this onto `backend/master` so we would have the reset migrations fix for Actions, then I cherry-picked all of the necessary commits so there weren't any merges / duplicate commits. Should be good to go now

gannonprudhomme:
FINALLY MUAHAHAHAH

PR #95, opened 2019-11-19
Bring meeting card into schedule

firejake308:
I am ready to implement the actual meeting card into the Schedule component. I will merge in the meeting card whenever you day it is ready.


autoscheduler/frontend/src/tests/MeetingCard.test.tsx
```diff
@@ -0,0 +1,51 @@
...
+  crn: 123456,
+  building: 'HRBB',
+  meetingDays: new Array(7).fill(true),
+  startTimeHours: 8,
+  startTimeMinutes: 0,
+  endTimeHours: 18,
+  endTimeMinutes: 50,
+  meetingType: MeetingType.LEC,
+  section: testSection,
+};
```
ryan-conn: Is there a reason `testSection`'s `instructor` uses the constructor, but `testSection` itself and `testMeeting` don't? Honestly I'm not sure if the best practice would be to use them or not, so I'm wondering what your thoughts are

firejake308: Oh. Huh. I didn't even think about that; I just kinda wrote it. I think it would probably be best to convert all of them to using the constructor, since that most closely replicates how the code will actually be used, so I'll fix that, then merge it in.

firejake308:
I merged it in locally. The result will be pushed to `frontend/schedule`

PR #96, opened 2019-11-19
Let React handle routing

firejake308:
If the path doesn't match `/admin/`, then React-router will handle routing. This is going straight to master because it's backend code (in a `.py` file) that affects mostly the frontend

### DIFF
--- a/autoscheduler/frontend/urls.py
+++ b/autoscheduler/frontend/urls.py
@@ -1,10 +1,8 @@
-from django.urls import path, re_path
+from django.urls import re_path
 
 from . import views
 
 urlpatterns = [
-    # match the root
-    path('', views.index),
-    # match all other pages
-    re_path(r'^(?:.*)/?$', views.index),
+    # match literally everything
+    re_path(r'.*', views.index),
 ]
--- a/autoscheduler/frontend/urls.py
+++ b/autoscheduler/frontend/urls.py
@@ -1,7 +1,10 @@
-from django.urls import path
+from django.urls import path, re_path
 
 from . import views
 
 urlpatterns = [
+    # match the root
     path('', views.index),
+    # match all other pages
+    re_path(r'^(?:.*)/?$', views.index),
 ]
--- a/autoscheduler/frontend/urls.py
+++ b/autoscheduler/frontend/urls.py
@@ -4,5 +4,5 @@ from . import views
 
 urlpatterns = [
     # match literally everything
-    re_path(r'.*', views.index),
+    re_path('.*', views.index),
 ]


autoscheduler/frontend/urls.py
```diff
@@ -1,7 +1,10 @@
-from django.urls import path
+from django.urls import path, re_path
 
 from . import views
 
 urlpatterns = [
+    # match the root
     path('', views.index),
+    # match all other pages
+    re_path(r'^(?:.*)/?$', views.index),
```
ryan-conn: I believe we could replace this with 
```
    # match all pages other than admin
    re_path(r'^(?!admin/).*', views.index),
```

gannonprudhomme: Won't we need `/api/` routes to be routed to `/scraper/`?

ryan-conn: This can be achieved with this line instead:
`re_path(r'^(?![(admin|api)]/).*', views.index),`
and add whatever other paths we don't want to match in the or

firejake308: I think that the `/api` oaths can be handled in `autoscheduler/urls.py`, the same way that `/admin` routes are handles currently

gannonprudhomme:
Oh yeah you’re totally right, completely forgot this was frontend/urls.py.

autoscheduler/frontend/urls.py
gannonprudhomme: Oh yeah you're totally right, didn't really think about that this was `frontend/views.py`.

gannonprudhomme:
> This is going straight to master because it's backend code (in a .py file) that affects mostly the frontend

Honestly, since this change doesn't need to be in the backend branches anytime soon, I think it's fine to merge this with `frontend/master` rather than `master. Also, since this branch is just `routing` rather than `frontend/routing`, we won't get a Danger error saying no backend changes in the frontend.

ryan-conn:
Actually, can't we have nothing but `re_path('.*', views.index),` in `frontend/urls.py` and just make sure it's listed last in `autoscheduler/urls.py`? Django goes through urls in the order they're listed in, so just using frontend as the default if it doesn't match with anything else should work fine.

firejake308:
> Actually, can't we have nothing but `re_path('.*', views.index),` in `frontend/urls.py` and just make sure it's listed last in `autoscheduler/urls.py`? Django goes through urls in the order they're listed in, so just using frontend as the default if it doesn't match with anything else should work fine.

You're suggesting replacing both lines with that one line? If so, done. Tbh I copied the regex from StackOverflow cuz I was too tired to figure out what all the symbols meant. I tested yours and it works just fine. Pushing and re-requesting review

firejake308:
Oh, lo, I thought the r stood for regex because I've only ever seen it used with regex

gannonprudhomme:
> Oh, lol, I thought the r stood for regex because I've only ever seen it used with regex

Honestly I assumed that too lmao

PR #100, opened 2019-11-20
Fixed Pylint errors + Backend action steps

gannonprudhomme:
Corrects the pylint errors in various files. Also makes the `Backend CI / Testing` step use `pytest` rather than Django's tests (as we have non-Django tests). Also corrects `pytest` to use the correct settings file, `autoscheduler.settings.base`.

Take a moment to look at the beautiful green check 😎


.github/workflows/backend-workflow.yml
```diff
@@ -23,4 +23,4 @@ jobs:
         uses: cclauss/GitHub-Action-for-pylint@master
         with:
         # Installs Django and pylint-django, then runs pylint on all of the files in autoscheduler/
-          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/** --load-plugins pylint_django
+          args: pip install -r .github/lint-requirements.txt ; pylint autoscheduler/scraper autoscheduler/scheduler autoscheduler/autoscheduler autoscheduler/manage.py --load-plugins pylint_django
```
gannonprudhomme: Doing `autoscheduler/**` lints non-Python files, such as `pytest.ini` or the README's, so we have to specify each individual directory.

We also can't just do `autoscheduler/`, since it doesn't have an `__init__.py`

autoscheduler/autoscheduler/settings/base.py
```diff
@@ -76,7 +76,8 @@
 
 AUTH_PASSWORD_VALIDATORS = [
     {
-        'NAME': 'django.contrib.auth.password_validation.UserAttributeSimilarityValidator',
```
gannonprudhomme: Line was too long, so split up

autoscheduler/scraper/models/section.py
```diff
@@ -19,7 +18,7 @@ class Section(models.Model):
 
     max_enrollment = models.IntegerField()
     current_enrollment = models.IntegerField()
-    instructor = models.ForeignKey(Instructor, on_delete=models.CASCADE)
+    instructor = models.ForeignKey('Instructor', on_delete=models.CASCADE)
```
gannonprudhomme: So the `from scraper.models import Instructor` was giving me a `cyclic-import` error only in Actions(not locally), but we shouldn't even have to import Instructor anyways as we can just specify the name of the model instead.

docker-entrypoint.sh
```diff
@@ -1,11 +1,17 @@
...
+
 echo "Creating migrations"
 python3 manage.py makemigrations --settings=autoscheduler.settings.docker scraper
 
 echo "Migrating"
 python3 manage.py migrate --settings=autoscheduler.settings.docker scraper
 
 echo "Starting tests"
-python3 manage.py test --settings=autoscheduler.settings.docker
+export DJANGO_SETTINGS_MODULE=autoscheduler.settings.docker
```
gannonprudhomme: This export overrides the `DJANGO_SETTINGS_MODULE=autoscheduler.settings.base` line in `pytest.ini`, so we don't have to make another `pytest.ini` only for Docker to use.

firejake308:
Oh my god it's a green check mark

firejake308:
This is mostly Python stuff, and while it seems legit to me, I'll let @ryan-conn review it before we merge it in. Also, that we he can see the beautiful green check mark

gannonprudhomme:
Squashed the last two commits together. Once checks run, will rebase & merge.

PR #107, opened 2019-11-29
Landing Page Fixes

firejake308:
I've fixed various mistakes in the landing page:

- eliminated an error due to Material-UI only accepting 'top' as the anchorOrigin.vertical prop
- eliminated an error due to having `Typography` inside of `Box`
- menu icon is now showing
- `useStyle` now refers to the app theme, as it is supposed to

I would like @f4alt to review these changes for understanding before merging them into the main feature branch for `frontend-welcomepage`

### DIFF
--- a/autoscheduler/frontend/src/components/LandingPage/HelpText.tsx
+++ b/autoscheduler/frontend/src/components/LandingPage/HelpText.tsx
@@ -21,25 +21,25 @@ const HelpText: React.SFC = function App() {
             Guidelines
           </Box>
         </Typography>
-        <Typography
-          variant="body1"
-          align="center"
-        >
-          <Box padding={2}>
-          Somebody once told me the world is gonna roll me
-          I ain&apos;t the sharpest tool in the shed
-          She was looking kind of dumb with her finger and her thumb
-          In the shape of an &quot;L&quot; on her forehead
-          Well the years start coming and they don&apos;t stop coming
-          Fed to the rules and I hit the ground running
-          Somebody once told me the world is gonna roll me
-          I ain&apos;t the sharpest tool in the shed
-          She was looking kind of dumb with her finger and her thumb
-          In the shape of an &quot;L&quot; on her forehead
-          Well the years start coming and they don&apos;t stop coming
-          Fed to the rules and I hit the ground running
-          </Box>
-        </Typography>
+        <Box padding={2}>
+          <Typography
+            variant="body1"
+            align="center"
+          >
+            Somebody once told me the world is gonna roll me
+            I ain&apos;t the sharpest tool in the shed
+            She was looking kind of dumb with her finger and her thumb
+            In the shape of an &quot;L&quot; on her forehead
+            Well the years start coming and they don&apos;t stop coming
+            Fed to the rules and I hit the ground running
+            Somebody once told me the world is gonna roll me
+            I ain&apos;t the sharpest tool in the shed
+            She was looking kind of dumb with her finger and her thumb
+            In the shape of an &quot;L&quot; on her forehead
+            Well the years start coming and they don&apos;t stop coming
+            Fed to the rules and I hit the ground running
+          </Typography>
+        </Box>
       </Paper>
     </Grid>
   );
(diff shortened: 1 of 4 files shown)


autoscheduler/frontend/src/components/LandingPage/HelpText.tsx
```diff
@@ -21,25 +21,25 @@ const HelpText: React.SFC = function App() {
             Guidelines
           </Box>
         </Typography>
-        <Typography
-          variant="body1"
-          align="center"
-        >
-          <Box padding={2}>
```
firejake308: Typography is rendered as a `<p>` component, so it can't have a `<div>` inside it

autoscheduler/frontend/src/components/LandingPage/SelectTerm.tsx
```diff
@@ -50,7 +52,7 @@ const SelectTerm: React.SFC = function App() {
           id="long-menu"
           anchorEl={anchorEl}
           anchorOrigin={{
-            vertical: 'bottom',
+            vertical: 'top',
```
firejake308: Honestly, I had to go digging through the Material-UI source code to fix this one, so I don't expect you to learn much from it. It's just weird.

autoscheduler/frontend/src/components/LandingPage/SelectTerm.tsx
```diff
@@ -72,8 +74,8 @@ const SelectTerm: React.SFC = function App() {
           {options.map((option) => (
             <MenuItem
               key={option}
-              selected={option === 'Pyxis'}
-              onClick={handleClose}
+              selected={option === selectedTerm}
```
firejake308: Pyxis was just a dummy value. By hooking it up to state, I've made it so that the option that matched the selected term will be highlighted

autoscheduler/frontend/src/components/LandingPage/SelectTerm.tsx
```diff
@@ -72,8 +74,8 @@ const SelectTerm: React.SFC = function App() {
...
+              onClick={(): void => handleClose(option)}
```
firejake308: I had to make this into an arrow function so that I could pass in the option as an argument. If you try `onClick={handleClose(option)}`, you'll get an error because the compiler will call `handleClose(option)`, get the return value `void` and then try to set `onClick={void}`

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -16,7 +18,7 @@ const useStyles = makeStyles((theme) => ({
 }));
 
 const NavBar: React.SFC = function App() {
-  const classes = useStyles(useStyles);
+  const classes = useStyles(appTheme);
```
firejake308: This is now using the theme defined in `theme.ts`, so if we make changes to that theme later, they'll be passed down to the NavBar

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -30,7 +32,9 @@ const NavBar: React.SFC = function App() {
             className={classes.menuButton}
             color="inherit"
             aria-label="menu"
-          />
+          >
+            <MenuIcon />
```
firejake308: You have to specify the icon that you want inside the button. I imported in the MenuIcon at the top of the file.

f4alt: Do we want this?

firejake308: Now that I think about it, I don't know if a navigation drawer is really necessary in this app. The only 2 routes are the landing page and the schedule page, so we might just delete the `<IconButton>` altogether. I'll let Gannon confirm which direction we want to go, though.

gannonprudhomme: Yeah honestly I don't think we'll need a navigation drawer, it should be that selecting a term brings you to the schedule view most likely.

firejake308:
I removed the menu button and then fixed a bug where the `SelectTerm` popover wouldn't disappear after you chose a semester. I also added tests to ensure that the bug does not return in the future. @f4alt If you're good with these changes, feel free to merge whenever you want

PR #108, opened 2019-11-29
Implement Redux for Schedule

firejake308:
Attempting to rebase the `frontend/redux` branch onto the updated `frontend/schedule` proved to be very difficult, so instead, I created an entirely new branch and cherry-picked the commits from the old branch. It should still do the same thing: display the schedule, but pulled from Redux rather than props


gannonprudhomme:
Since I'm honestly clueless in regards to Redux, I'm going to try to implement this myself locally, then once I have my bearings I'll review this.

autoscheduler/frontend/src/components/App/testMeetings.ts
```diff
@@ -0,0 +1,106 @@
...
+  id: 123456,
+  subject: 'SUBJ',
+  courseNum: 234,
+  sectionNum: 500,
+  minCredits: 3,
+  maxCredits: null,
+  currentEnrollment: 56,
+  instructor: new Instructor({
+    id: 123456,
+    name: 'Aakash Tyagi',
```
gannonprudhomme: Going to be an upcoming change in #94, but the instructor's id/primary key is now their name (recent Banner update changed the data), so we'll want to update the Instructor type accordingly

gannonprudhomme: Going to be an upcoming change in #94, but the instructor's id/primary key is now their name (recent Banner update changed the data), so we'll want to update the Instructor type accordingly

autoscheduler/frontend/src/tests/Redux.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  // act
+  store.dispatch(addMeeting(testMeeting1));
+  store.dispatch(replaceMeetings([testMeeting2]));
+
+  // assert
+  expect(store.getState().meetings).toHaveLength(1);
+  expect(store.getState().meetings).toContain(testMeeting2);
+  expect(store.getState().meetings).not.toContain(testMeeting1);
+});
```
gannonprudhomme: Since I assume one of the intended behaviors with the `replaceMeetings` action is to completely replace all of the current meetings regardless of how many there are, there should also be a test which tests that the action works with multiple meetings (i.e. insert 2, replace w/ 1)

autoscheduler/frontend/src/redux/actions.ts
```diff
@@ -3,3 +3,28 @@
...
+    meeting,
+  };
+}
+
+export function replaceMeetings(meetings: Meeting[]): MultiMeetingAction {
+  return {
+    type: REPLACE_MEETINGS,
+    meetings,
+  };
+}
```
gannonprudhomme: While I think the other two are self explanatory, I think there should be a description for this, and mention when this would be used / what situations. I assume it's for quickly changing out meetings when schedules are swapped?

autoscheduler/frontend/src/index.tsx
```diff
@@ -1,9 +1,27 @@
...
+interface MyWindow {
+  __REDUX_DEVTOOLS_EXTENSION__: () => StoreEnhancer;
+}
+// when converting to production, remove this interface, the second argument
+// to createStore, and the ESLint overrides at the top
+
+// initialize Redux store
+const store = createStore(autoSchedulerReducer,
+  (window as MyWindow & Window & typeof globalThis).__REDUX_DEVTOOLS_EXTENSION__
+  && (window as MyWindow & Window & typeof globalThis).__REDUX_DEVTOOLS_EXTENSION__());
```
gannonprudhomme: Assuming that this for the redux devtools browser extension, add a comment mentioning it cause it's kind of wonky lol

autoscheduler/frontend/src/tests/Redux.test.ts
ryan-conn: I agree with this, since the function replaces the whole array for `meetings`

autoscheduler/frontend/src/tests/Redux.test.ts
```diff
@@ -0,0 +1,91 @@
...
+  const store = createStore(autoSchedulerReducer);
+
+  // act
+  store.dispatch(addMeeting(testMeeting1));
+  store.dispatch(replaceMeetings([testMeeting2]));
+
+  // assert
+  expect(store.getState().meetings).toHaveLength(1);
+  expect(store.getState().meetings).toContain(testMeeting2);
+  expect(store.getState().meetings).not.toContain(testMeeting1);
```
ryan-conn: I think we should replace this with
```
  expect(store.getState().meetings).toEqual([testMeeting2]);
```

gannonprudhomme: Yeah I agree, that statement does all of the three in one.

autoscheduler/frontend/src/redux/actions.ts
firejake308: Yes, essentially.

firejake308: Yes

firejake308:
Re-requesting review for one final look-over before we merge this, to make sure that my fixes are adequate

Issue #106, opened 2019-11-25
Implement grade distribution scraping


gannonprudhomme:
I need to look into how we can go about scraping grade distributions, as its a major feature of Backend v0.2.

I'm mainly going to base it on Gabriel Britain's [Good Bull Schedules](https://github.com/good-bull-schedules/Good-Bull-Schedules/blob/master/goodbullschedules/scraper/management/commands/scrape_grades.py), as scraping & parsing the grade distributions shouldn't have changed in the most recent Banner update (whereas course search did)

Also includes making the `Grades` model.

You can check out progress for it in `backend/scraper/scrape_grades`.

gannonprudhomme:
Just a thought, but it may be useful to also show the average GPA of the course (for the previous semester or over all time). Not sure where it would be display thought.

 The grade distribution PDF's have them included, but calculating them should be easy as well.

gannonprudhomme:
I can't really run the `scrape_grades` command without pushing a bunch of error fixes for `scrape_courses.py`, (as the `Grades` model requires its according `Section` to be in the database), so I won't probably pushing it until #94 has the required changes.

gannonprudhomme:
I have this functioning basically completely locally, but again imma wait till #94 to push the rest of my changes. Plus I'd just like to wait until we have less active PR's.

Since this is a relatively big PR, I'm going to be doing these changes in sections(at least 2)
- the actual pdf parser that Gabriel Britain wrote (which I slightly cleaned up)
- the Grade model
- the scrape grades command + the rest of it

PR #116, opened 2019-12-15
Implement Course Select Card

### LINKED ISSUE #109: Course Selection Card
Make this card into a React component

course_selection_card

firejake308:
I have more or less implemented the course select card. I believe it has full functionality at this point, and all that remains is fixing some spacing issues and writing tests to ensure that multiple meetings/sections are handled properly. i'll leave it as a draft until the spacing and tests are fixed.

Closes #109


gannonprudhomme:
Also, I think that the meeting place should be showed last, and have the meeting days be shown in its place, since the building is honestly like the last thing most people consider when picking classes (if at all)

firejake308:
>Also, should we do the displaying of grade distributions in this, or wait until #106 is completed?

I get the feeling that this PR is gonna get a little large if we add grade distros now, so I think waiting would be better

firejake308:
> I think that the meeting place should be showed last, and have the meeting days be shown in its place,

Agreed. Latest commit replaces meeting place with meeting days. That was an oopsie on my part for trying to remember the prototype without actually looking at it.

> 1. ... the labels (options for basic select, professor names for section select) for basic/section select are different colors/sizes and the top spacing is different. I think using the same styling as "Options" does for professor names would look better

I've changed to the `<ListSubheader>` component provided by the Material-UI library. Do you think this is better, or should the form labels be uniform with the subheaders? Also, top spacing was reduced from 8px to 4px

> 2. Everything except the search itself should probably be hidden until a course is selected (another solution would be to show a message like "Select a course to pick sections" if the user clicks on section customization instead of just being blank)

It felt weird having a bunch of elements pop out of nowhere when I hid everything except the search, so I elected to show a message in in place of `<SectionSelect>`. If that doesn't feel good, we could try it the first way, but we'd need some better CSS transitions

> 3. I think section info should be styled differently from the section name, I think making the text smaller and grey would look nice

I'm not so sure about that. Take a look at it yourself, but I think it would be better if we made the text black and normal-sized. Having different sizes feels discontinuous, and when the details are gray, there's just too much grey and not enough black. We could try something like bolding the section number if you want to make that pop out more. If you think it's fine, though, we can keep it as it is.

> 4. I like that you can click anywhere in a section's information to select an option, but it should be more clear that you can (perhaps darken the background color so the user can tell what section they're hovering over, and do the same for basic select). The cursor should probably change to a pointer as well when hovering over a section. In addition, you probably shouldn't be able to select any of the text for section info/name, since the user will be clicking it to select/deselect sections (same applies for honors only/web only text, and the text for collapse/remove at the top).

Accomplished all of these at once by adding the `button` prop to `<ListItem>`. Highlighting doesn't work on Firefox but honestly I don't think that's a big deal.

> 5. Clicking honors only/web only changes the font color of "options" between maroon and grey depending on focus but it's a little wonky when clicking multiple times, setting focused=false on the FormLabel will make it always grey

ok

Next commit will be making the card collapsible, followed by testing. Let me know if there is anything else I'm forgetting.

ryan-conn:
>
> I've changed to the `<ListSubheader>` component provided by the Material-UI library. Do you think this is better, or should the form labels be uniform with the subheaders? Also, top spacing was reduced from 8px to 4px

I like it how it is now, the spacing looks better now too

> It felt weird having a bunch of elements pop out of nowhere when I hid everything except the search, so I elected to show a message in in place of `<SectionSelect>`. If that doesn't feel good, we could try it the first way, but we'd need some better CSS transitions

I think showing text if a section isn't selected is probably better, as it looks nicer and is more informative, so I'm fine with keeping it as it is now.
> > 1. I think section info should be styled differently from the section name, I think making the text smaller and grey would look nice
> 
> I'm not so sure about that. Take a look at it yourself, but I think it would be better if we made the text black and normal-sized. Having different sizes feels discontinuous, and when the details are gray, there's just too much grey and not enough black. We could try something like bolding the section number if you want to make that pop out more. If you think it's fine, though, we can keep it as it is.

I think we should condense the section details (currently, `line-height` is 1.5 so there's a lot of empty space, I changed the value a little bit and 1 or 1.1 look better imo), if we were to do that we should keep the smaller text size, but I agree though that leaving the color as black looks good. Here are a few screenshots for comparison:
`line-height` 1.1, 0.8rem font:
image
`line-height` 1.1, normal font size:
image
`line-height` 1.5, 0.8rem font:
image
`line-height` 1.5, normal font size:
image

Personally I think the top looks best, but it's your call.

> > 1. Clicking honors only/web only changes the font color of "options" between maroon and grey depending on focus but it's a little wonky when clicking multiple times, setting focused=false on the FormLabel will make it always grey
> 
> ok

I still see this happen on the new version, did you forget to change it or did you want to leave it the same?

> Next commit will be making the card collapsible, followed by testing. Let me know if there is anything else I'm forgetting.

Honors/web only select should be a `<ListItem>` as well so it looks the same as selecting a section

firejake308:
@ryan-conn After polling a "consumer focus group" (my little sister), I elected `line-height: 1.1; font-size: 1.0rem;`. Also, I converted the `<BasicSelect>` to use `ListItem`s instead of `FormControl`s. 

Finally, the last commit I uploaded makes the card collapsible. I've tested it visually, but mocking the Redux is proving to be a bit of a challenge, so I'll have to spend some time on that before I get the test passing.

firejake308:
With testing done, I've got everything I can think of. Let me know if y'all notice anything

gannonprudhomme:
Now that I think about it, shouldn't `Section` have an honors and web flag? That way when we select Honors/Web only, we can filter out the sections shown to only those. That being said, I wouldn't mind if we put that functionality in a separate PR, since this is already big enough.

In that same vein, do you think it's beneficial to specify that it's "Honors only" (rather than just "Honors"), or do you think it's intuitive enough seeing just Honors? (same goes for Web/Web only ofc)

gannonprudhomme:
Also, is this test error noteworthy? From `npm test`

```
(node:2692) UnhandledPromiseRejectionWarning: ReferenceError: Response is not defined
(node:2692) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:2692) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Likewise, can you run `npm audit fix`, as there's 1 moderate vulnerability in one of our packages.

firejake308:
>Now that I think about it, shouldn't Section have an honors and web flag? That way when we select Honors/Web only, we can filter out the sections shown to only those. That being said, I wouldn't mind if we put that functionality in a separate PR, since this is already big enough.

Good point, but I believe that honors sections are always 200's and web sections are always 900's, so we might be able to use that instead of adding another field. Definitely should do some research first.  If not, let's do it in a separate PR and simultaneously fix the backend types.

>In that same vein, do you think it's beneficial to specify that it's "Honors only" (rather than just "Honors"), or do you think it's intuitive enough seeing just Honors? (same goes for Web/Web only ofc)

Honors Only and Web Only make more sense than what I have right now based on how I envision them being used. Will fix.

>Also, are these test errors noteworthy? From npm test

They don't indicate anything wrong with the real code itself, but I should be able to patch them up

gannonprudhomme:
Also, could you set it up so that each course has a different set of sections? That way we can confirm that it correctly switches between the courses & updates the sections accordingly.

firejake308:
That sounds like a good thing to add to the tests. I'll add it to the to-do
list, but my laptop is broken right now, so it might take a while.

gannonprudhomme:
> That sounds like a good thing to add to the tests. I'll add it to the to-do list, but my laptop is broken right now, so it might take a while. 

Any idea when it'll be fixed? Also what happened to it? Been there and not having a laptop is annoying af, but at least it's not during the school year lol

firejake308:
Charger stopped working, I should get the new one in a day or two. Expect
it to be done some time this weekend 🤞

autoscheduler/frontend/src/tests/CourseSelectCard.test.tsx
```diff
@@ -0,0 +1,67 @@
...
+  await waitForElement(() => getByText(
+    (content, element) => ignoreInvisible(content, element, '501'),
+  ));
+  const checked2 = document.getElementsByClassName('Mui-checked').length;
+
+  // assert
+  // checks that document queries are working
+  expect(checked1).toEqual(1);
+  // checks that the card remembers how many elements were checked
+  expect(checked2).toEqual(checked1);
```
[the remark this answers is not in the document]
firejake308: Unfortunately, that doesn't quite work because class names generated for CSS modules have randomly generated hashes. So, for example, during the first render, one of the MUI elements might have a class called `PrivateSwitchBase-root-479`, but in the next render, it would be `PrivateSwitchBase-root-282`, causing the test to fail even though the content would visually appear to be the same

firejake308:
Okay, I got most of the simple stuff, but I'm probably gonna need one more day to get the test with different sections for different courses going

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -57,5 +57,6 @@ module.exports = {
         ["in", "instanceof"],
       ],
     }],
+    "@typescript-eslint/no-explicit-any": 0,
```
firejake308: There are times when you are okay with a variable having any type. Specifically, I need them to test Thunk actions in `Redux.test.ts`

autoscheduler/frontend/src/components/CourseSelectCard/CollapsedCourseCard/CollapsedCourseCard.css
```diff
@@ -0,0 +1,8 @@
+.maroon-card:global(.MuiCard-root) {
+    background-color: #500000;
+    color: white;
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    align-items: center;
+}
```
firejake308: Ah, shoot, I have to fix this

autoscheduler/frontend/src/tests/CourseSelectCard.test.tsx
```diff
@@ -65,3 +64,43 @@ test('Remembers state after collapse', async () => {
   // checks that the card remembers how many elements were checked
   expect(checked2).toEqual(checked1);
 });
+
+test('Changes sections in response to changing course', async () => {
+  // arrange
+  const nodeProps = Object.create(Node.prototype, {});
```
firejake308: This line and the one after it just mock some types that don't exist in the Jest environment but are required for Material-UI Popper elements. TypeScript has to ignore line 72 because otherwise, we would have to define every property expected in a `Range` object, but this allows us to define only the ones that we need.

autoscheduler/frontend/src/tests/CourseSelectCard.test.tsx
```diff
@@ -65,3 +64,43 @@ test('Remembers state after collapse', async () => {
...
+  // fill in course
+  const courseEntry = getByLabelText('Course') as HTMLInputElement;
+  act(() => { fireEvent.click(courseEntry); });
+  act(() => { fireEvent.click(getByText('CSCE 121')); });
+  // switch to section view
+  act(() => { fireEvent.click(getByText('Section')); });
+  const course1Sections = (await waitForElement(() => getAllByText(/50\d/))).length;
+
+  // change course and read sections again
+  act(() => { fireEvent.change(courseEntry, { target: { value: 'MATH 15' } }); });
```
firejake308: The reason this commit took so long is that initially, I was trying to use `fireEvent.click` and `fireEvent.keyPress` to simulate user input. However, in Jest, clicks and key presses don't actually fire a `change` event, and because the autocomplete is a controlled input, its value will not update until it receives a `change` event. Good to keep in mind if you're writing frontend tests in the future.

autoscheduler/frontend/src/tests/Redux.test.ts
```diff
@@ -8,6 +8,7 @@ import Section from '../types/Section';
 import Instructor from '../types/Instructor';
 import Meeting, { MeetingType } from '../types/Meeting';
 import { CustomizationLevel } from '../types/CourseCardOptions';
+import 'isomorphic-fetch';
```
firejake308: This eliminates those unresolved promise errors that Gannon pointed out

PR #118, opened 2019-12-26
Reset migrations before creating them

ryan-conn:
Running docker with GitHub actions doesn't clear the database, meaning that if some migrations are already applied, our backend CI can fail because the table is incorrect. This fixes the problem by first running `makemigrations --empty scraper`, meaning migrations are now created from scratch as they should.

Check out [this pull request on my fork](https://github.com/ryan-conn/Automatic-Aggie-Scheduler/pull/1) for verification that the fix works, it uses Carlos' course scraping branch as a base (where our CI failed) and applies this change over it


gannonprudhomme:
I'm downloading Docker rn on my laptop and will get this set up just so I can double check.

PR #126, opened 2020-01-16
Fix schedule background

### LINKED ISSUE #117: Schedule has no background
#83 adds a gray background color to the page that exposes the fact that the `<Schedule>` component has no background of its own. Once that PR is merged into master, the background of the schedule component needs to be fixed.

firejake308:
Fixes the transparent background of the schedule component. The only problem I will note is that the time bars now look bad on Google Chrome but fine on Firefox for some reason.

Google Chrome:

Firefox:

Closes #117


gannonprudhomme:
chrome_ckG8WJEOab

Looks fine for me

PR #128, opened 2020-01-23
Added PDF parser for scrape_grades

gannonprudhomme:
Adds the PDF parser for grade distribution scraping.

This is originally based off of Gabriel's Good Bull Schedules, but I cleaned it up a bit to work better for our purposes + some stylistic changes.

This will be merged into `backend/scraper/scrape_grades`, as well as the other parts of this overall issue. You can read more about my PR/merging plan at #106.

### Regarding tests

There are quite a few functions I didn't test, but some of them I didn't see a logical reason to. For instance, `parse_pdf` really just calls `parse_page` on each page in the PDF, and since really the only thing to test would be that it works on multiple pages (and manually writing out 2 pages of grade distributions would be annoying - although checking the count is right may be beneficial).

I may add a test for `extract_letter_grades` just so that people can see what it does more clearly in the future (i.e. to see the actual inputs + outputs). Also might add another one for `parse_page` to double check that it works for edge case inputs, like a pdf without any grades on it.

Please lmk if there's anything you think that's worth testing that I missed.


autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,144 @@
...
+    letter_grades = {
+        l: int(grade) # Map each letter to the respective amount of grad
+        for l, grade in zip(LETTERS, letter_grades)
+    }
+
+    return letter_grades
+
+
+def parse_page(
+    page_obj: PyPDF2.pdf.PageObject # FIXME I don't like this rule
```
gannonprudhomme: So pylint wants this to be indented 4 spaces more, but I like how it's indented currently, so I may try to change the rules so this is allowed. Thoughts?

firejake308: What you have now looks better than what PyLint suggests

gannonprudhomme:
Rebased to be up to date with `backend/master`

autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    with open(pdf_path, "rb") as file:
+        pdf_reader = PyPDF2.PdfFileReader(file)
+        pdf_data = []
+
+        # Iterate through all pdf pages
+        for i in range(pdf_reader.getNumPages()):
+            # Parse the individual page
+            page_data = parse_page(pdf_reader.getPage(i))
```
ryan-conn:
```suggestion
            page_data = parse_page(page)

```

autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,144 @@
...
+
+        Args:
+            pdf_path: The path to the pdf
+        Returns:
+            A list of GradeData, with each GPA calculated
+    """
+
+    with open(pdf_path, "rb") as file:
+        pdf_reader = PyPDF2.PdfFileReader(file)
+        pdf_data = []
```
ryan-conn: Since this is never modified anywhere, why does this function return an empty array? Or am I missing something

gannonprudhomme: Yeah it should be `pdf_data += page_data` after the `for grade in page_data:` loop

autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    text = sanitize_page(page_obj) # Splits the text into separate lines
+    i = 0 # The current row in the pdf we're on
+    grade_data = [] # list of GradeData objects
+
+    while i < len(text):
+        skip_count = get_pdf_skip_count(text[i])
+        old_pdf_style = skip_count[0]
+        count = skip_count[1] # The actual count of how many rows should be skipped
```
ryan-conn: I don't know how I missed this before, but you can unpack the tuple directly here instead of assigning it to `skip_count`

autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,145 @@
...
+        Returns:
+            The calculated gpa.
+    """
+    weights = [4.0, 3.0, 2.0, 1.0, 0.0] # A to F GPAs, respectively
+    grades = [letter_grades[char] for char in ["A", "B", "C", "D", "F"]]
+    num_students = sum(grades)
+
+    gpa = 0.0
+    for students_with_grade, weight in zip(grades, weights):
+        gpa += students_with_grade * weight
```
ryan-conn: If you manually find the sum in the for loop instead of using `sum(grades)`, you can avoid iterating through grades twice

autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,145 @@
...
+        for i in range(pdf_reader.numPages):
+            # Parse the individual page
+            page_data = parse_page(pdf_reader.getPage(i))
+
+            # Calculate the GPA for each grade distribution
+            for grade in page_data:
+                grade.gpa = calculate_gpa(grade.letter_grades)
+
+            # Add this page's data to the list of GradeData to be returned
+            pdf_data += page_data
```
ryan-conn:
```suggestion
            pdf_data.extend(page_data)
```
This should be **very** marginally faster and is more generic, so this way is preferable

autoscheduler/scraper/management/commands/utils/pdf_parser.py
gannonprudhomme: Ooooh I really like this, fixing.

autoscheduler/scraper/management/commands/utils/pdf_parser.py
```diff
@@ -0,0 +1,144 @@
...
+        Returns:
+            A dictionary of letter grades, with the key as the letter and the count of how
+            many people received that letter grade
+    """
+
+    abcdf_slice = slice(1, 10, 2)
+    isuqx_slice = slice(13, 18)
+    if old_pdf_style: # Old pdfs are arranged differently
+        abcdf_slice = slice(4, 9)
+        isuqx_slice = slice(10, 15)
```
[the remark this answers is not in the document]
gannonprudhomme: Imma keep it as it is as I think it'll be easier to read, but going forward I'll use this way.

gannonprudhomme:
Was saying this was 76 commits behind for some reason after I rebased onto `backend/master`, so changed the base to `backend/master` then back to `backend/scraper/scrape_grades`

Issue #125, opened 2020-01-15
Add tests for scraper API routes


gannonprudhomme:
Need to add tests for the scraper API routes before they're actually implemented.

See #115 for more details

Issue #115, opened 2019-12-10
Implement scraper API routes


gannonprudhomme:
Need to implement the scraper routes (i.e. `/api/courses`) so that we can start actually providing the frontend with data. This also means that we need to get to a point on the frontend that we're actually requesting data, or in other words, the frontend will define what routes they need, and we will then satisfy it for them.

For instance, the frontend may send a query such as `/api/courses?text="CS"` and we would return `CSCE 111, 121, ...`. Alternatively, `/api/terms` would simply return all of the current available terms that one can choose out of. 

Depending on the size of them, this may be an epic that has subtasks for each of the different routes. However I am expecting most of them to be no more than 10 lines each.

Routes that I can think of(names are TBA):
- `/api/terms` - for getting all the possible terms (to put on the landing page drop down)
  - Parameters: none
  - May want to send a fixed amount, then send more if they scroll up/down?
- `/api/course/search/` - for searching courses for the given term - only autocompletes
  - Parameters: the current search text in the course search box, as well as term?
  - Will need to decide how we filter what to send, but generally the results should be sorted with the smallest result first (i.e. given "CS", send `CSCE 121` before `CSCE 221`)
- `/api/sections/` - Give all sections for the given course
  - Parameters: Course (id?), term
  - Returns: Section data w/
    - Instructor GPA distribution (if applicable), meeting times, if there are web / honors classes
- `/api/course/` - Give all of the data about a given course?
  - Parameters: Subject, course #, term
  - Returns: Course title(i.e. "Programming Languages" for 314) & credit hours

Edit by Ryan:
When creating views for the other models, the results should be formatted the same as the expected data in the test cases I created:
`/api/terms`:
```
{
    "202031": "Fall 2020 - College Station",
    "201931": "Fall 2019 - College Station",
    "201831": "Fall 2018 - College Station",
}
```
`/api/sections`:
```json
[
    {
        "id": "10923201931",
        "crn": 10923,
        "instructor_gpa": null,
        "instructor_name": "John M. Moore",
        "honors_only": null,
        "meetings": [
            {
                "id": "4683040",
                "days": [
                    true,
                    false,
                    true,
                    false,
                    true,
                    false,
                    false
                ],
                "start_time": "11:30:00",
                "end_time": "12:20:00",
                "type": "LEC"
            },
            {
                "id": "4683041",
                "days": [
                    true,
                    false,
                    true,
                    false,
                    false,
                    false,
                    false
                ],
[...]

PR #134, opened 2020-02-05
Added scraper API tests

ryan-conn:
The tests are mostly complete, and done enough for backend to be able to start working on them so I'm going to merge it, see #125 & #115

### DIFF
--- a/autoscheduler/scraper/models/section.py
+++ b/autoscheduler/scraper/models/section.py
@@ -15,13 +15,12 @@ class Section(models.Model):
     min_credits = models.IntegerField() # Will never be null
     max_credits = models.IntegerField(null=True) # Will be null in most cases
 
-    honors_only = models.BooleanField(null=True)
-    web_only = models.BooleanField(null=True)
+    honors = models.BooleanField(null=True)
+    web = models.BooleanField(null=True)
 
     max_enrollment = models.IntegerField()
     current_enrollment = models.IntegerField()
     instructor = models.ForeignKey('Instructor', on_delete=models.CASCADE, null=True)
-    instructor_gpa = models.FloatField(null=True)
 
     class Meta:
         db_table = "sections"
--- a/autoscheduler/scraper/models/section.py
+++ b/autoscheduler/scraper/models/section.py
@@ -14,9 +14,14 @@ class Section(models.Model):
     crn = models.IntegerField(db_index=True, default=0)
     min_credits = models.IntegerField() # Will never be null
     max_credits = models.IntegerField(null=True) # Will be null in most cases
+
+    honors_only = models.BooleanField(null=True)
+    web_only = models.BooleanField(null=True)
+
     max_enrollment = models.IntegerField()
     current_enrollment = models.IntegerField()
     instructor = models.ForeignKey('Instructor', on_delete=models.CASCADE, null=True)
+    instructor_gpa = models.FloatField(null=True)
 
     class Meta:
         db_table = "sections"
--- a/autoscheduler/scraper/serializers.py
+++ b/autoscheduler/scraper/serializers.py
@@ -11,19 +11,23 @@ class CourseSerializer(serializers.ModelSerializer):
 class SectionSerializer(serializers.ModelSerializer):
     """ Serializes a section """
     instructor_name = serializers.SerializerMethodField()
-    meeting_times = serializers.SerializerMethodField()
+    meetings = serializers.SerializerMethodField()
 
     class Meta:
         model = Section
-        fields = ['instructor_gpa', 'instructor_name', 'honors_only', 'meeting_times',
+        fields = ['crn', 'instructor_gpa', 'instructor_name', 'honors_only', 'meetings',
                   'section_num', 'web_only']
 
     def get_instructor_name(self, obj):
         """ Get the name (id) of this section's instructor """
         return obj.instructor.id
 
-    def get_meeting_times(self, obj):
-        """ Get this section's meetings and return an array of their start/end times """
-        section_id = obj.id
-        meetings = Meeting.objects.filter(section__id=section_id)
-        return [[meeting.start_time, meeting.end_time] for meeting in meetings]
+    def get_meetings(self, obj):
+        """ Gets meeting information for this section """
+        meetings = Meeting.objects.filter(section__id=obj.id)
+        return {str(meeting.id): {
+            'days': meeting.meeting_days,
+            'start_time': meeting.start_time,
+            'end_time': meeting.end_time,
+            'type': meeting.meeting_type,
+        } for meeting in meetings}
--- a/autoscheduler/scraper/serializers.py
+++ b/autoscheduler/scraper/serializers.py
@@ -1,0 +1,23 @@
+from rest_framework import serializers
+from .models.course import Course
+from .models.section import Section, Meeting
+
+class CourseSerializer(serializers.ModelSerializer):
+    """ Serializes a course """
+    class Meta:
+        model = Course
+        fields = ['title', 'credit_hours']
+
+class SectionSerializer(serializers.ModelSerializer):
+    """ Serializes a section """
+    meeting_times = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Section
+        fields = ['instructor_gpa', 'honors_only', 'web_only', 'meeting_times']
+
+    def get_meeting_times(self, obj):
+        """ Get this section's meetings and return an array of their start/end times """
+        section_id = obj.id
+        meetings = Meeting.objects.filter(section__id=section_id)
+        return [[meeting.start_time, meeting.end_time] for meeting in meetings]
--- a/autoscheduler/scraper/serializers.py
+++ b/autoscheduler/scraper/serializers.py
@@ -3,31 +3,36 @@ from .models.course import Course
 from .models.section import Section, Meeting
 
 class CourseSerializer(serializers.ModelSerializer):
-    """ Serializes a course """
+    """ Serializes a course into an object with information needed by /api/course """
     class Meta:
         model = Course
         fields = ['title', 'credit_hours']
 
 class SectionSerializer(serializers.ModelSerializer):
-    """ Serializes a section """
+    """ Serializes a section into an object with information needed by /api/sections """
     instructor_name = serializers.SerializerMethodField()
     meetings = serializers.SerializerMethodField()
 
     class Meta:
         model = Section
-        fields = ['crn', 'instructor_gpa', 'instructor_name', 'honors_only', 'meetings',
-                  'section_num', 'web_only']
+        fields = ['id', 'crn', 'instructor_name', 'honors', 'meetings',
+                  'section_num', 'web']
 
-    def get_instructor_name(self, obj):
-        """ Get the name (id) of this section's instructor """
+    def get_instructor_name(self, obj): # pylint: disable=no-self-use
+        """ Get the name (id) of this section's instructor.
+            This function is used to compute the value of the instructor_name field.
+        """
         return obj.instructor.id
 
-    def get_meetings(self, obj):
-        """ Gets meeting information for this section """
+    def get_meetings(self, obj): # pylint: disable=no-self-use
+        """ Gets meeting information for this section
+            This function is used to compute the value of the meetings field.
+        """
         meetings = Meeting.objects.filter(section__id=obj.id)
-        return {str(meeting.id): {
+        return [{
+            'id': str(meeting.id),
             'days': meeting.meeting_days,
             'start_time': meeting.start_time,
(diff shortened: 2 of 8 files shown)


autoscheduler/scraper/tests/api_tests.py
```diff
@@ -0,0 +1,253 @@
...
+            meeting.save()
+
+    def test_api_terms_displays_all_terms(self):
+        """ Tests that /api/terms returns ordered list of all terms in database """
+        # Arrange
+        expected = {
+            '201831': 'Fall 2018 - College Station',
+            '201931': 'Fall 2019 - College Station',
+            '202031': 'Fall 2020 - College Station',
+        }
```
gannonprudhomme: This should be something like

```python
expected = {
    '202011': 'Spring 2020 - College Station',
    '202012': 'Spring 2020 - Galveston',
    '202013': 'Spring 2020 - Qatar'
    '202014': 'Spring 2020 - Half Year Term'
    '201931': 'Fall 2019 - College Station'
    ....
}
```

Ideally this test (and the `/api/terms` route) should update to the most recent semesters that can be taken, but this should be fine for now.

See [here](https://github.com/aggie-coding-club/Automatic-Aggie-Scheduler/blob/ddae9f45dd00e742229b60cf73d27d6d1b270e2b/autoscheduler/scraper/banner_requests.py#L33) in `banner_requests.py` to see how term formatting works.

We may want to restrict it to be only College Station / Galveston / Qatar if that's what they choose frequently using cookies or whatever, but that's something we can do at a later point.

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -0,0 +1,253 @@
...
+
+        # Assert
+        self.assertEqual(expected, serializer.data)
+
+    def test_api_course_gives_valid_response_csce(self):
+        """ Tests that /api/course?dept=CSCE&course_num=181&term=201931 gives the
+            correct output
+        """
+        # Arrange
+        expected = CourseSerializer(self.courses[0])
```
gannonprudhomme: If the `CourseSerializer` is wrong but the `api/course` route is correct, this test will fail (when it shouldn't), so we should probably change the expected to be hard-coded. Could also just make a helper function that takes in a course and returns the necessary fields.

autoscheduler/scraper/views.py
```diff
@@ -1,4 +1,14 @@
-# Once we start adding routes, uncomment this line
-# from django.shortcuts import render
+from rest_framework import generics
+from .serializers import CourseSerializer
+from .models.course import Course
 
-# Create your views here.
+class RetrieveCourseView(generics.RetrieveAPIView):
+    """ API endpoint for viewing course information """
```
gannonprudhomme:
```suggestion
    """ API endpoint for viewing course information 
         Used by /api/course
    """
```

Doesn't have to be formatted like this, but the view docstring should contain what route it's used.

autoscheduler/scraper/serializers.py
```diff
@@ -0,0 +1,33 @@
...
+
+    def get_meetings(self, obj):
+        """ Gets meeting information for this section """
+        meetings = Meeting.objects.filter(section__id=obj.id)
+        return {str(meeting.id): {
+            'days': meeting.meeting_days,
+            'start_time': meeting.start_time,
+            'end_time': meeting.end_time,
+            'type': meeting.meeting_type,
+        } for meeting in meetings}
```
gannonprudhomme: Maybe somewhere in here explain that `get_instructor_name` is used by `instructor_name` / the respective `SerializerMethodField` (same for `get_meetings`).

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -0,0 +1,253 @@
...
+                    'start_time': first_start,
+                    'end_time': first_end,
+                    'type': 'LEC',
+                },
+                '11': {
+                    'days': meeting_days,
+                    'start_time': second_start,
+                    'end_time': second_end,
+                    'type': 'LEC',
+                },
```
gannonprudhomme: Honestly I think meetings should be an array of meetings, as the ID's for them really aren't relevant (since they're just the section ID + an integer count for which meeting it is)

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -0,0 +1,253 @@
...
+        self.sections = [
+            Section(crn=12345, id='000001', subject='CSCE', course_num='310',
+                    section_num='501', term_code='201931', min_credits='3',
+                    honors_only=False, web_only=False, max_enrollment=50,
+                    current_enrollment=40, instructor=test_instructors[0]),
+            Section(crn=12346, id='000002', subject='CSCE', course_num='310',
+                    section_num='502', term_code='201931', min_credits='3',
+                    honors_only=False, web_only=False, max_enrollment=50,
+                    current_enrollment=40, instructor=test_instructors[1],
+                    instructor_gpa=3.2),
```
gannonprudhomme: Shouldn't these sections correspond to a Course that's in the DB? I'd change the `course_num` from 310 to 315. Not sure if this will change anything in the tests, but logically that's what it should be.

autoscheduler/scraper/serializers.py
```diff
@@ -0,0 +1,33 @@
...
+from .models.section import Section, Meeting
+
+class CourseSerializer(serializers.ModelSerializer):
+    """ Serializes a course """
+    class Meta:
+        model = Course
+        fields = ['title', 'credit_hours']
+
+class SectionSerializer(serializers.ModelSerializer):
+    """ Serializes a section """
```
gannonprudhomme: Maybe more thoroughly explain what a Serializer does in the docstrings of them? Kind of the same thing with the `SerializerMethodField`, as if everyone was completely familiar with Django we wouldn't need to explain this stuff, but I think it is beneficial to explain them just for clarity.

autoscheduler/scraper/tests/api_tests.py
gannonprudhomme: On that same note, we may need some form of storage of the most recent term that's updated somehow so that the server automatically knows what the most recent / upcoming term is. More of a thought rather than a review comment though lol.

autoscheduler/scraper/views.py
gannonprudhomme: Also probably explain that this view, given a dept, course_num, and term, returns the respective Course's title & credit hours.

autoscheduler/scraper/tests/api_tests.py
ryan-conn: > 
> 
> This should be something like
> 
> ```python
> expected = {
>     '202011': 'Spring 2020 - College Station',
>     '202012': 'Spring 2020 - Galveston',
>     '202013': 'Spring 2020 - Qatar'
>     '202014': 'Spring 2020 - Half Year Term'
>     '201931': 'Fall 2019 - College Station'
>     ....
> }
> ```

I think there should be separate tests making sure that term codes are formatted correctly, since it better follows the principle of only testing one component at a time, and we're just wanting to make sure that all the term codes in the database are displayed. If you disagree though, respond to this comment and I can change the test

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -0,0 +1,253 @@
...
+                    },
+                    '21': {
+                        'days': meeting_days_false,
+                        'start_time': second_start,
+                        'end_time': second_end,
+                        'type': 'LAB',
+                    },
+                },
+                'section_num': '502',
+                'web_only': False,
```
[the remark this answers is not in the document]
ryan-conn: I thought it looked nicer and using ids as properties for sections/meetings is still pretty intuitive, but if you prefer, we can do the following instead (this new version is how django rest framework formats list views by default). I'd like if @firejake308 could say which he prefers as well, I get where you're coming from and it would be a very easy change to make:
```json
[
    {
        "id": "10923201931",
        "crn": 10923,
        "instructor_gpa": null,
        "instructor_name": "John M. Moore",
        "honors_only": null,
        "meetings": [
            {
                "id": "4683040",
                "days": [
                    true,
                    false,
                    true,
                    false,
                    true,
                    false,
                    false
                ],
                "start_time": "11:30:00",
                "end_time": "12:20:00",
                "type": "LEC"
            },
            {
                "id": "4683041",
                "days": [
                    true,
                    false,
                    true,
                    false,
                    false,
                    false,
                    false
                ],
                "start_time": "15:00:00",
                "end_time": "15:50:00",
                "type": "LAB"
            },
        ],
        "section_num": "503",
        "web_only": null
    },
    {
        "id": "39493201931",
        "crn": 39493,
        "instructor_gpa": null,
        "instructor_name": "John M. Moore",
        "honors_only": null,
        "meetings": [
            {
                "id": "4911320",
                "days": [
                    true,
                    false,
                    true,
                    false,
                    true,
                    false,
                    false
                ],
                "start_time": "11:30:00",
                "end_time": "12:20:00",
                "type": "LEC"
            },
            {
                "id": "4911321",
                "days": [
                    true,
                    false,
                    true,
                    false,
                    false,
                    false,
                    false
                ],
                "start_time": "19:00:00",
                "end_time": "19:50:00",
                "type": "LAB"
            }
        ],
        "section_num": "504",
        "web_only": null
    },
]
```

autoscheduler/scraper/models/section.py
```diff
@@ -14,9 +14,14 @@ class Section(models.Model):
...
     min_credits = models.IntegerField() # Will never be null
     max_credits = models.IntegerField(null=True) # Will be null in most cases
+
+    honors_only = models.BooleanField(null=True)
+    web_only = models.BooleanField(null=True)
+
     max_enrollment = models.IntegerField()
     current_enrollment = models.IntegerField()
     instructor = models.ForeignKey('Instructor', on_delete=models.CASCADE, null=True)
+    instructor_gpa = models.FloatField(null=True)
```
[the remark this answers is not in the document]
ryan-conn: Since the field is already in all of the test cases, I figure we might as well just leave it in

ryan-conn:
All changes should be addressed now

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -0,0 +1,267 @@
...
+        ]
+        for course in self.courses:
+            course.save()
+        for section in self.sections:
+            section.save()
+        for meeting in self.meetings:
+            meeting.save()
+
+    def test_api_terms_displays_all_terms(self):
+        """ Tests that /api/terms returns ordered list of all terms in database """
```
gannonprudhomme: So I was just thinking that `expected` should be in order, but since it's a dictionary realized that there isn't any order.

As such, I don't think it matters the order they're returned in, as the frontend can just sort them by the keys in descending order. Thus, the docstring should change to 

```suggestion
        """ Tests that /api/terms returns a list of all terms in database """
```

autoscheduler/scraper/serializers.py
```diff
@@ -0,0 +1,36 @@
...
+    meetings = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Section
+        fields = ['id', 'crn', 'instructor_name', 'honors', 'meetings',
+                  'section_num', 'web']
+
+    # Used to get value for instructor_name field
+    def get_instructor_name(self, obj):
+        """ Get the name (id) of this section's instructor """
```
gannonprudhomme: > (just putting this to have it on record)

Change this comment to be in the docstring

autoscheduler/scraper/views.py
```diff
@@ -1,4 +1,16 @@
...
 
-# Create your views here.
+class RetrieveCourseView(generics.RetrieveAPIView):
+    """ API endpoint for viewing course information, used by /api/course.
+        This view returns a serialized course, should return its title and credit hours.
+    """
+    serializer_class = CourseSerializer
+
+    def get_object(self):
+        """ Overrides default behavior of get_object() to work without a primary key """
```
gannonprudhomme: What does you mean by without a primary key? Does `RetrieveApiView` just use the serializer's model and query using the parameters by default?

autoscheduler/scraper/tests/api_tests.py
ryan-conn: Since python 3.6, dictionaries actually do keep insertion order:
https://stackoverflow.com/questions/1867861/how-to-keep-keys-values-in-same-order-as-declared
So as long as we make a dict using a sorted list of terms, the keys will stay sorted, which we'll want to do.

autoscheduler/scraper/views.py
ryan-conn: By default, django rest framework requires you to provide the primary key of the item you're querying (see https://www.django-rest-framework.org/api-guide/generic-views/#get_objectself), and throws an error if it isn't provided, so I overwrote the `get_object(self)` method to instead get an object based on the query parameters.

PR #141, opened 2020-02-11
Added Grades model for scrape_grades

gannonprudhomme:
Adds the Grades model and GradeManager

After this is completed I'll do the actual `scrape_grades` command

### DIFF
--- a/autoscheduler/scraper/models/grades.py
+++ b/autoscheduler/scraper/models/grades.py
@@ -1,6 +1,6 @@
 """ Heavily based off of Good Bull Schedules """
 
-from typing import Dict
+from typing import Dict, Union
 from django.db import models
 
 class GradeManager(models.Manager):
@@ -10,12 +10,12 @@ class GradeManager(models.Manager):
 
     def instructor_performance(
             self, dept: str, course_num: str, instructor: str
-    ) -> Dict[str, float]:
+    ) -> Dict[str, Union[int, float]]:
         """ Aggregates all of the GPA's / sum of each grade type into one object
             so we can quickly get how an instructor performed in past sections
             for a given course
 
-            Returns a dictionary of string-integer, pairs, where str is the attribute
+            Returns a dictionary of string-integer/float pairs, where str is the attribute
             (i.e. gpa), and float is the according value (is an integer for grade counts)
         """
 
@@ -51,13 +51,13 @@ class Grades(models.Model):
     # Section is the primary key, since one section = one grade distribution
     section = models.OneToOneField("Section", on_delete=models.CASCADE, primary_key=True)
 
-    # Overrides the default Grades.object member, which allows us
+    # Overrides the default Grades.objects member, which allows us
     # to call Grades.object.instructor_performance
     objects = GradeManager()
 
     gpa = models.FloatField()
 
-    # Each of these represent many students received the given grade
+    # Each of these represent how many students received the given grade
     # See http://student-rules.tamu.edu/rule10/ for more details
     A = models.IntegerField(db_index=True)
     B = models.IntegerField(db_index=True)
--- a/autoscheduler/scraper/models/grades.py
+++ b/autoscheduler/scraper/models/grades.py
@@ -1,0 +1,75 @@
+""" Heavily based off of Good Bull Schedules """
+
+from typing import Dict
+from django.db import models
+
+class GradeManager(models.Manager):
+    """ Connects to the Grades models so we can call
+        Grades.object.instructor_performance
+    """
+
+    def instructor_performance(
+            self, dept: str, course_num: str, instructor: str
+    ) -> Dict[str, float]:
+        """ Aggregates all of the GPA's / sum of each grade type into one object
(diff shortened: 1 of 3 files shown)


autoscheduler/scraper/tests/models_tests.py
```diff
@@ -1,10 +1,13 @@
...
+import django.test
 
 from scraper.models.course import generate_course_id
 from scraper.models.department import generate_department_id
-from scraper.models.section import generate_meeting_id
+from scraper.models.section import generate_meeting_id, Section
+from scraper.models.grades import Grades
+from scraper.models.instructor import Instructor
 
 # A lot of these funcitons have to be self in order to be part of the TestCase and thus,
```
ryan-conn:
```suggestion
# A lot of these functions have to be self in order to be part of the TestCase and thus,

```
I know you didn't actually change this but we should fix this typo anyways

autoscheduler/scraper/models/grades.py
```diff
@@ -0,0 +1,75 @@
...
+from django.db import models
+
+class GradeManager(models.Manager):
+    """ Connects to the Grades models so we can call
+        Grades.object.instructor_performance
+    """
+
+    def instructor_performance(
+            self, dept: str, course_num: str, instructor: str
+    ) -> Dict[str, float]:
```
ryan-conn: Since the values are floats or ints, we can use `Union` from `typing` to show that it can be either:
https://stackoverflow.com/questions/33945261/how-to-specify-multiple-return-types-using-type-hints

autoscheduler/scraper/tests/models_tests.py
```diff
@@ -36,8 +39,103 @@ class MeetingTests(unittest.TestCase):
...
+        Section.objects.bulk_create(sections)
+
+        grades = [
+            Grades(section=sections[0], gpa=4.0, A=1, B=0, C=0, D=0, F=0, I=0, S=0, U=0,
+                   Q=0, X=0),
+            Grades(section=sections[1], gpa=3.0, B=1, A=0, C=0, D=0, F=0, I=0, S=0, U=0,
+                   Q=0, X=0),
+        ]
+
+        # Save all at once
```
ryan-conn: Really minor but this comment should be above the first `bulk_create` instead of this one

autoscheduler/scraper/models/grades.py
```diff
@@ -0,0 +1,75 @@
...
+class Grades(models.Model):
+    """ Represents a collection of the grade distribution values for a
+        specific section
+    """
+
+    # Section is the primary key, since one section = one grade distribution
+    section = models.OneToOneField("Section", on_delete=models.CASCADE, primary_key=True)
+
+    # Overrides the default Grades.object member, which allows us
+    # to call Grades.object.instructor_performance
```
ryan-conn: Should be `Grades.objects`

autoscheduler/scraper/models/grades.py
```diff
@@ -0,0 +1,75 @@
...
+                A=models.Sum("A"),
+                B=models.Sum("B"),
+                C=models.Sum("C"),
+                D=models.Sum("D"),
+                F=models.Sum("F"),
+                I=models.Sum("I"),
+                S=models.Sum("S"),
+                U=models.Sum("U"),
+                Q=models.Sum("Q"),
+                X=models.Sum("X"),
```
ryan-conn: Where are we going to calculate percentages? I assume in the api route (we could cache it after calculating once if we do so), but just making sure we're on the same page

gannonprudhomme: Honestly I was thinking that the frontend would deal with the percentages and we just give them the sums, but we can calculate percentages if you think that's better. My thought process was that giving them the sums would provide the frontend with a bit more flexibility with what they want to do with it.

ryan-conn: True, calculating it client-side will be faster for us and provide more flexibility with how they display grade distributions (we could show total students, number of students who got each letter grade, etc), so that seems like the way to go.

autoscheduler/scraper/models/grades.py
gannonprudhomme: So would this look like `Dict[str, Union(int, float)]`?

ryan-conn: Correct

autoscheduler/scraper/tests/models_tests.py
gannonprudhomme: So we can actually remove these comment entirely, as well as the `#pylint: disable=no-self-use`, since we're using `self.assertCondition` instead of `assert condition` now

autoscheduler/scraper/tests/models_tests.py
gannonprudhomme: Honestly I'm just gonna remove this comment. I think `bulk_create` is pretty straightforward, plus it's a really easy search to see what it does in the case it's not.

gannonprudhomme:
Force pushed as I reworded the commit message just btw

PR #150, opened 2020-02-19
Input Availability

firejake308:
Users can now input their availability using the `<Schedule />` component. 

Closing #122


gannonprudhomme:
While functionally this looks great (although rn it's making *every* time box you drag default to 30 min), this is a pretty huge PR, and if we review this in one piece I feel like it'll become a #94 with a bunch of comments and become hard to jumble all the changes. As such, could you try to split this up into ~2-3 logical pieces if possible? I know given the nature of React + Redux so many of the changes are tied together, but I don't think we can effectively review this otherwise.

firejake308:
I fixed the bug in the latest commit, and to make the reviewing process more reasonable, please review PR #151 before this one

ryan-conn:
Here are my comments from the meeting:
- Can still create availabilities before 8 AM or past 9 PM
- Creating 2 blocks and then one that overlaps both will merge twice
- Time selection thing should disappear when clicking since you can see the time anyways, or when time is before 8 AM
- Dragging to create a new availability with same start or end time as an existing one will immediately delete the existing one

firejake308:
Okay, I've investigated mocking mouse events, and apparently JSDOM supports them, but not well enough to meet our needs. Consequently, there won't be much automatic UI testing for this feature

firejake308:
Rather than continuously rebasing and force-pushing, I have made a fresh branch and an accompanying PR. See #160.

PR #152, opened 2020-02-20
Section API Fixes

ryan-conn:
Fixed the expected formatting in section API tests, and changed `SectionSerializer` to format time objects correctly. I'd like @btgathright to review these changes and make sure everything is okay

### DIFF
--- a/autoscheduler/scraper/serializers.py
+++ b/autoscheduler/scraper/serializers.py
@@ -1,6 +1,11 @@
+from datetime import time
 from rest_framework import serializers
+from time import strftime
 from .models.course import Course
 from .models.section import Section, Meeting
+
+def format_time(t: time) -> str:
+    return '' if t is None else t.strftime('%H:%M')
 
 class CourseSerializer(serializers.ModelSerializer):
     """ Serializes a course into an object with information needed by /api/course """
@@ -32,7 +37,7 @@ class SectionSerializer(serializers.ModelSerializer):
         return [{
             'id': str(meeting.id),
             'days': meeting.meeting_days,
-            'start_time': meeting.start_time,
-            'end_time': meeting.end_time,
+            'start_time': format_time(meeting.start_time),
+            'end_time': format_time(meeting.end_time),
             'type': meeting.meeting_type,
         } for meeting in meetings]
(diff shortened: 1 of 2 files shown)


ryan-conn:
Apparently my local linter is broken now, will fix linter errors soon

autoscheduler/scraper/serializers.py
```diff
@@ -1,11 +1,11 @@
...
 from rest_framework import serializers
-from time import strftime
 from .models.course import Course
 from .models.section import Section, Meeting
 
-def format_time(t: time) -> str:
-    return '' if t is None else t.strftime('%H:%M')
+def format_time(time_obj: time) -> str:
+    """ Formats a time object to a string HH:MM, for use with section serializer """
+    return '' if time_obj is None else time_obj.strftime('%H:%M')
```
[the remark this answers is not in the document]
ryan-conn: I'm going to leave it like this since it's a relatively simple ternary, and leaving it as one line makes me feel like a 1337 python god

autoscheduler/scraper/serializers.py
```diff
@@ -32,7 +37,7 @@ def get_meetings(self, obj): # pylint: disable=no-self-use
         return [{
             'id': str(meeting.id),
             'days': meeting.meeting_days,
-            'start_time': meeting.start_time,
-            'end_time': meeting.end_time,
+            'start_time': format_time(meeting.start_time),
+            'end_time': format_time(meeting.end_time),
```
firejake308: So this change is just pretty-printing the time?

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -196,7 +249,7 @@ def test_api_sections_gives_valid_response(self):
                         'type': 'LEC',
                     },
                     {
-                        'id': '20',
+                        'id': '11',
```
firejake308: What's the reason for this ID change?

ryan-conn: The previous ID was incorrect, so the test was failing even when it should pass

autoscheduler/scraper/serializers.py
ryan-conn: Basically, since before this returned a time object, which would then be transformed by the view into the format `HH:MM:SS`. Instead of relying on this behavior (and including seconds in the time despite the fact that they'll never be needed), I decided to convert time to a string in the serializer

PR #155, opened 2020-02-27
Additional testing for /api/course/search

ryan-conn:
Since the current tests for `/api/course/search` aren't all that thorough, I added tests to make sure that lowercase letters in searches, as well as numbers and spaces (encoded as `%20`), are correctly handled.

### DIFF
--- a/autoscheduler/scraper/tests/api_tests.py
+++ b/autoscheduler/scraper/tests/api_tests.py
@@ -8,9 +8,10 @@ from scraper.serializers import CourseSerializer, SectionSerializer
 
 class APITests(APITestCase):
     """ Tests API functionality """
-    def setUp(self):
-        self.client = APIClient()
-        self.courses = [
+    @classmethod
+    def setUpTestData(cls):
+        cls.client = APIClient()
+        cls.courses = [
             Course(id='CSCE181-201931', dept='CSCE', course_num='181',
                    title='Introduction to Computing', term='201931', credit_hours=3),
             Course(id='CSCE315-201931', dept='CSCE', course_num='315',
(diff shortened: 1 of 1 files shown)


autoscheduler/scraper/tests/api_tests.py
```diff
@@ -265,3 +269,33 @@ def test_api_course_search_gives_correct_results_c(self):
...
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.json(), expected)
+
+    def test_api_course_search_gives_correct_results_csce_3(self):
+        """ Tests that /api/course/search?search=CSCE%203&term=201731 gives correct
+            output
+        """
+        # Arrange
+        expected = {'results': ['CSCE 310', 'CSCE 315']}
+        data = {'search': 'CSCE%203', 'term': '201731'}
```
firejake308: Does this actually test that the 3 is processed? I think it would be better to add another CSCE course in the 100s or 200s that would not be matched by this test

ryan-conn: Done, I also added a test to make sure a lowercase search with numbers (such as "csce 3") is handled correctly

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -265,3 +271,48 @@ def test_api_course_search_gives_correct_results_c(self):
...
+        data = {'search': 'csce', 'term': '201731'}
+
+        # Act
+        response = self.client.get('/api/course/search', data=data)
+
+        # Assert
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), expected)
+
+    def test_api_course_search_gives_correct_results_csce_3_lower(self):
```
firejake308: This last test might not be strictly necessary, since each feature (lowercase and number) works individually and there's no reason for them to interfere with other. However, I wouldn't change it now that it's been done. This PR looks good to me, I'll just let one of the backend guys look at it once for final approval.

ryan-conn:
Changes addressed, and the force push shouldn't cause a problem unless you pulled the last commit I made a couple minutes before it

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -265,3 +267,48 @@ def test_api_course_search_gives_correct_results_c(self):
...
+        """ Tests that /api/course/search?search=csce%203&term=201731 gives correct
+            output (lowercase search with a number works)
```
gannonprudhomme:
```suggestion
        """ Tests that /api/course works with lowercase subject and a number works
```

Doesn't have to be this exact phrasing, but there's no reason to say "correct output" when you could instead just say the situation it's actually testing.

Likewise, I honestly don't think the `/api/course/search?search=csce...` stuff is that helpful, and I think it would be more useful if you just dropped the URL parameter stuff and just straight up either said what the situation was like this, or just mention what the parameters are.

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -265,3 +267,48 @@ def test_api_course_search_gives_correct_results_c(self):
...
+        # Act
+        response = self.client.get('/api/course/search', data=data)
+
+        # Assert
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(response.json(), expected)
+
+    def test_api_course_search_gives_correct_results_csce_lower(self):
+        """ Tests that /api/course/search?search=csce&term=201731 gives correct
+            output
```
gannonprudhomme: Same with this, update the specific cases with what you're actually testing

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -265,3 +267,48 @@ def test_api_course_search_gives_correct_results_c(self):
         # Assert
         self.assertEqual(response.status_code, 200)
         self.assertEqual(response.json(), expected)
+
+    def test_api_course_search_gives_correct_results_csce_3(self):
+        """ Tests that /api/course/search?search=CSCE%203&term=201731 gives correct
+            output
```
gannonprudhomme: Also this

ryan-conn:
Docstrings updated, let me know what you think

> Other than the docstring changes, like we discussed I'd say split up all of these tests into separate classes in `api_tests.py`, such as `CourseSearchAPITests`, etc.

I agree, but since it doesn't specifically have to do with this PR, I created issue #161 to address this.

PR #154, opened 2020-02-25
Scraper API Term & Course Search

ThatJuanGuy:
Implements api/terms and api/course/search. For api/course/search, white space between the department and course number is trimmed. Department must also be entered before course number in the search parameter.

PR #159, opened 2020-03-03
Added scrape_grades command

gannonprudhomme:
So this is mostly implemented other than the tests. I made this a PR (although probs should be a draft) just so we could discuss the testing for it, since a lot of the functions in it are kind of funky to test.

I'm also considering adding optional `--year` and `--college` arguments to it so we can quickly scrape a small amount, rather than scraping everything from 2013 - 2019.

I also need to figure out how to add SSL certificate creation to `requests.get` call in `fetch_page_data`, as not doing so gives a warning on the function call.


firejake308:
I got this error: `No such file or directory: '/home/firejake308/AAS/autoscheduler/documents/grade_dists/grd20193GV.pdf'` Do I need to download a PDF for this to work?

gannonprudhomme:
Damn I was afraid of that. Do you have the `autoscheduler/documents/grade_dists` folder created? It should have created it automatically, but that's generally why `FileNotFound` errors occur for me. The `download_pdf` function downloads the PDF and then saves them to the directory w/ `save_pdf`, so it's nothing on your end

firejake308:
So I resolved the FileNotFound by creating the directory manually, but now I get this error:

> django.db.utils.ProgrammingError:` column sections.honors does not exist
LINE 1: ...ections"."min_credits", "sections"."max_credits", "sections"...

I tried `makemirgrations` and `migrate`. Do I need to migrate the schema from a different branch to get the sections.honors field?

gannonprudhomme:
> So I resolved the FileNotFound by creating the directory manually, but now I get this error:
> 
> > django.db.utils.ProgrammingError:` column sections.honors does not exist
> > LINE 1: ...ections"."min_credits", "sections"."max_credits", "sections"...
> 
> I tried `makemirgrations` and `migrate`. Do I need to migrate the schema from a different branch to get the sections.honors field?

No the honors field is in [there](https://github.com/aggie-coding-club/Rev-Registration/blob/730932f044f2f756b5f293d3a63ae40c62ddf191/autoscheduler/scraper/models/section.py), I'm honestly not sure why that's happening. In situations like these I generally do the following:

1. `pip install django-extensions` 
2. Go to `autoscheduler/autoscheduler/settings/base.py` and add "django_extensions" under `INSTALLED_APPS`. (tutorial for setting it up [here](https://pypi.org/project/django-extensions/))
3. Run `python manage.py reset_db`
4. Delete all of the migrations in `scraper/migrations`
5. Run `makemigrations` and `migrate`

firejake308:
Ok it owrks now. I'm assuming that eventually `term` will be an argument to this command, just like it is for `scrape_depts` and `scrape_courses`?

autoscheduler/scraper/tests/commands/scrape_grades_tests.py
```diff
@@ -0,0 +1,83 @@
...
+                   section=self.section),
+        ]
+
+        # Act
+        result = scrape_pdf(grade_dists, term)
+
+        # Assert
+        self.assertEqual(expected, result)
+
+    # Test that it throws an error on a section-not-found?
```
firejake308: I would add that test, but other than that, the tests look good to me for now. `scrape_pdf` is a pretty simple functionality, so there's not much to test, and if you can figure out how to mock file IO and network requests, the rest of it should be good too.

gannonprudhomme:
> Ok it works now. I'm assuming that eventually `term` will be an argument to this command, just like it is for `scrape_depts` and `scrape_courses`?

So grade distributions don't really do terms like Banner does them. Instead they're ordered by year+semester and the school/college its under. For instance, there's a `20191` (spring 2019) for `EN`, which is the College of Engineering in College Station, as well as a `20191` for `GV` which is for all courses in TAMU Galveston.

So like I mentioned in the PR description, we could add a year argument that takes the `20191` part of the term, as well as a college argument that takes the `EN`. That being said, if we can always get all of the grade distributions I'd say it's good leaving it as the default, cause ideally for deploying we only want to run most of the scraping one time to fill the DB (other than running `scrape_courses` periodically for updating the current seat count). Plus since grade distributions uses past semesters, it's important to have all of the data we can get.

For testing though I'd say it'd be useful to have arguments just so we can fill the DB quickly for testing various features that use the grade distributions

firejake308:
Oh, right, that makes more sense. If it's possible to get all of the data at once, then I think that's the way to go

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,259 @@
...
+    tasks = []
+    for year in years:
+        for semester in semesters:
+            for college in colleges:
+                year_semester_term = str(year) + semester
+                tasks.append(retrieve_pdf(year_semester_term, college))
+
+    grades = [] # list of Grades models
+    for scraped_grades in await asyncio.gather(*tasks, loop=loop):
+        if scraped_grades is not None and len(scraped_grades) > 0:
```
ryan-conn:
```suggestion
        if scraped_grades:
```
Handles both cases at once

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,259 @@
...
+    """ Given a term in the format {YEAR}{SEMESTER}, converts it to include the location
+
+        Ex: 20193 (Fall 2019) -> 201931 (Fall 2019 College Station)
+    """
+
+    location = "1" # Default, College Station
+    if college == "GV": # TAMU Galveston
+        location = "2"
+    elif college == "QT": # TAMU Qatar
+        location = "3"
```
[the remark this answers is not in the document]
gannonprudhomme: Well the thing about the locations is that for the other ones its just assumed that it's in College Station and doesn't actually specify it. Ex. EN is always the College of Engineering classes in College Station, so this isn't really possible.

gannonprudhomme:
So I'm not sure how I just realized this, but the old pdf style (believe it changed around 2016 Fall) isn't actually parsed correctly, so none of the PDF's before 2016 Fall will actually have grades for them. You can see this if you try to run it for 2015 (change `years` in `handle()` from `get_available_years` to `[2015]`), you'll see that it says "No grades scraped". The only time it should do this is if it returns a bunch of `Section not found`'s, which would indicate that you just haven't run `scrape_courses` for that term. But since it doesn't, this shows that it actually doesn't scrape *any* grades for the PDF's of that style. Working on a fix right now, and will add tests for it in `pdf_parser_tests`

gannonprudhomme:
Just added the fix for the above comment. You can read the commit description for more information about it.

gannonprudhomme:
I also added optional CLI arguments for scrape_grades, so you can run `python manage.py scrape_grades --year 2015 --college EN`, or for short, `python manage.py scrape_grades -y 2015 -c EN`. (The case of the college doesn't actually matter, so `en` would also work)

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,271 @@
...
+
+    return grades
+
+def fetch_page_data() -> bs4.BeautifulSoup:
+    """ Retrieves the ROOT_URL HTML data and converts it to BeautifulSoup
+        so we can parse it to get the available years & colleges to scrape grades for
+    """
+
+    # TODO Add SSL certification
+    response = requests.get(ROOT_URL, verify=False)
```
ryan-conn:
```suggestion
    response = requests.get(ROOT_URL, verify=True)
```
Might want to make an issue to make this more robust, but this is better than having no authentication at all

gannonprudhomme: Imma be honest I have no idea why this works now as it didn't for me before but I'll take it lol

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,270 @@
...
+    for year in years:
+        for semester in semesters:
+            for college in colleges:
+                year_semester_term = str(year) + semester
+                tasks.append(retrieve_pdf(year_semester_term, college))
+
+    grades = [] # list of Grades models
+    for scraped_grades in await asyncio.gather(*tasks, loop=loop):
+        if scraped_grades:
+            grades.extend(scraped_grades)
```
ryan-conn:
```suggestion
    # Flatten grades scraped from each college into a list
    grades = [grade for scraped_grades in await asyncio.gather(*tasks, loop=loop)
              for grade in scraped_grades]
```
Same as my other list comprehension comment, up to you if this is readable enough to use

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,270 @@
...
+
+    semesters = [SPRING, SUMMER, FALL]
+
+    # Gather all of the retrieve functions into an array
+    tasks = []
+    for year in years:
+        for semester in semesters:
+            for college in colleges:
+                year_semester_term = str(year) + semester
+                tasks.append(retrieve_pdf(year_semester_term, college))
```
ryan-conn:
```suggestion
    tasks = [retrieve_pdf(f"{year}{semester}", college) for year in years
             for semester in semesters for college in colleges]
```
This does the same thing if you want to change it but I understand if you don't want to use a list comprehension like this

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,270 @@
...
+            return None
+
+    semesters = [SPRING, SUMMER, FALL]
+
+    # Gather all of the retrieve functions into an array
+    tasks = []
+    for year in years:
+        for semester in semesters:
+            for college in colleges:
+                year_semester_term = str(year) + semester
```
ryan-conn: If you decide not to use my list comprehension, change this to an f string

autoscheduler/scraper/management/commands/scrape_grades.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # Retrieve the page data and get the available colleges & years from it
+        page_soup = fetch_page_data()
+        years = _get_available_years(page_soup)
+        if options['year']:
+            years = [options['year']]
+
+        colleges = _get_colleges(page_soup)
+        if options['college']:
+            colleges = [options['college']]
```
ryan-conn: Change to a ternary or if/else so we don't have to call `select` unless the values are being used

autoscheduler/scraper/management/commands/scrape_grades.py
gannonprudhomme: Going to not use list comprehension here, since I don't think it's very clear that you're unfolding the list (so that grades is just a normal list rather than a list of lists) in this. Same with the `if scraped_grades:` check

gannonprudhomme:
I'm just gonna wait until #154 is merged so I can deal with rebasing here rather than in that PR

gannonprudhomme:
Since this is completed, I'm going to change the base from `backend/scraper/scrape_grades` to `backend/master`, instead of merging this into `scrape_grades` then from there making a PR to merge into `backend/master`.

gannonprudhomme:
whelp glad I checked. I rebased wrong, so I'm going to undo all of this and do it again.

I accidentally rebased onto `master` originally, hence why the dependabot and README update commits are in here.

gannonprudhomme:
Ok so I think I fixed it, but to double check I'm gonna reset the base back to `backend/scraper/scrape_grades`

gannonprudhomme:
Ok should be good to go now. Changing the base back to `backend/master` then rebase & merging.

PR #160, opened 2020-03-06
Availabilty UI

### LINKED ISSUE #122: Implement time availablity component
Fill in with more details once design is done (see #121)

firejake308:
I have separated out the changes to the availability UI the same way I did for the Redux. I have also successfully added tests for most of the bugs that @ryan-conn pointed out and resolved them as well. This should be ready for review.

Closes #122


autoscheduler/frontend/src/components/Schedule/Schedule.css
```diff
@@ -74,11 +76,12 @@
     flex-grow: 1;
     display: flex;
     flex-direction: column;
-    z-index: -1;
-    background-color: white;
+    background: linear-gradient(transparent 0, transparent 0.5rem, white 0.5rem);
```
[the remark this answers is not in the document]
firejake308: You aren’t missing anything, as there is not actually any real gradient. Instead, I’m using linear gradient to make the top 0.5rem of the calendar day transparent so that it doesn’t  cover up the header.

firejake308:
All changes addressed. Large areas of code had to be changed to fix some of them, so please make sure that my changes sort of make sense. I have attempted to make the UI behave in a way that is intuitive, but if anything feels off, lmk.

autoscheduler/frontend/src/tests/AvailabilityUI.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+      const monday = getByLabelText('Monday');
+      fireEvent.mouseDown(monday, startEventProps);
+      fireEvent.mouseMove(monday, startEventProps);
+      fireEvent.mouseDown(queryByLabelText('Adjust End Time'), startEventProps);
+      fireEvent.mouseMove(monday, endEventProps);
+      fireEvent.mouseUp(monday, endEventProps);
+
+      // assert
+      expect(getByText('BUSY')).toBeInTheDocument();
+      expect(queryByText(/NaN/)).not.toBeInTheDocument();
```
gannonprudhomme: Really just for my understanding, but what does querying for `/NaN/` test for?

autoscheduler/frontend/src/components/ScheduleCard/ScheduleCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+  backgroundColor?: string;
+  backgroundStripes?: boolean;
+  firstHour: number;
+  lastHour: number;
+  onResizeWindow?: (isBig: boolean) => void;
+  onDragHandleDown?: (evt: React.MouseEvent<HTMLDivElement, MouseEvent>,
+    endSelected: boolean) => void;
+}
+
+type ScheduleCardProps = React.PropsWithChildren<BasicProps>;
```
gannonprudhomme: Just for my understanding, but how come you have to use `PropsWithChildren`, rather than just using the props interface directly like you do in other places?

autoscheduler/frontend/src/tests/AvailabilityUI.test.tsx
```diff
@@ -0,0 +1,267 @@
...
+      const monday = getByLabelText('Monday');
+      fireEvent.mouseDown(monday, mouseEventProps);
+      fireEvent.mouseUp(monday, mouseEventProps);
+
+      // assert
+      expect(getByText('BUSY')).toBeInTheDocument();
+      // ensures that all times are actually calculated based on mocked measurements
+      expect(queryByText(/NaN/)).not.toBeInTheDocument();
+      expect(queryByLabelText('Adjust Start Time'))
+        .toHaveAttribute('aria-valuetext', expectedStart);
```
gannonprudhomme: Do you think it's worth asserting that the size of the availability component on click defaults to 30 minutes long, or is that more of a Redux/functional test, rather than a UI test?

autoscheduler/frontend/src/components/ScheduleCard/ScheduleCard.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const cardRoot = React.useRef<HTMLDivElement>(null);
+  const cardContent = React.useRef<HTMLDivElement>(null);
+  React.useEffect(() => {
+    const handleResize = (): void => {
+      // set initial height for future use
+      contentHeight = contentHeight || cardContent.current.clientHeight;
+
+      if (contentHeight >= cardRoot.current.clientHeight) {
+        if (onResizeWindow) onResizeWindow(false);
+      } else if (onResizeWindow) onResizeWindow(true);
```
ryan-conn:
```suggestion
      if (onResizeWindow) {
        onResizeWindow(contentHeight < cardRoot.current.clientHeight);
      }
```
I think this is functionally identical, the way it's written right now (checking if `onResizeWindow` in a different place depending on the branch) is a little confusing. Also could use a comment explaining what the bool value represents (it's hard to tell without looking at `MeetingCard` since that's the only component that currently uses this value)

autoscheduler/frontend/src/tests/MeetingCard.test.tsx
```diff
@@ -33,39 +36,23 @@ const testMeeting = new Meeting({
...
+    <Provider store={store}>
+      <MeetingCard
+        meeting={testMeeting}
+        bgColor="#500000"
+        firstHour={8}
+        lastHour={21}
+      />
+    </Provider>,
   );
+  expect(container).toBeTruthy();
```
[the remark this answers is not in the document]
firejake308: I'm leaning towards leave it, just because with the new `ScheduleCard`, just to test the basic functionality of the `MeetingCard` and to keep up the code coverage of the tests.

autoscheduler/frontend/src/tests/AvailabilityUI.test.tsx
firejake308: Before I figured out how to properly mock mouse events, the start time and end time of each availability block would be `NaN`, so this makes sure that it's actually calculating the times using the mocked `MouseEvent`s thar I'm giving

autoscheduler/frontend/src/components/ScheduleCard/ScheduleCard.tsx
firejake308: `PropsWithChildren` includes `props`, the normal props object that we pass into most components, and `children`, a special argument that contains all of the elements inside of this one. So, for example, if we have:

```
<ScheduleCard>
  <p>Hello</p>
</ScheduleCard>
```
then `children` is the inner `<p>` tag

autoscheduler/frontend/src/tests/AvailabilityUI.test.tsx
firejake308: The 30-minute minimum is enforced by the UI, not Redux. Therefore, I will asjust the test to check that.

autoscheduler/frontend/src/components/ScheduleCard/ScheduleCard.tsx
firejake308: Naw you right

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -60,9 +275,45 @@ const Schedule: React.FC<RouteComponentProps> = () => {
       />
     );
   }
+  function renderAvailability(availability: Availability): JSX.Element {
+    return (
+      <AvailabilityCard
+        availability={availability}
+        firstHour={FIRST_HOUR}
+        lastHour={LAST_HOUR}
+        key={`${availability.startTimeHours}:${availability.startTimeMinutes}`}
```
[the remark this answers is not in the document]
firejake308: I made one in `timeUtil.ts`

firejake308:
Aight check it out now

gannonprudhomme:
This is looking pretty close to finished. However, do you think we could have some global state so that we can quickly switch between everything using 12-hr time vs 24-hr time? As it is right now some things use 12-hr time and others us 24-hr time, so I think we should be consistent on what we do. Plus this way we can easily add a setting so that the user can switch between using them.

firejake308:
In general, I use 24-hour for implementation, as the supporting code needs to be able to differentiate between 8 AM and 8 PM, but I believe I always use 12-hour time when displaying times to the user, since that is what we Americans understand more easily. Unless there is a deviation from these ground rules, I do not believe any changes are necessary.

autoscheduler/frontend/src/timeUtil.ts
```diff
@@ -0,0 +1,14 @@
...
+export const LAST_HOUR = 21;
+/**
+ * Given the hours and minutes of a time, returns a formatted string representation of
+ * that time. By default, this uses 24-hour format, but an extra argument can be specified
+ * to use 12-hour format
+ * @param h hours
+ * @param m minutes
+ * @param use12Hour if true, applies 12-hour format instead of the default 24-hour format
+ */
+export function formatTime(h: number, m: number, use12Hour = false): string {
```
gannonprudhomme: So there's actually quite a few places that are using 24-hr time rather than 12-hr time. Instead of changing them in place, I'd say just change the default here to be to use12Hour time, and remove any place it's listed as true (see [this](https://github.com/aggie-coding-club/Rev-Registration/pull/160/commits/a8afbc8f25cbc5e70e228d55ac951f55df3b7f07#) commit)

Instead of listing the code places, you can see it showing 24-hr time here:

chrome_3o3lGv7l64

It also uses 24-hr time for the MeetingCards.

```suggestion
export function formatTime(h: number, m: number, use12Hour = true): string {
```

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -281,7 +280,7 @@ const Schedule: React.FC<RouteComponentProps> = () => {
         availability={availability}
         firstHour={FIRST_HOUR}
         lastHour={LAST_HOUR}
-        key={`${availability.startTimeHours}:${availability.startTimeMinutes}`}
+        key={formatTime(availability.startTimeHours, availability.startTimeMinutes, true)}
```
gannonprudhomme:
```suggestion
        key={formatTime(availability.startTimeHours, availability.startTimeMinutes)}
```

Per my previous comment, this is now redundant.

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -1,25 +1,252 @@
...
+      }
+      // new time blocks cannot be later than 9 PM / 2100
+      return [{
+        ...avl,
+        time2: 21 * 60,
+      }, {
+        ...avl,
+        time1: 21 * 60,
+        time2: 20 * 60 + 30,
+      }];
```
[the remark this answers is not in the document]
firejake308: The code that you tried is exactly what I had initially, but I had to change it to deal with a bug. The reason for this bug has to do with the implementation of `updateAvailability`. In particular, the way I determine which availability to update is by finding the availability currently in the Redux store whose start time or end time matches the provided `time1`. As such, if you try to replace the original availability with the 20:30 - 21:00 one in one go, the reducer will not find any existing availability that either starts or ends at 20:30, so the update will not be registered. 

The current implementation first replaces the end time of the existing availability with 21:00, so that in the second step, it can successfully find an availability whose start or end time is 21:00, and then update the start time of that availability to 20:30. Kinda confusing, I know, but it's the simplest way to get it to work without doing a ground-up rewrite of the availability reducer.

autoscheduler/frontend/src/timeUtil.ts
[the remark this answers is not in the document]
firejake308: I actually had one in there, but ESLint automatically removes it if you provide a default value, since the type can be inferred from that of the default value

firejake308: Oof, forgot about that

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -1,25 +1,252 @@
...
+    }
+
+    // if there are no problems, just use avl as is
+    return [avl];
+  }
```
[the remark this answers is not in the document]
firejake308: This looks more or less okay to me, but I don't want to accidentally break this, so I'm going to move this into `Availability.ts`, add unit tests, and then apply the suggested changes. Just letting you know so you know where to look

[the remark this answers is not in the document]
firejake308: Oh wait, on second thought, I don't think this is going to work due to the assumption that `time2` is always after `time1`. When updating availabilities, `time1` actually needs to be whichever time is not being dragged, regardless of whether that is the start time or end time. Using the suggested code would result in `time1` sometimes being changed to the unstable, currently-dragged time, if the user was dragging toward the top, and this would cause a failed update, as I explained in a different comment.

This behavior is documented in the docstring for `updateAvailability`, but is incorrectly documented in the docstring for `AvailabilityArgs`. I have updated the documentation to reflect the correct usage.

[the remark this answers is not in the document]
ryan-conn: Okay, if you're not making this change then I think we should change the functionality slightly for the function you have: right now, if a block less than 30 minutes is created before 8:30, it will "snap" to a block from 8 to 8:30 which is inconsistent with any other time (keep start time and expand downwards). If you just remove the check for if it's before 8:30 then it behaves as I would expect. Is there a reason you decided to make it a special case?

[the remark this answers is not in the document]
gannonprudhomme: Yeah this was a bug that I reported I believe. If you start before 8:30 AM and drag upwards it automatically defaults to 30 minutes like we've had for awhile, so it'll force to 30 minutes before the start time (i.e. if you start at 8:20 AM and drag up, it'll default to 7:50 AM). What were you expecting for starting before 8:30 AM? Would you rather them default to 10-20 min / default to the start time being 8:00 AM regardless of what time we started it at?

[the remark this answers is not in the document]
gannonprudhomme: Ok nvm scratch that ^. Ryan's talking about starting from before 8:30 and dragging down (so it's the end time), while I'm talking about starting before 8:30 and dragging up (so it's the end time), but the two issues are related.

Likewise right now if you start at 8:10 and drag down, then delete it, it makes another availability from 8:00 AM to 8:30 AM.

gannonprudhomme:
So I had a hard time reproducing it, but I found a bug with overlaps. When it occurs and you delete the top-most availability, you can't perform any action on the bottom availability, and some times attempting to drag it caused it to crash. Also for some reason it didn't appear that the bottom availability was actually in the session, as exporting the state/session with Redux Dev Tools and loading it again only had one of the availabilities shown rather than both of them, but this may be invalid due to my poor understanding of Redux lol.

It occurred when I would make two availbilities, then fiddle with them a bit, then dragging one of the edges to overlap with the other. Again you may not be able to reproduce it as I could only get it to occur like ~4 times, but just wanted to note it.

chrome_RnAJmkdbEc

chrome_PyXFe3cFXB

gannonprudhomme:
Also, if you drag inside of an availability but not on one of the handles, it throws an `Cannot read property 'startTimeHours' of undefined` error. I assume we don't actually need to do anything in this case, so we should just catch this error.

firejake308:
For that last bug, I'm working on preventing dragging from within the
availability. That should also address some of Ryan's concerns as well.

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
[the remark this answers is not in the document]
firejake308: So let me clarify the expected behavior:

1. If you start at 8:20 and drag up to 8:10, it will expand to 8:00 - 8:30.
2. If you start at 8:10 and drag down to 8:20, it will expand to 8:10 - 8:40.

Would y'all agree?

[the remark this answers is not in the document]
ryan-conn: Yeah, since that's what happens with other times on the calendar I think that would be the best solution. Looking at it again, it seems like the same thing happens with times close to 9 PM (expanding to 8:30-9 rather than basing it on which way you drag), so that should be made consistent as well.

ryan-conn:
Looking at your last commit, my intention was for you to just change the z buffer of the hovered time rather than hide it when you're hovering over a card. We can discuss what we'd rather do in the meeting though

firejake308:
>Looking at your last commit, my intention was for you to just change the z buffer of the hovered time rather than hide it when you're hovering over a card. We can discuss what we'd rather do in the meeting though

The problem with this is that if I do this, the start and end times of the ScheduleCards will never be shown, since my mouse will be hovering over the `<HoveredTime />` and not the `<ScheduleCard />`, like in the image below. If we prefer this behavior, we can go with it, but I thought it would be wiser just to hide the HoveredTime.

>Also, if you drag inside of an availability but not on one of the handles, it throws an Cannot read property 'startTimeHours' of undefined error. I assume we don't actually need to do anything in this case, so we should just catch this error.

Fixed by preventing drag from starting inside of an AvailabilityCard

>So I had a hard time reproducing it, but I found a bug with overlaps. When it occurs and you delete the top-most availability, you can't perform any action on the ...

I was unable to reproduce this bug. I'll keep an eye out for it, but I don't think I can debug it until I can see it.

firejake308:
Aight I got the two things we talked about at last night's meeting. Anything else?

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
```diff
@@ -260,7 +260,8 @@ const Schedule: React.FC<RouteComponentProps> = () => {
         availability={availability}
         firstHour={FIRST_HOUR}
         lastHour={LAST_HOUR}
-        key={formatTime(availability.startTimeHours, availability.startTimeMinutes, true)}
+        key={`${formatTime(availability.startTimeHours, availability.startTimeMinutes, true)}
+        - ${formatTime(availability.endTimeHours, availability.endTimeMinutes)}`}
```
ryan-conn: Although this has no effect on functionality, both of the `formatTime` functions should use 24 hour time for consistency

autoscheduler/frontend/src/tests/ReduxAvailability.test.tsx
```diff
@@ -155,6 +155,61 @@ describe('Availabilities', () => {
       // assert
       expect(store.getState().availability).toEqual(expected);
     });
+
+    // doesn't seeem to be doing anything
```
ryan-conn: Why is this comment here? Does this test not properly fail with the old code (since the problem seems inconsistent to replicate) or something else

firejake308: old comment, should've been removed when i fixed it

autoscheduler/frontend/src/components/Schedule/Schedule.tsx
firejake308: oops my bad

autoscheduler/frontend/src/tests/AvailabilityUI.test.tsx
```diff
@@ -0,0 +1,360 @@
...
+      const {
+        getByText, getByLabelText, queryByText, queryByLabelText,
+      } = render(
+        <Provider store={store}>
+          <Schedule />
+        </Provider>,
+      );
+      const startEventProps = {
+        button: 0,
+        clientY: 500,
```
ryan-conn: Can you abstract this into a function to convert a time into clientY (similar to `eventToTime` in schedule but the other way around)? It would make it more straightforward to tell what times are being dragged rather than just looking at the expected start/end times

firejake308: Good idea, I'll try to get it done after classes tonight

autoscheduler/frontend/src/tests/AvailabilityUI.test.tsx
```diff
@@ -200,7 +203,7 @@ describe('Availability UI', () => {
       };
       const endEventProps = {
         button: 0,
-        clientY: 100,
+        clientY: 110,
       };
```
ryan-conn: There are a few manually calculated eventProps like this, is there a reason this can't use `timeToEvent` along with the rest of them? It seems to break `queryByLabelText` (doesn't find a DOM element matching the query) but since this should be the same as what `timeToEvent` returns I don't see why it should matter

firejake308: I chose to leave some of them because those tests focus more on the actual position of the mouse rather than the time. I think you're right about this one, though, so I converted it to use `timeToEvent`. The reason it was throwing an error had to do with the missing `offset`, so I fixed that. Now, the only test that doesn't use`timeToEvent` is the one that checks for correct behavior when dragging below the bottom, which I think just makes more sense if you see the clientY's directly

gannonprudhomme:
Also just small fixes like `Miscellaneous fixes from code review`

Issue #153, opened 2020-02-25
Split up Redux files into separate files


gannonprudhomme:
Per #151, we need to split up the Redux files into logical pieces. For instance, in `redux/availability`, we would have a `reducers.ts`, `actions.ts`, etc so each piece can be logically separated which should make things much more organized and easier to read.

firejake308:
Oops for accidentally coding the issue. Do y'all want this done before or after #151 is merged?

gannonprudhomme:
I'd def say after #151 just cause we'd have to make the changes all over again

PR #164, opened 2020-03-20
Frontend/redux rearrangement

firejake308:
I was able to refactor the `redux` folder to make separate files for the action types and reducers of each slice of the Redux state. However, I elected to leave all of the actions together in one file because the two thunk-ified actions, `updateCourseCard` and `mergeThenSelectAvailability` require access to the `RootState` defined in `reducer.ts`. However, `reducer.ts` requires each of the state-slice files, so including thunk-ified actions in the state-slice files would lead to a circular dependency.

One alternative solution would be to move all of the non-thunkified actions into the state-slice files and leave the thunkified actions together in a separate file, but this would be un-intuitive to use since non-thunkified and thunkified actions appear identical to the user and I wouldn't know where to import the action from.

Another solution would be to just leave the Redux as it is. Personally, I am leaning toward this solution, as it allows us to commit to one single organizational strategy rather than compromising between two.

Closing #153


gannonprudhomme:
Since this is based off of `frontend/availability-ui` for now, and it's a draft PR, I'm gonna change the base branch of the PR to to `availability-ui` just so we can focus on the actual changes in this

gannonprudhomme:
Sorry for the delay, but I would say I like how this is formatted so this is 👍 from me.

firejake308:
Okay, I have rebased this to work with the changes from #160

firejake308:
That should work. Will implement when I get a chance.

firejake308:
One thing to note is that action types are currently in the reducers folder, not the actions folder. I don't know that it really makes much difference, as both actions and reducers reference the action types

firejake308:
Also force push was because I forgot to finish refactoring everything, so I just redid the last commit

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -0,0 +1,81 @@
+import { ThunkAction } from 'redux-thunk';
+import fetch from '../testData';
+import { CourseCardOptions, SectionSelected } from '../../types/CourseCardOptions'; import {
+  AddCourseAction, ADD_COURSE_CARD, RemoveCourseAction, REMOVE_COURSE_CARD, UpdateCourseAction,
+  UPDATE_COURSE_CARD,
+} from '../reducers/courseCards'; import { RootState } from '../reducer'; import Meeting from '../../types/Meeting';
```
gannonprudhomme, line 6:
```suggestion
} from '../reducers/courseCards'; 
import { RootState } from '../reducer'; 
import Meeting from '../../types/Meeting';
```

Don't think this was intentional, but these should be separated on different lines

PR #166, opened 2020-03-24
Implemented course picker

### LINKED ISSUE #123: Implement course picker
In order to select classes in their schedule, users will first see a single course select card to select a course, then be able to click a button to add a class/course select card to their schedule. I made a quick draft of what it should look like once a few classes have been added here (note that this component should be able to scroll):
image
This is by no means a final design, and we don't have to use the design, I just wanted to make something to explain the concept better

gannonprudhomme:
Here's what it looks like with & without the scroll bar right now:

chrome_IWJabQOiTk
chrome_5DHIWFLFNY

It currently shrinks the course cards when the scrollbar appear, which is something we may want to fix (although I have no idea how to).

Was considering adding a test that tested if the scrollbar appeared after enough course cards were added, but wasn't sure if that's actually something that's realistic to test.

Still some quality-of-life stuff I'd like to add to the course card that we discussed in #123, but I think it's fine if we delay doing them until later.

Closes #123


autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
```diff
@@ -48,15 +51,21 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({ onCollapse, id
   return (
     <Card>
       <div className={styles.header}>
-        <div className={styles.headerGroup}>
+        <div
+          className={styles.headerGroup}
+          onClick={(): void => onRemove(id)}
```
firejake308: When this button is clicked, `ExpandedCourseCard` will actually call the `onRemove` function given to it in its props, which is actually an anonymous function in `CourseSelectCard.tsx`. That anonymous function then calls the `onRemove` function given to it in *its* props, which is actually an anonymous function in `CoursesSelectColumn.tsx`. This is actually a design pattern in React, and it's called "prop drilling," because the actual function we want to call in `CourseSelectCard` is separated from the event handler in `ExpandedCourseSelectCard`, so we drill it through `CourseSelectCard` in the form of props.

However, in this case, we are not actually piercing the heavens with our drill, as this drilling is rather unnecessary. It makes more sense to just move the `dispatch` call into `ExpandedCourseCard`, which gets rid of all of the unneccesary `onRemove` props.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
```diff
@@ -48,15 +51,21 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({ onCollapse, id
...
+        >
           <RemoveIcon />
           Remove
         </div>
         <div
           className={styles.headerGroup}
           onClick={(): void => onCollapse(course)}
           role="button"
-          tabIndex={0}
+          tabIndex={-1}
```
firejake308: If the remove button is tab index 0, shouldn't the collapse button be tab index +1?

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -0,0 +1,36 @@
+.container {
+    width: 20%;
```
firejake308: Before we merge this with master, let's make this width 100%. That way, I can put it inside of a container when I'm composing the master layout and let that dictate the width of the column

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -0,0 +1,36 @@
...
+}
+
+/* Having this wrapper lets us fix the top & bottom of the .course-select-column of the page
+   so the picker scrolls "under" the button while still being within the bounds of the
+   page
+*/
+.column-wrapper {
+    position: relative;
+    width: 100%;
+    height: 93%;
```
firejake308: 93% seems like an aribtrary number. I believe this may the reason why the Add Course Button gets cut off on my screen.

image

Try replacing it with `calc(100% - 52px)`. This gives a better result on my screen, and it should theoretically account for differences across all screen sizes, because it leaves excalty 52px for the button regardless of screen height.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const rows: JSX.Element[] = [];
+
+  // Add all of the course cards to rows to be displayed
+  for (let i = 0; i < courseCards.numCardsCreated; i++) {
+    if (courseCards[i]) {
+      rows.push(
+        <div
+          className={styles.row}
+          key={`${i} stuff`}
```
firejake308: let's make this key `\`${course-card-i}\`` just to keep it looking a little more professional/descriptive

autoscheduler/frontend/src/components/Empty/Empty.tsx
```diff
@@ -1,11 +1,10 @@
 import * as React from 'react';
 import { RouteComponentProps } from '@reach/router';
 import * as styles from './Empty.css';
-import CourseSelectCard from '../CourseSelectCard/CourseSelectCard';
 
 const Empty: React.FC<RouteComponentProps> = () => (
   <div className={styles.empty}>
-    <CourseSelectCard id={0} />
+    Stuff
```
firejake308: lol forgot that this existed

autoscheduler/frontend/src/tests/CourseSelectColumn.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+          <CourseSelectColumn />
+        </Provider>,
+      );
+
+      // act
+      // Press the button
+      act(() => { fireEvent.click(getByText('Add Course')); });
+
+      // Get the course cards
+      const cardsCount = document.getElementsByClassName('MuiPaper-root').length;
```
firejake308: React Testing Library discourages testing using CSS classes since users won't actually see those, and RTL encourages testing components as users will actually see/use them. Consequently, I recommend using some variant of `getAllByText` instead

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
gannonprudhomme: Tbh I made this change but I think it got lost somewhere along the way when I was squashing/rebasing. Will add

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
gannonprudhomme: Yeah that makes sense. I'll leave it how it is rn and will remove it at the end (along with the `/course` route definition in `App.tsx`)

autoscheduler/frontend/src/tests/CourseSelectColumn.test.tsx
gannonprudhomme: Yeah that makes total sense. Tbh I was struggling with how to select it in the first place and was just waiting on what you suggested, will fix.

autoscheduler/frontend/src/components/Empty/Empty.tsx
gannonprudhomme: Yeah I didn't notice until changing the props for CourseSelectCard that we still had it. We might as well delete it

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
gannonprudhomme: It's totally an arbitrary number lol. I figured I had to do some kind of calculation but didn't know where to start really.

Setting that it to what you suggested made it touch the bottom for both 1080p & 1440p (1440p's even lower down than 1080)

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
gannonprudhomme: While that makes sense to me, whenever I tried to make it a positive value I got the eslint error: `Avoid positive integer values for tabIndex.eslint(jsx-a11y/tabindex-no-positive)` (documentation for the rule [here](https://github.com/reactjs/react-a11y/blob/master/docs/rules/tabindex-no-positive.md))

Turns out that they should both be zero 0? That's the only way both Remove and Collapse can be both be selected with tab for me.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
gannonprudhomme: Also in that same vein I also need to change the `CourseSelectColumn` to not use the `RouteComponentProps` since it's only needed to be a route for this PR.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.css
```diff
@@ -26,6 +26,7 @@
     align-items: center;
     cursor: pointer;
     user-select: none;
+    margin: 3px;
 }
 
 .content {
```
ryan-conn: Change these properties in `.content` to make it so when a course has too many sections, it will overflow (I can't actually make suggestions because you didn't edit these lines):
```css
{
    padding: 8px;
    overflow: auto;
    max-height: 400px; /* could be more/less, see how you like different values of this */
}
```
Also remove the `margin` property so that the scrollbar will show on the right edge rather than 8px inside the card.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
gannonprudhomme: To be honest I have no idea what the best way to remedy this, as every solution I have tried gives a different distance from the Add Course button and the button between 1080p and 1440p. The only thing I can think of is that we could fix the bottom of the Add Course button to a certain amount of pixels above the bottom (similar to what we do with the cards column), then there may be a marginal difference between the heights of the column between screens, but as long as it falls at or below the top of the Add Course button it shouldn't matter? But I really have no idea how well that would work.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -0,0 +1,65 @@
...
+        size="medium"
+        variant="contained"
+        id={styles.addCourseButton}
+        onClick={(): void => {
+          dispatch(addCourseCard({ customizationLevel: CustomizationLevel.BASIC, sections: [] }));
+        }}
+      >
+        Add Course
+      </Button>
+    </div>
```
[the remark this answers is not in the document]
gannonprudhomme: What div do you want it to be inside of exactly? The width is originally set by `.container` so they should have the same width.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.css
gannonprudhomme: Yeah this is a good point, we should definitely limit a course card's height as we don't want one course with a bunch of sections to take up a massive portion of the column. That being said, it might look weird having two scroll bars right next to each other. I'll try to implement it and report back

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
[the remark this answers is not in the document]
ryan-conn: I meant the div containing all the cards (so when the scrollbar appears it has the same width as the other cards), but you could just add a top margin to the button so it doesn't touch the scrollbar/course cards and keep the original width, since it might be awkward to try to anchor it to the bottom of that div and I don't think we necessarily have to match the widths of the course cards and add course buttons.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -0,0 +1,62 @@
...
+          key={`courseSelectCardRow-${i}`}
+        >
+          <CourseSelectCard
+            key={`courseSelectCard-${i}`}
+            id={i}
+          />
+        </div>,
+      );
+    }
+  }
```
ryan-conn: image
Could you remove the bottom padding from the last course card so it touches the bottom of the div it's in?

gannonprudhomme: Like you want it so that it touches the Add Course button?

firejake308: I think remove the padding from the last course card using a CSS pseudo-selector (last-of-kind or something) and then add marginBottom to the entire container div

ryan-conn: Yeah I meant the way Adel said, so that we still have the spacing between the end of the course card component and the add course button, and the scroll bar doesn't touch the button either

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.css
gannonprudhomme: So I tried this two different ways. One by adding the scrollbar to the entire card, then another by adding it on the `SectionSelect` component rather than the entire card. It doesn't happen as much with the `SectionSelect` one, but it's kind of annoying having two scroll bars since you basically have to hover wherever the course-card scroll area is _not_, which is a pretty small area when you have it in section select.
 
Top picture is what ryan suggested, bottom is it on the `SectionSelect`.
chrome_RSWrLiS23h
dPfBk8hWax

gannonprudhomme: Although putting it on the `SectionSelect`'s List does make it so that the first prof name won't scroll with the rest of the section rows, any idea what may be causing this @firejake308 ?

chrome_XIlM6VIYGK

ryan-conn: This is what it should look like if you make the changes I suggested:
image
I think you forgot to remove the margin, this is what `.content` looks like:
```css
.content {
    padding: 8px;
    overflow: auto;
    max-height: 200px;
    align-self: stretch;
    display: flex;
    flex-direction: column;
    color: black;
}
```

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
firejake308: Check out #171 for my proposed fix to this

gannonprudhomme: Merged #171, resolving

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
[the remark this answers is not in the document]
gannonprudhomme: Was this fixed by #171?

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -0,0 +1,36 @@
...
+    overflow-y: auto;
+    width: 100%;
+
+    padding-top: 1%;
+    padding-left: 1%;
+}
+
+.row {
+    padding-bottom: 10px;
+    padding-right: 5%;
```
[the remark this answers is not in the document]
gannonprudhomme: I also think the padding-right should probably be on the entire column, rather than the individual cards so I'm gonna change it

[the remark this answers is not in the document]
gannonprudhomme: So in regards to dynamically determining if there's a scroll bar: I have it implemented, but it's always "one step behind", so it won't actually detect the scroll bar until I add a course card when there is a scroll bar. I believe this is because I'm loading in the previous rendering's `course-select-column` element as it do it before we returning the new components to be rendered.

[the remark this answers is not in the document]
gannonprudhomme: Changed the padding to 2% though

gannonprudhomme:
Still some unresolved comments, but I think this is ready for the next round of reviews

firejake308:
When does the internal scrollbar appear?

gannonprudhomme:
> When does the internal scrollbar appear?

So there isn't enough sections / meetings for it with the test data that's given, but I'll send one in the Discord that I'm using that should make it appear.

gannonprudhomme:
> Spacing looks a lot better on this version. I generally don't like scroll bars inside scroll bars, but unless we can identify a better solution, this is how I would want it to look and feel. I put up one little bugfix in #173, but other than that and the font and fetching fixes for CourseCard, I'm okay with this

Yeah I definitely agree that scroll bars inside scroll bars should be avoided, but I think it's fine leaving it like this for now, and we can come back to it and improve/change it once we get all of the basics implemented.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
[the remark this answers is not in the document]
ryan-conn: Yes, this looks how I wanted it to now

autoscheduler/frontend/src/tests/CourseSelectColumn.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+
+import { render, fireEvent, act } from '@testing-library/react';
+import { Provider } from 'react-redux';
+import { createStore, applyMiddleware } from 'redux';
+import thunk from 'redux-thunk';
+import autoSchedulerReducer from '../redux/reducers';
+import CourseSelectColumn from '../components/CourseSelectColumn/CourseSelectColumn';
+
+describe('CourseSelectColumn', () => {
+  describe('Clicking the Add Course button', () => {
```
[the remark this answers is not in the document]
gannonprudhomme: idk why I did it this way but I do agree we should do it the way you mentioned, with the expected situation in the `describe` as it's most likely more flexible. That's also what I suggested Ryan do with his tests for #175 in Discord so idk what I was thinking lol

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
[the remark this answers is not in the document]
ryan-conn: Resolving this since #175 was merged

gannonprudhomme:
Since I think this is close to being ready to be merged, I'm going to change the column width to 100% like Adel suggested, as well as remove the `RouteComponentProps` from CourseSelectColumn and make it not be a route in `App.tsx`.

Will also rebase it with `frontend/master`.

gannonprudhomme:
Rebased to be up to date with `frontend/master`. Going to squash the commits that have "Squash into ..." in them now, then will add the last set of changes that I mentioned above

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -0,0 +1,81 @@
+import * as React from 'react';
+import { RouteComponentProps } from '@reach/router';
```
firejake308, line 2: Approving, but remember to fix this linter warning before merging

gannonprudhomme: To be honest I thought the GitHub Actions frontend linting step would fail with warnings (so I assumed I was good lol), is that something we want to add?

gannonprudhomme:
Fixed linting issue & squashed it, merging now

PR #167, opened 2020-03-26
Implemented scheduling algorithm

ryan-conn:
I implemented course scheduling using a brute force algorithm, as this seems to be the most efficient way to do so: using a greedy algorithm is impossible since intervals contain multiple intervals.

First, random combinations of classes are generated (so that the schedules will be diverse rather than in lexographical order, likely leading to many schedules with the same first 1-3 sections), until the desired number of schedules are generated, it's probable enough that a combination will be checked again, or the set containing checked combinations gets too large. This is controlled by the value `threshold`, the max number of random combinations to check.

After that, every possible combination is tried in order to generate the remaining schedules.

One thing to note is that we should add an index on `Meeting.section_id`: in order to fetch and efficiently group meetings for multiple sections, they have to be in order of `section_id` so that `group_by()` can create a single iterable for each section.

I haven't made any tests yet (because I'm lazy), but I've manually checked quite a few schedules and they all seem to be valid: likewise, all the combinations of sections that are found to be invalid seem to actually be invalid as well.

Right now you have to manually enter the term/courses into the command, I figured that since this will only be used in our API instead of actually being a command, parsing arguments would be too much effort to be worth it.


autoscheduler/scheduler/management/commands/scheduling.py
```diff
@@ -0,0 +1,148 @@
...
+        A dict of sections for the course with the meeting number as the key
+        and attributes for each meeting as the value
+    """
+    subject, course_num = course
+    # Create list of section_nums matching desired course
+    sections = (Section.objects.filter(course_num=course_num, subject=subject,
+                                       term_code=term)
+                .values('id', 'section_num'))
+    # Dictionary for translating meeting ids into section numbers
+    meeting_nums = {section['id']: section['section_num'] for section in sections}
```
ryan-conn:
```suggestion
    section_nums = {section['id']: section['section_num'] for section in sections}
```
Why did I call this meeting nums when it's section numbers? Who knows but I'm not going to make a commit just to fix a variable name

ryan-conn:
Something to note: I can probably reduce the number of DB queries to 1 instead of the number of courses by using `group_by` on sections instead of just meetings, but I think this will be

ryan-conn:
I'm done writing tests for this, so I'm changing it from a draft to an actual PR. This should be ready for review now.

ryan-conn:
I changed the tests to make their own models, so this should be ready for you to review now.
Also pylint isn't upset about duplicate code anymore even though there's even more than before, 10/10 good job pylint

ryan-conn:
Force pushed to make each test only make its own meetings instead of all the models

ryan-conn:
Force pushed to make each test only make its own meetings instead of all the models

ryan-conn:
Force pushed to make each test only make its own meetings instead of all the models

ryan-conn:
Just realized it didn't actually push because github went down when I was pushing, so I did it again

gannonprudhomme:
Still need to make it take in a `List[str]` of specific sections that the user wants to take. I'd say change the courses input from `List[Tuple[str, str]]` to `List[Tuple[str, str, List[str]]]` but it's up to you

ryan-conn:
Should be ready for review now that there are filters for times, honors, web, and section numbers

autoscheduler/scheduler/tests/scheduling_tests.py
```diff
@@ -0,0 +1,520 @@
+""" Tests all of the models functions """
```
gannonprudhomme, line 1: I'm not sure the best method to do it, but the fact that GitHub doesn't load this files diff by default since it's so large makes me think we should def split this up into different files.

autoscheduler/scheduler/utils.py
```diff
@@ -0,0 +1,70 @@
+from datetime import time
+from functools import reduce
+from operator import mul
+from random import sample
+from typing import Any, Iterable, List, NamedTuple, Tuple, Union
+
+def random_product(*iterables: Iterable[Iterable], limit=100_000) -> Tuple[Any]:
```
gannonprudhomme, line 7: Just a thought, but I thought it might be nice if we somehow used the `CourseFilter` input to be the seed for this, so if on the frontend they did the exact same scheduling-generation query twice it would return the same thing

autoscheduler/scheduler/tests/scheduling_tests.py
```diff
@@ -0,0 +1,520 @@
...
+        expected_schedules = set(((2, 5),))
+        num_expected_schedules = 1
+
+        # Act
+        schedules = create_schedules(courses, term, unavailable_times, num_schedules=10)
+        valid_generated_schedules = expected_schedules.intersection(schedules)
+
+        # Act
+        self.assertEqual(len(schedules), num_expected_schedules)
+        self.assertEqual(len(valid_generated_schedules), len(schedules))
```
gannonprudhomme, line 457: Should probably just compare the `schedules` with `expected_schedules`

autoscheduler/scheduler/tests/scheduling_tests.py
```diff
@@ -0,0 +1,520 @@
...
+        # Arrange
+        course = CourseFilter("CSCE", "123")
+        term = "201931"
+        unavailable_times = []
+
+        # Act
+        meetings = _get_meetings(course, term, unavailable_times)
+
+        # Assert
+        self.assertFalse(meetings)
```
gannonprudhomme, line 122: Is this assert the same thing as doing `self.assertEqual(len(meetings), 0)`?

autoscheduler/scheduler/create_schedules.py
```diff
@@ -0,0 +1,137 @@
...
+            first_end = first_meeting.end_time
+            if first_start is None or first_end is None:
+                continue
+            for second_meeting in checking_section:
+                second_start = second_meeting.start_time
+                second_end = second_meeting.end_time
+                # Only compare meetings if they have valid times and share a day
+                if second_start is None or second_end is None:
+                    continue
+                if first_meeting.meeting_days.isdisjoint(second_meeting.meeting_days):
```
gannonprudhomme, line 85: I would explain this `isdisjoint` call briefly above this line

gannonprudhomme:
Oh yeah we should also probably deal with excluding sections that don't have any seats left

autoscheduler/scheduler/tests/scheduling_tests.py
ryan-conn: Yeah, iterables evaluate as False only if they're empty

ryan-conn:
I made all of your requested changes except for:
>I'm not sure the best method to do it, but the fact that GitHub doesn't load this files diff by default since it's so large makes me think we should def split this up into different files.

Since I moved the `random_product` tests to a different file, `scheduling_tests.py` is a bit shorter now so let me know if you think they should be separated more

gannonprudhomme:
Just need to change the "include full sections" passing in from `CourseFilter` to its own parameter in `create_schedules` then I think this should be good to go.

PR #175, opened 2020-04-02
Made course card padding depend on if there's a scrollbar or not

ryan-conn:
I made the course cards in the course select column change their padding based on whether there is a scroll bar in the component or not, instead of always keeping 2% padding. Here's what it looks like:
dynamic width
I'm not sure if the way I select elements is the best way to do so (`getElementById` and `getElementsByClassName`). As far as I can tell, we would have to change this to a class component to use refs in order to modify CSS after rendering, but if there's a better way to do so I'd like to know.

### DIFF
--- a/autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
+++ b/autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
@@ -13,7 +13,7 @@ import { addCourseCard } from '../../redux/actions';
  */
 
 const updateCourseCardPadding = (): void => {
-  const container = document.getElementById('courseSelectContainer');
+  const container = document.getElementById('course-select-container');
   const hasScrollBar = container.scrollHeight > container.clientHeight;
   const cards = Array.from(document.getElementsByClassName(styles.row) as
                            HTMLCollectionOf<HTMLElement>);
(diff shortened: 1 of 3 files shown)


ryan-conn:
Going to fix the linter errors now, I think my local linter broke because it didn't catch any of them

ryan-conn:
Linter errors fixed via good ol' force push so hard reset if you checked this branch out

ryan-conn:
Tests created for checking that this works correctly when changing the course cards or resizing the window. I believe these are the only events that could change whether or not to add padding.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -40,7 +61,7 @@ const CourseSelectColumn: React.FC<RouteComponentProps> = () => {
   return (
     <div className={styles.container}>
       <div className={styles.columnWrapper}>
-        <div className={styles.courseSelectColumn}>
+        <div className={styles.courseSelectColumn} id="courseSelectContainer">
```
firejake308, line 64: Let's use `hyphenated-names` for CSS, just because that's what I see more often

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -20,6 +32,15 @@ const CourseSelectColumn: React.FC<RouteComponentProps> = () => {
 
   const rows: JSX.Element[] = [];
 
+  // Recalculate padding of course cards after rendering component or resizing window
+  React.useEffect(() => {
```
firejake308, line 36: Could this be simplified to `React.useEffect(updateCourseCardPadding, [rows]);`?

ryan-conn: Yes, I changed it to that now

ryan-conn:
I implemented the changes you suggested and that Gannon posted on discord

gannonprudhomme:
I'd also say to squash that commit into the previous, just b/c we don't really want a commit title "Implemented suggestions" in the commit history lol

Issue #138, opened 2020-02-08
Change scraping to use bulk_create


gannonprudhomme:
After testing deploying to the server using Google Cloud Platform, I discovered that saving the courses individually takes a very long time (over an hour, while the actual scraping takes < 1 second).

In order to remedy this, we can add each model to a list of their own type, then call `Model.objects.bulk_create(models_list)`

This reduced the time to save all of the courses to 9 seconds in GCP and 3 seconds locally (locally it normally took ~25 seconds).

Likewise for department scraping on GCP, this changed the time to save from 25 seconds to < 1 second.

gannonprudhomme:
We should be able to do this and #143 at the same time

PR #177, opened 2020-04-05
Improvements to course scraping

ryan-conn:
Making the scheduling algorithm, I realized that we don't actually scrape whether sections are web or online (which I need to test section filtering), so I added that functionality. I also made slight improvements to `parse_course` (see my commit messages) fixed a bug with honors section titles (which start with `HNR-`) being parsed directly, and added tests for these new functionalities.

### DIFF
--- a/autoscheduler/scraper/tests/commands/scrape_courses_tests.py
+++ b/autoscheduler/scraper/tests/commands/scrape_courses_tests.py
@@ -19,6 +19,12 @@ class ScrapeCoursesTests(django.test.TestCase):
                                      if course["id"] == 491083)
         self.engl_section_json = next(course for course in course_list
                                       if course["id"] == 511984)
+        self.pols_section_json = next(course for course in course_list
+                                      if course["id"] == 469982)
+        self.acct_section_json = next(course for course in course_list
+                                      if course["id"] == 467471)
+        self.csce_web_section_json = next(course for course in course_list
+                                          if course["id"] == 515269)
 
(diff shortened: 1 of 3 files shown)


gannonprudhomme:
While we're here, we might as well make the course scraping use `bulk_create` and knock out #138.

Downside to this is that I think the tests will have to be refactored, and we might need to make the `parse_` functions return the models in some cases.

I'd also say to just merge this into `backend/master` rather than your scheduling branch, so these changes won't be in the changed files for the scheduling algorithm.

gannonprudhomme:
So talked to Ryan about it a bit more, and we're gonna just wait to do #138 in a different issue so we can merge this for use in the scheduling PR, so scratch out my above comment.

autoscheduler/scraper/tests/commands/scrape_courses_tests.py
```diff
@@ -105,6 +113,65 @@ def test_parse_meeting_does_save_model_correct(self):
                             start_time=begin_time, end_time=end_time,
                             meeting_type=meeting_type, section=section)
 
+    def test_parse_meeting_unescapes_building(self):
+        """ Tests that parse meeting converts escaped HTML characters like "&amp;" into
+        their actual buildings
```
gannonprudhomme, line 118:
```suggestion
        """ Tests that parse meeting converts escaped HTML characters like "&amp;" into
            their actual buildings
```

Need to indent this correctly (squash into prev commit)

Issue #181, opened 2020-04-07
Hovered time doesn't disappear


firejake308:
As I was working on the scheduling page, I noticed that the `<HoveredTime />` component does not disappear when the cursor leaves the `<Schedule />` component. This does not look very good, so it should be fixed when we get a chance.

PR #188, opened 2020-04-09
Scheduling page

firejake308:
The initial layout for the scheduling page is more or less complete at this point. A few things remain in terms of fetching schedules, but that can be handled later. Also, I have elected not to implement tests for the scheduling page at this time, since the individual units are unit-tested. If you feel that this is unwise, however, we can change that and add integration tests as well.


firejake308:
Updated so this should be merged after #196

firejake308:
This also closes #181

gannonprudhomme:
Update this with `master` when you can and I'll begin review

gannonprudhomme:
You might need to reset hard to master and cherry pick some of these commits, as the `OptionsCard` is in here even though you renamed it to `ConfigureCard` in #196

autoscheduler/frontend/src/tests/redux/Schedule.test.tsx
```diff
@@ -0,0 +1,228 @@
+import { createStore } from 'redux';
```
gannonprudhomme, line 1: We should also probably have some basic UI tests for the entire Schedule page, i.e. clicking on a schedule makes the respective section(s) show up on the Schedule, etc

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -231,46 +232,52 @@ const Schedule: React.FC<RouteComponentProps> = () => {
...
-        lastHour={LAST_HOUR}
-      />
+  // let's see if useMemo reduces our render time
+  const meetingsForDays = React.useMemo(() => {
+    const uniqueSections = [...new Set(schedule.map((mtg: Meeting) => mtg.section.id))];
+    // build each day based on schedule
+    function getMeetingsForDay(day: number): Meeting[] {
+      // meetingDays = UMTWRFS
+      // day = MTWRF
+      return schedule.filter((meeting) => meeting.meetingDays[day + 1]);
```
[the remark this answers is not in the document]
firejake308: Nice catch

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -231,46 +232,52 @@ const Schedule: React.FC<RouteComponentProps> = () => {
...
+    const uniqueSections = [...new Set(schedule.map((mtg: Meeting) => mtg.section.id))];
+    // build each day based on schedule
+    function getMeetingsForDay(day: number): Meeting[] {
+      // meetingDays = UMTWRFS
+      // day = MTWRF
+      return schedule.filter((meeting) => meeting.meetingDays[day]);
+    }
+    function renderMeeting(meeting: Meeting): JSX.Element {
+      const colors = ['#500000', '#733333', '#966666', '#b99999', '#dccccc',
+        '#871b1e', '#9f494b', '#b77678', '#9a1d26', 'c2777d'];
```
gannonprudhomme, line 246: could you add this amount of sections/meetings to the test data so we can see the different colors?

autoscheduler/frontend/src/types/DayOfWeek.ts
```diff
@@ -3,3 +3,5 @@ enum DayOfWeek {
 }
 
 export default DayOfWeek;
+export const DAYS_MWF = [true, false, true, false, true, false, false];
+export const DAYS_TR = [false, true, false, true, false, false, false];
```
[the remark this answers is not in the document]
firejake308: Makes sense

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
firejake308: Done. Also added a test for it in `ui/Schedule.test.tsx`

autoscheduler/frontend/src/tests/redux/Schedule.test.tsx
```diff
@@ -0,0 +1,228 @@
...
+      store.dispatch(addSchedule(schedule1));
+      store.dispatch(addSchedule(schedule2));
+
+      // assert
+      expect(store.getState().schedules).toContainEqual(schedule1);
+      expect(store.getState().schedules).toContainEqual(schedule2);
+    });
+  });
+  describe('removes schedules', () => {
+    test('when the middle schedule is deleted', () => {
```
ryan-conn, line 146:
```suggestion
    test('when the first schedule is deleted', () => {
```

ryan-conn, line 146:
```suggestion
    test('when the first schedule is deleted', () => {
```

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -1,23 +1,24 @@
...
-const Schedule: React.FC<RouteComponentProps> = () => {
+const Schedule: React.FC = () => {
   // these must be unique because of how they're used below
   const DAYS_OF_WEEK = ['M', 'T', 'W', 'R', 'F'];
 
   // "props" derived from Redux store
-  const schedule = useSelector<RootState, Meeting[]>((state) => state.meetings);
+  const schedule = useSelector<RootState, Meeting[]>(
+    (state) => state.schedules[state.selectedSchedule] || [],
+  );
```
ryan-conn, line 21:
```suggestion
  const emptySchedule: Meeting[] = [];
  const schedule = useSelector<RootState, Meeting[]>(
    (state) => state.schedules[state.selectedSchedule] || emptySchedule,
  );
```
Since this returns a reference to a new list every time, it will rerender the schedule every time the state is changed if there isn't a schedule selected, which can be prevented if we create the reference beforehand

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
```diff
@@ -0,0 +1,74 @@
...
+import { createStore } from 'redux';
+import { Provider } from 'react-redux';
+import {
+  render, fireEvent, waitForDomChange, queryByText as queryContainerByText,
+} from '@testing-library/react';
+import autoSchedulerReducer from '../../redux/reducer';
+import SchedulingPage from '../../components/SchedulingPage/SchedulingPage';
+
+describe('Scheduling Page UI', () => {
+  describe('shows an error message', () => {
```
ryan-conn, line 11:
```suggestion
  describe('indicates that there are no schedules', () => {
```
Something like this, the way it's worded now confused me at first since it isn't really an error message, just placeholder content

ryan-conn, line 11:
```suggestion
  describe('indicates that there are no schedules', () => {
```
Something like this, the way it's worded now confused me at first since it isn't really an error message, just placeholder content

autoscheduler/frontend/src/tests/redux/Schedule.test.tsx
ryan-conn, line 146:
```suggestion
    test('when the first schedule is deleted', () => {
```

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -130,7 +131,10 @@ const Schedule: React.FC<RouteComponentProps> = () => {
       time2 = 8 * 60 + 0;
     } else if (evt.clientY > evt.currentTarget.getBoundingClientRect().bottom) {
       time2 = 21 * 60 + 0;
-    } else {
+    } else if (
+      // these conditions ensure that the time disappears if the mouse leaves to the side
+      evt.clientX >= evt.currentTarget.parentElement.getBoundingClientRect().left
```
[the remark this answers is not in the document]
firejake308: 2 of them are actually for `currentTarget`, and 2 are actually for the `parentElement`. I think I'll go ahead and make local variables for readability/brevity anyway, but it might not help performance much

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -231,46 +232,52 @@ const Schedule: React.FC<RouteComponentProps> = () => {
...
+  const meetingsForDays = React.useMemo(() => {
+    const uniqueSections = [...new Set(schedule.map((mtg: Meeting) => mtg.section.id))];
+    // build each day based on schedule
+    function getMeetingsForDay(day: number): Meeting[] {
+      // meetingDays = UMTWRFS
+      // day = MTWRF
+      return schedule.filter((meeting) => meeting.meetingDays[day]);
+    }
+    function renderMeeting(meeting: Meeting): JSX.Element {
+      const colors = ['#500000', '#733333', '#966666', '#b99999', '#dccccc',
```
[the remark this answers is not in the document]
firejake308: Material color tool agrees with you. Will implement suggestion.

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -231,46 +232,52 @@ const Schedule: React.FC<RouteComponentProps> = () => {
...
+        <MeetingCard
+          meeting={meeting}
+          bgColor={colors[uniqueSections.indexOf(meeting.section.id) % colors.length]}
+          key={meeting.id}
+          firstHour={FIRST_HOUR}
+          lastHour={LAST_HOUR}
+        />
+      );
+    }
+    return [0, 1, 2, 3, 4].map(
```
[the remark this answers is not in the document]
firejake308: Done for lines 257 and 277, but on line 16, that array is used somewhat differently, so I have left it as is for now. Line 16 uses an array of the one-letter keys of the days rather than an array of the numerical keys of the days. This would complicate the usage of the `DayOfWeek` enum, which has only numerical values.

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
firejake308: Hmm, I think the local variable `emptySchedule` would also be re-created on every render, so I've used a static variable instead. That look good to you?

Also, once we have basic functionality, I think it would be interesting to write tests for optimization, to check if expensive calculations were being repeated on re-renders. Just a thought for now.

ryan-conn: From testing locally (by putting what I suggested and `console.log`ging something in the `useMemo` for meetingsForDays to check if it rerenders), it seems to still work fine if you put `const emptySchedule` inside of the functional component, but since it's pretty much the same and you already put it outside it's fine the way you changed it to

autoscheduler/frontend/src/tests/ui/Schedule.test.tsx
```diff
@@ -37,24 +38,70 @@ const testMeeting1 = new Meeting({
...
-    <Provider store={store}>
-      <Schedule />
-    </Provider>,
-  );
-
-  // assert
-  expect(container).toBeTruthy();
+      // assert
+      // there are 10 unique colors
+      expect(uniqueSections.size).toBeLessThanOrEqual(10);
```
gannonprudhomme, line 96: Shouldn't this be `toBeGreaterThanOrEqual(10)`? Otherwise this would pass when there are less than 10 unique colors

firejake308: I'm actually using this to assert that the test data hasn't been changed to add more than 10 sections. Because if that were the case, then I would loop back and start re-using colors.

firejake308: It's more of a pre-condition than a post-condition, so it might even make sense to check this in the arrange phase

gannonprudhomme: yeah i figured it was something weird like that since it was checking `uniqueSections`, if's not actually part of the assert then I'd move it up and add a comment explaining it

autoscheduler/frontend/src/tests/ui/Schedule.test.tsx
```diff
@@ -0,0 +1,107 @@
...
+      const meetingCards = cardEls.map((card) => {
+        // find the colored element that is a parent of the div with text
+        let coloredEl = card;
+        while (!coloredEl.style.backgroundColor) coloredEl = coloredEl.parentElement;
+
+        return {
+          backgroundColor: coloredEl.style.backgroundColor,
+          textContent: card.textContent,
+        };
+      });
```
ryan-conn, line 95: I'm not really a fan of how you're getting all of the meeting cards in this test, we should be able to use [`react-test-renderer`](https://reactjs.org/docs/test-renderer.html) to find components by their react types instead of selecting by text, which should be able to fix this and provide us with some useful queries when writing tests going forward, but let me know what you think

firejake308: What's the difference between react-test-renderer and Enzyme?

firejake308: In the latest commit, I attempt to resolve this issue within the Testing Library framework by using Test ID. Let me know if this is acceptable, or if you would like to move forward with `react-test-renderer`

ryan-conn: That's fine, the reason I didn't say to use test ids is that I remember you saying something along the lines of not wanting to use test ids since they'll only be used for tests, but I'm okay with using them if you are

gannonprudhomme: Tbh I don't like the look of test id's too much, but if we don't use them too often I think it'll be fine

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -231,46 +237,54 @@ const Schedule: React.FC<RouteComponentProps> = () => {
...
+    function getMeetingsForDay(day: number): Meeting[] {
+      // meetingDays = UMTWRFS
+      // day = MTWRF
+      return schedule.filter((meeting) => meeting.meetingDays[day]);
+    }
+    function renderMeeting(meeting: Meeting): JSX.Element {
+      const colors = [
+        '#500000', '#733333', '#966666', '#b99999',
+        '#871b1e', '#9a1d26', '#c2777d', '#9f494b', '#b76778',
+      ];
```
ryan-conn, line 253: Can you move this array to a separate file and export it? This would let us do a couple things:
- Prevent recreating the colors array every time meetings are rendered
- Use the array directly in the `uses unique colors for up to x different sections` test, which would let us assert that the are less unique sections than colors in the array instead of hardcoding the number
  - We could also check that all the background colors are in the array, but the only reason I would see this being helpful is if the the part in the test that calculates `meetingCards` somehow breaks at some point and selects incorrect elements so it's your call if you want to check this or not

firejake308: I elected not to implement the optional part of this change, as the computed style is an `rgb()` string and the provided styles are hex codes and it's too much of a hassle to conver between the two. I did, however, separate out the colors in their own file in the latest commit.

gannonprudhomme:
Could you make the GPA in the schedule previews show 2 decimal places instead of one real quick?

Issue #147, opened 2020-02-18
Connect /api/terms to landing page


gannonprudhomme:
Need to have it so that the landing page retrieves all of the possible terms from /api/terms and displays them in the terms box at the bottom of the page.

PR #189, opened 2020-04-10
Connects landing page to main page

### LINKED ISSUE #146: Connect landing page with the main scheduling page
For the landing page, we need to make it so that once the user clicks a term, it automatically brings them to the main scheduling page (and saves the term in Redux)

gannonprudhomme:
Makes it so that when you select the term on the landing page, it brings you to `/scheduling`, and sets the current term in the Redux store so it can be accessed when we fetch to the backend.

Closes #146

### DIFF
--- a/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
+++ b/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {
   Menu, MenuItem, Button,
 } from '@material-ui/core';
+import { navigate } from '@reach/router';
 import { useDispatch } from 'react-redux';
 import setTerm from '../../../redux/actions/term';
 import * as styles from './SelectTerm.css';
@@ -52,7 +53,7 @@ const SelectTerm: React.SFC = () => {
     dispatch(setTerm({ term }));
 
     // Redirect to the main page
-    window.location.assign('/schedule');
+    navigate('/schedule');
   };
 
   return (
--- a/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
+++ b/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
@@ -9,12 +9,12 @@ import * as styles from './SelectTerm.css';
 
 const options = [
   'None',
-  'Semester 1',
-  'Semester 2',
-  'Semester 3',
-  'Semester 4',
-  'Semester 5',
-  'Semester 6',
+  'Fall 2020',
+  'Summer 2020',
+  'Spring 2020',
+  'Fall 2019',
+  'Summer 2019',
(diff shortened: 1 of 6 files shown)


gannonprudhomme:
Force pushed a small comment fix

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
```diff
@@ -14,19 +16,43 @@ const options = [
   'Semester 6',
 ];
 
+// Maps between the term description and its actual value (which we'll use to store)
+// We might not need this in the future?
```
firejake308, line 20: Is this termMap going to come from the bcakend in the future? Even so, for now, I would set it to something like "Fall 19", "Fall 19(1)", and "Fall 19(2)" instead of "Semester 1" so that we know what test data we're looking at

autoscheduler/frontend/src/redux/reducers/term.ts
```diff
@@ -0,0 +1,30 @@
+/**
+ * Stores the current term
+ */
+
+// type
+export class TermType {
```
firejake308, line 6: I'm not understanding why we need a `TermType` wrapper around a single number. Is `TermType` planned to include multiple fields in the future?

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
```diff
@@ -14,19 +16,43 @@ const options = [
...
-    selectTerm(option);
+    setSelectedTerm(option);
+
+    // Get the corresponding option given the term's description
+    const term: number = termMap[option];
+
+    dispatch(setTerm({ term }));
+
+    // Redirect to the main page
+    window.location.assign('/schedule');
```
firejake308, line 55: This works, but I think it owuld be better to use `navigate` from the [Reach router library](https://reach.tech/router/api/navigate), since that's our standard navigation library

autoscheduler/frontend/src/redux/reducers/term.ts
gannonprudhomme: See I don't either, but frankly I couldn't figure out how to store just a single number in the Redux store, but I'll try again

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
gannonprudhomme: Yeah it'll come from #147 . We might need to change the data structure of it to whatever api/terms uses but we can deal with it in that issue as it won't be that much to change regardless

autoscheduler/frontend/src/redux/reducers/term.ts
gannonprudhomme: Oh nvm I got it working. I swear I could've sworn I tried this exact same thing before but I guess not lmao

PR #196, opened 2020-04-11
New components for scheduling page

### LINKED ISSUE #169: Create Schedule Preview component
We need to create this component:

Figma_GaZDgBZ9IY

This won't include making the schedule preview, but should handle everything else needed for this. Should also include making the Redux types for a generated Schedule.

Also need to decide if there's any extra information we want to add to it, like the professor for the course (which we only have for one of the courses currently)

firejake308:
This PR includes the `OptionsCard` and `SchedulePreview` components for the scheduling page.

Closes #169


autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+      endTime: formatTime(avl.endTimeHours, avl.endTimeMinutes).replace(':', ''),
+      day: avl.dayOfWeek,
+    }));
+
+    fetch('/api/scheduling/generate', {
+      body: JSON.stringify({
+        term: '202011',
+        courses,
+        availabilities,
+      }),
```
gannonprudhomme, line 49:
```suggestion
    fetch('/api/scheduling/generate', {
      method: 'POST',
      body: JSON.stringify({
        term: '202011',
        includeFull,
        courses,
        availabilities,
      }),
```

It won't actually affect anything right now (at least I don't think it will), but since it'll be a POST request we might as well start with it set up like one.

Should also pass along `includeFull` with the request since it'll be part of it when the route is actually implemented (not sure of the correct syntax for it but you get the point)

autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+        >
+          <ListItemIcon>
+            <Checkbox color="primary" checked={includeFull} />
+          </ListItemIcon>
+          <ListItemText>
+            Include full sections
+          </ListItemText>
+        </ListItem>
+        <Button variant="contained" color="primary" onClick={fetchSchedules}>
+          Fetch Schedules
```
gannonprudhomme, line 81:
```suggestion
          Generate Schedules
```

Not sure what I like better, `Generate Schedules` or `Fetch Schedules`. Thoughts @ryan-conn @ThatJuanGuy @firejake308 ?

autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.css
```diff
@@ -0,0 +1,10 @@
+.button-container {
+    display: flex;
+    flex-direction: column;
+    padding: 8px;
+    justify-content: space-between;
+}
```
ryan-conn, line 6:
```suggestion

}
```

autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.css
```diff
@@ -0,0 +1,10 @@
+.button-container {
+    display: flex;
+    flex-direction: column;
+    padding: 8px;
+    justify-content: space-between;
```
ryan-conn, line 5:
```suggestion
    justify-content: space-between;
    user-select: none;
```
I don't think we should enable text highlighting on components that will be repeatedly clicked (like buttons) since it looks bad when it happens

autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const avsList = useSelector<RootState, Availability[]>((state) => state.availability);
+  const dispatch = useDispatch();
+  const fetchSchedules = React.useCallback(() => {
+    // make courses object
+    const courses = [];
+    for (let i = 0; i < courseCards.numCardsCreated; i++) {
+      if (courseCards[i] && courseCards[i].course) {
+        courses.push({
+          subject: courseCards[i].course.split(' ')[0],
+          courseNum: courseCards[i].course.split(' ')[1],
```
ryan-conn, line 30:
```suggestion
        const [subject, courseNum] = courseCards[i].course.split(' ');
        courses.push({
          subject,
          courseNum,
```
Avoids creating the array from `split` twice

autoscheduler/frontend/src/components/GenericCard/GenericCard.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+interface GenericCardProps {
+    header?: JSX.Element;
+    style?: React.CSSProperties;
+}
+
+/**
+ * Makes a generic card with a maroon header. Any children are rendered inside the body, and
+ * elements can be rendered inside the header by setting the header prop
+ * @param props Includes header, children, and minWidth, all optional
```
ryan-conn, line 12:
```suggestion
 * @param props Includes header and children, which are optional
```
Something like this, since `minWidth` isn't a  parameter

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -0,0 +1,67 @@
...
+            </span>
+          ))
+        }
+      />
+    </ListItem>
+  );
+
+  return (
+    <GenericCard
+      style={{ flexGrow: 1, flexBasis: 0, overflowY: 'auto' }}
```
ryan-conn, line 53: I'd rather we make a `SchedulePreview.css` for the styles here, if there were only one or two then inline would be fine imo but since there are 5 I think making a new file for them is worthwhile. Wondering what @gannonprudhomme thinks about this

Same applies for `OptionsCard`, there's already a .css file for it but still 3 inline styles

autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.tsx
ryan-conn: I'm fine with either, I think I'd slightly prefer "Generate schedules" since "fetch schedules" might sound a little weird if you don't get the joke

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
gannonprudhomme: Yeah I agree, not sure about `OptionsCard` but we should def do it for `SchedulePreview`

autoscheduler/frontend/src/components/SchedulingPage/OptionsCard/OptionsCard.tsx
firejake308: I'll change it to Generate Schedules for now, and maybe change it to Fetch later if and when we add more canine imagery

autoscheduler/frontend/src/components/GenericCard/GenericCard.tsx
firejake308: oh lmao i forgot that i changed this

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -0,0 +1,96 @@
...
+  return (
+    <GenericCard
+      header={
+        <div id={styles.cardHeader}>Configure</div>
+      }
+    >
+      <div className={styles.buttonContainer}>
+        <div id={styles.instructions}>
+          Click and drag in the calendar on the right to block off times when you
+          are unavailable, then press Fetch Schedules below.
```
gannonprudhomme, line 74:
```suggestion
          are unavailable, then press Generate Schedules below.
```

firejake308:
I ammended the requested change into the previous commit, hence the force push

PR #198, opened 2020-04-11
Added scrape all to scrape_depts and scrape_courses

### LINKED ISSUE #138: Change scraping to use bulk_create

### LINKED ISSUE #143: Add scrape all argument to scrape_depts and scrape_courses
We need an easy way to run scraping for both `scrape_depts` and `scrape_courses` on all of the terms possible. This is mainly so we can fill the production database in one command, rather than running scraping on all of the terms individually.

Ideally this would also be threaded so that they're running concurrently. In order to do this we'll need to add a semaphore that's passed into `BannerRequests` so we can limit the amount of requests going out at a single time (the limit for `aiohttp` is 100, but may be lower for Windows at ~64)

We'll also probably need to limit how fast the requests are made, as testing locally had Banner force-denying my requests after ~3 terms ran for `scrape_courses` concurrently.

This would be executed like:

`python manage.py scrape_depts --all` or `python manage.py scrape_courses --all`

gannonprudhomme:
Adds an `all` option for `scrape_depts` and `scrape_courses`, in the same position as the term (i.e. `python manage.py scrape_depts all`). 

`scrape_courses all` takes about ~260 seconds to run, with ~55 seconds of that being saving the models, and the other 205 seconds downloading & parsing the data. `scrape_depts all` takes ~8 seconds.

I added a semaphore to banner_requests as there's a limit to how many sockets you can have opened at once, and experimented with the value and found that 50 resulted in the best run times much to my surprise, but lmk what works best for you.

Also added a retry loop to `perform_search` so that if there's a network error it'll retry it, which it always seem to succeed to after only one retry. There's also this weird edge case where it'll return from the request rather than throwing one of the two network errors, in which cause the returned `courses_list` will be `None`. 

We should also consider making the default of `scrape_depts` and `scrape_courses` be `all`, then if you want to specify a term you can use `-t`/`--term`, like how we do it for `scrape_grades`.

Also adds bulk_create usage for the commands (#138)

Closes #138 , closes #143


autoscheduler/scraper/management/commands/scrape_depts.py
```diff
@@ -1,28 +1,36 @@
...
+from typing import List
 
 from django.core.management import base
 from scraper.banner_requests import BannerRequests
 from scraper.models import Department
+from scraper.management.commands.utils.scraper_utils import get_all_terms
 
-def parse_departments(json, term):
+def parse_departments(json, term) -> List[Department]:
     """ Takes json list of departments and saves them as Department objects """
```
ryan-conn, line 12: Update since the functionality of this changed

autoscheduler/scraper/management/commands/scrape_depts.py
```diff
@@ -1,28 +1,36 @@
...
 
+    depts = []
```
ryan-conn, line 14: Can't suggest since there are removed lines but you can replace this function with
```python
    return [Department(id=dept["code"]+term, code=dept["code"],
                       description=dept["description"], term=term)
            for dept in json]
```

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -0,0 +1,19 @@
+from typing import List
+
+def get_all_terms() -> List[str]:
+    """ Generates all of the terms, from 2019 until now """
+
+    # Figure out the best way to do this
+    # Might want to read in the current year?
+    years = [2013 + i for i in range(7)]
+    semesters = [1, 2, 3]
+    locations = [1, 2, 3]
```
ryan-conn, line 10:
```suggestion
    current_year = datetime.now().year
    years = range(2013, current_year + 1)
    semesters = range(1, 4)
    locations = range(1, 4)
```
Can read in the current year like this and change these iterables to `range`s

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+    terms = []
+
+    for year in years:
+        for semester in semesters:
+            for location in locations:
+                terms.append(f"{year}{semester}{location}")
+
+    return terms
```
ryan-conn, line 19:
```suggestion
    return [f"{year}{semester}{location}"
             for year, semester, location in product(years, semesters, locations)]
```

autoscheduler/scraper/tests/banner_requests_tests.py
```diff
@@ -90,16 +90,32 @@ class BannerRequestsTests(AioTestCase):
 
         # Arrange
         term = "201931" # Fall 2019
-        request = BannerRequests(term)
+        request = BannerRequests()
 
-        depts = ["CSCE", "MATH", "ECEN"]
+        depts = ["CSCE", "ECEN", "MATH"]
+        depts_terms = zip(depts, [term for i in range(len(depts))])
```
ryan-conn, line 96: Use `repeat` or if you want this to stay simple just replace `i` with `_`

autoscheduler/scraper/tests/banner_requests_tests.py
```diff
@@ -90,16 +90,32 @@ class BannerRequestsTests(AioTestCase):
...
+
+        def spy(course_list, _, __, ___):
```
ryan-conn, line 98: You only have to use 1 underscore to ignore a parameter, also since you're using multiple in a row you can use `*_` instead of repeating

autoscheduler/scraper/banner_requests.py
```diff
@@ -149,24 +148,58 @@ def get_departments(self, amount: int = 300) -> List[Dict]:
...
+                            course_list = await self.get_courses(session, session_id,
+                                                                 dept, term, amount)
+
+                        # We only want to limit the requests, not parsing, so call this
+                        # outside of the semaphore
+                        ret = parse_all_courses(course_list, term, courses_set,
+                                                instructors_set)
+
+                        if ret is False:
+                            continue # Failure (course_list was none), retry
```
ryan-conn, line 190: Why would you want to retry here? You seem to be catching all the network errors in the `except` block, and if the request was valid, shouldn't continuing here just return the same results?

autoscheduler/scraper/banner_requests.py
```diff
@@ -149,24 +148,58 @@ def get_departments(self, amount: int = 300) -> List[Dict]:
...
+
+                        result.extend(ret)
 
-                result = await self.get_courses(session, session_id, dept, amount)
+                    except (ClientConnectorError, ContentTypeError):
+                        # Empty lines help it stand out from the rest of the outputs
+                        print(f"\n\nNETWORK ERROR: Retrying {dept} {term}: Take {i} \n\n")
+                        continue # Error occurred, retry
+                    break # If we pass it, no need to keep looping
```
ryan-conn, line 198: Move the `break` up to the end of the `try` block and remove the `continue`, I was confused looking at this at first because I didn't realize the `break` was supposed to execute after the `try` block finishes

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -134,53 +146,125 @@ def parse_course(course_data):
...
+    def handle(self, *args, **options): # pylint: disable=too-many-locals, too-many-statements
+        depts_terms = []
+        start_all = time.time()
+
+        if options['term'] == 'all':
+            terms = get_all_terms()
+
+            for term in terms:
+                depts = get_department_names(term)
+                zipped = zip(depts, [term for i in range(len(depts))])
```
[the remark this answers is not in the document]
gannonprudhomme: I knew there was some way to do this cleaner, nice

autoscheduler/scraper/management/commands/utils/scraper_utils.py
gannonprudhomme: ooooh see I didn't think want to do list comprehension cause I generally don't like having multiple loops but freaking frick this is clean, nice

autoscheduler/scraper/banner_requests.py
gannonprudhomme: Yeah this makes much more sense, will fix

autoscheduler/scraper/banner_requests.py
gannonprudhomme: It's a really weird edge case that only occasionally happens, but the request will return without throwing an error but it returns `None` when it shouldn't, in which case we need to retry. That being said we should probably handle this case in `perform_search` rather than `parse_all_courses`.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -111,11 +116,18 @@ def parse_instructor(course_data):
         email = faculty_data['emailAddress']
 
         instructor_model = Instructor(id=name, email_address=email)
-        instructor_model.save()
 
-        return instructor_model
+        # Causes an error for some reason?
```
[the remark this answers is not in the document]
gannonprudhomme: Nah it's not but honestly I have no idea so I'll just remove it

autoscheduler/scraper/banner_requests.py
gannonprudhomme: To be honest, we should just change this line to `if course_list is None: continue`, then remove the `Union[bool, list]` or whatever it is from `parse_all_courses`, as there's no reason we shouldn't be checking for this case in `perform_search`.

gannonprudhomme:
> We should also consider making the default of scrape_depts and scrape_courses be all, then if you want to specify a term you can use -t/--term, like how we do it for scrape_grades.

Do you think this is something we should do?

autoscheduler/scraper/tests/banner_requests_tests.py
```diff
@@ -90,16 +91,32 @@ async def test_search_does_retrieve_all(self):
...
-        result = await request.search(depts, 1)
+        await request.search(depts_terms, asyncio.Semaphore(3), spy, 1)
+        result = spy.result
 
         # Get all of the according subjects for the retrieved courses
         depts_result = [result[i][0]['subject'] for i in range(0, len(result))]
 
+        # Sort them so it doesn't matter what order they're retrieved in
+        depts_result.sort()
+        depts.sort()
```
ryan-conn, line 118: Change these 2 objects to sets instead of sorting them

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -0,0 +1,14 @@
+from datetime import datetime
+from typing import List
+from itertools import product
+
+def get_all_terms() -> List[str]:
+    """ Generates all of the terms, from 2019 until now """
```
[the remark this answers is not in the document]
gannonprudhomme: The reason we need to actually get all of them is since `scrape_grades` needs a Section model to attach the Grade models to, otherwise the `scrape_grades` will fail for that given term. 

However, we could make an option for the `all` to make it only scrape a given year / location / semester. I.e. `scrape_courses -c 1` would scrape all of the courses for all years & semesters only for college station, or `scrape_courses -y 2019` would scrape all of the semesters/locations for 2019. (Basically like what we have for `scrape_grades`)

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -134,53 +146,125 @@ def parse_course(course_data):
...
+    def handle(self, *args, **options): # pylint: disable=too-many-locals, too-many-statements
+        depts_terms = []
+        start_all = time.time()
+
+        if options['term'] == 'all':
+            terms = get_all_terms()
+
+            for term in terms:
+                depts = get_department_names(term)
+                zipped = zip(depts, repeat(term, times=len(depts)))
```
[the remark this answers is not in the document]
gannonprudhomme: > Inside the `get_department_names` function, we could just return a generator of `(dept, term)` tuples and use comprehensions here and in the branch for when a term code is provided

tbh imma need an example for this cause I'm not positive what you mean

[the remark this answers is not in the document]
gannonprudhomme: > Also idk why I didn't think of this before but you don't actually have to use repeat:

Is there a difference here between doing `[(dept, term) for dept in get_department_names(term)]` vs how you're doing it? Just curious

gannonprudhomme:
Oh shit did not mean to push these GCP commits, gonna force push in a bit and undo them

gannonprudhomme:
Need #207 to be merged so I can add the migrations to the commit that increases the `Section.building` length

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -134,53 +146,120 @@ def parse_course(course_data):
...
 
-    # Parse the instructor, then send the returned Instructor model to parse_section
-    instructor_model = parse_instructor(course_data)
-    parse_section(course_data, instructor_model)
 
-def get_department_names(term_code: str) -> List[str]:
+def get_department_names(term: str) -> List[str]:
     """ Queries database for list of all departments """
-    return [dept.code for dept in Department.objects.filter(term=term_code)]
+    return ((dept.code, term) for dept in Department.objects.filter(term=term))
```
ryan-conn, line 165: Can't make a suggestion because there's a removed line, but we can only do 1 query by doing this and passing in terms as a set:
```python
def get_department_names(terms: List[str]) -> List[str]:
    """ Queries database for list of all departments """
    depts = (Department.objects.filter(term__in=terms).order_by('term')
                               .values('code', 'term'))
    grouped = groupby(depts, key=lambda x: x['term'])
    return ((dept['code'], term) for term, group in grouped for dept in group)
```
I ran the querying part of the code 20 times for a single term and all terms, here's how the times compare:
|           |  old   | new    |
|-----------|--------|--------|
| one term  | 0.0320 | 0.0302 |
| all terms | 0.8955 | 0.4021 |

gannonprudhomme: Yeah great point, will add

gannonprudhomme:
Since your last suggestion should be the last one, I'm gonna add that change, then squash all of the commits that need to be and rebase to be up to date with master

autoscheduler/scraper/management/commands/scrape_courses.py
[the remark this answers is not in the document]
ryan-conn: Yeah, the way I did it is a generator comprehension which has less memory usage and is slightly faster. Internally, a list comprehension translates to `list(generator_comprehension)`, so this avoids evaluating all the items at once and translating them into a list.

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -134,53 +146,123 @@ def parse_course(course_data):
...
+    dept_name = ''
+    if len(course_list) > 0:
+        dept_name = course_list[0]['subject'] if 'subject' in course_list[0] else ''
+
+    for course in course_list:
+        ret.append(parse_course(course, courses_set, instructors_set))
+
+    print(f'{dept_name} {term}: Scraped {len(course_list)} sections')
+
+    return ret
```
ryan-conn, line 187:
```suggestion
    dept_name = course_list[0].get('subject', '') if course_list else ''

    print(f'{dept_name} {term}: Scraped {len(course_list)} sections')

    return (parse_course(course, courses_set, instructors_set) for course in course_list)
```

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -134,53 +146,123 @@ def parse_course(course_data):
...
-        course_model.save()
+        courses_set.add(course_id)
+        return (course_model, instructor_model, section_data)
 
-    COURSES_SET.add(course_id)
+    return (None, instructor_model, section_data)
 
-    # Parse the instructor, then send the returned Instructor model to parse_section
-    instructor_model = parse_instructor(course_data)
-    parse_section(course_data, instructor_model)
```
ryan-conn, line 162: Remove extra newline

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -66,13 +67,15 @@ def parse_section(course_data, instructor: Instructor): # pylint: disable=too-ma
...
-    section_model.save()
 
     # Parse each meeting in this section. i is the counter used to identify each Meeting
+    meetings = [] # Ryan will probably want this to be a tuple
     for i, meetings_data in enumerate(course_data['meetingsFaculty']):
-        parse_meeting(meetings_data, section_model, i)
+        meetings.append(parse_meeting(meetings_data, section_model, i))
 
-def parse_meeting(meetings_data, section: Section, meeting_count: int):
+    return (section_model, meetings)
```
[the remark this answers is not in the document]
gannonprudhomme: Wow that was a really obvious one idk how I didn't notice it either when I was writing it. Fixed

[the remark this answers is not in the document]
gannonprudhomme: Lmao I even wrote "Ryan will probably want this to be a tuple" (meaning a generator, I mix them up since they both use parentheses)

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -111,11 +117,17 @@ def parse_instructor(course_data):
...
         instructor_model = Instructor(id=name, email_address=email)
-        instructor_model.save()
 
-        return instructor_model
+        if name not in instructors_set:
+            instructors_set.add(name)
+            return instructor_model
 
-def parse_course(course_data):
+        return None
```
ryan-conn, line 125: Delete this, every python function implicitly returns None when it reaches the end

autoscheduler/scraper/banner_requests.py
```diff
@@ -149,24 +148,52 @@ def get_departments(self, amount: int = 300) -> List[Dict]:
...
 
-        async def perform_search(dept: str):
-            result = None
+        courses_set = set()
+        instructors_set = set()
+
+        async def perform_search(dept: str, term: str):
             async with ClientSession(loop=loop) as session:
-                session_id = await self.create_session(session)
+                course_list = []
```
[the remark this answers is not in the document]
gannonprudhomme: While idk why I had it there, tbh I didn't realize that variables can be used outside of context managers / I assumed context managers had its own scope. Just gonna squash this into a previous commit

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Fixed, squashed into "updated scrape_courses to use bulk_create"

gannonprudhomme:
Sorry for all the force pushes lol, but I:
- Squashed "Updated scrape_courses to use bulk_create`
- Squashed the `course_list` removal into "Added retrying & semaphore usage to banner_requests"
- Rebased to be up-to-date with master, and added the migration for the Meeting.building length change

autoscheduler/scraper/banner_requests.py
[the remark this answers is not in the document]
ryan-conn: Just as a note: the only things that actually create new scopes in python are modules, classes, and functions, not even conditional branches make their own scope

Issue #186, opened 2020-04-09
Add displaying of empty/filled seats to course cards


gannonprudhomme:
We should add the ability for displaying how many empty seats there are for a given section to the course card.

(insert design here)

PR #199, opened 2020-04-11
Course Select Card Improvements

### LINKED ISSUE #172: Course Select Card improvements

firejake308:
* Fixed font of header buttons
* Clicking anywhere on the header now expands/collapses
* changed remove icon to trash can for consistency

### DIFF
--- a/autoscheduler/frontend/src/components/App/App.tsx
+++ b/autoscheduler/frontend/src/components/App/App.tsx
@@ -10,6 +10,7 @@ import fetchSavedSchedule from './testMeetings';
 import { replaceMeetings } from '../../redux/actions/meetings';
 import NavBar from '../NavBar';
 import LandingPage from '../LandingPage/LandingPage';
+import Test from './Test';
 
 const App: React.SFC = function App() {
   // connect to Redux store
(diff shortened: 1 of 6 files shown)


firejake308:
Both Ryan and Edgar pointed out that they expected clicking anywhere on the header to expand the course cards, so I went ahead and implemented that. I assume that collapsing should work the same, but I definitely want y'all to check how that feels before we finalize it

gannonprudhomme:
Honestly, I say we do #186 in a different PR, as I like how we've been keeping a few of these PR's really small, as they're really easy to review & merge in a short amount of time (which we have struggled at in the past)

gannonprudhomme:
But this works fine on my end, functions exactly how I'd expect it. My only comment would be we might want the ability to remove the course cards when they're collapsed, but I'm not sure what we would do with the text so I think it's fine if we add it at a later date

firejake308:
I've added the remove icon in the collpased card. Does the positioning make sense?

gannonprudhomme:
Oh lmao idk why I didn't think about that, that makes total sense 👍

PR #201, opened 2020-04-11
Fix redundant course search

### LINKED ISSUE #172: Course Select Card improvements

firejake308:
The fix required updating Material UI, so you'll have to rerun `npm install`. Also, be on the lookout for any breaking changes. One minor bug I've noticed is that the autocomplete menu is sometimes misplaced for some reason, but it seems to be fixed when there are 2 course cards.

Closes #172


gannonprudhomme:
I think you might've rebased something wrong, as you have a few duplicate commits in this that are different from the original `frontend/course-card-route-connection` branch

firejake308:
I missed the force-push to fix the typo in `Sections.ts`. I'll fix it and then force push my own branch

firejake308:
Replaced by #202

PR #202, opened 2020-04-12
Fix redundant course search take 2

### LINKED ISSUE #172: Course Select Card improvements

firejake308:
I just made a new branch to fix the incorrect rebase from #201 

The fix required updating Material UI, so you'll have to rerun `npm install`. Also, be on the lookout for any breaking changes. One minor bug I've noticed is that the autocomplete menu is sometimes misplaced for some reason, but it seems to be fixed when there are 2 course cards.

TODO automatically select if the user types the exact course

Closes #172


gannonprudhomme:
You mean to merged this into `frontend/course-card-route-connection`, right?

firejake308:
Yeah, just realized that. Also rebased again to get it to have just the one commit that I added

firejake308:
Btw, if you want to get the linting to pass, I'm pretty sure reading onto the current master should take care of it

firejake308:
And for the tests, you'll have to mock fetch in order to get that passing

PR #207, opened 2020-04-15
Began tracking migrations

gannonprudhomme:
Removes migrations from the `.gitignore` and adds the initial migration + the building modification one. This is mainly just getting prepared for deploying to GCP, as migrations will needed to be tracked whenever we make schema changes so we can correctly apply them to Google Cloud SQL

Was going to do this in #198, but since it's not really needed for it I figured we might as well do it on a separate branch.


gannonprudhomme:
Now that I think about it, `master` doesn't currently have Meeting.building schema alteration, so I'm going to rebase this and remove that migration. Then once this is merged in `master`, I'll add that migration in #198 to the commit that actually modifies the `Meeting` model.

.gitignore
```diff
@@ -3,8 +3,6 @@
 *.code-workspace
```
ryan-conn, line 3: Looking at the gitignore, in line 1 there's extra space after `VSCode` that's going to bother me unless it gets removed, can you just fix it in this PR

PR #208, opened 2020-04-15
Implemented scheduling api route

### LINKED ISSUE #165: Add route for scheduling algorithm
Will fill out more once #149 makes more progress

gannonprudhomme:
Implements the scheduling api at /scheduler/generate.

We can't include the `api` in the route since it redirects to scraper.urls, but we can modify it if we want to.

Closes #165

### DIFF
--- a/autoscheduler/scheduler/tests/api_tests.py
+++ b/autoscheduler/scheduler/tests/api_tests.py
@@ -1,7 +1,7 @@
 from unittest.mock import patch
 from datetime import time
 from rest_framework.test import APITestCase, APIClient
-from scheduler.views import (_parse_course_filter, _parse_unavailable_times,
+from scheduler.views import (_parse_course_filter, _parse_unavailable_time,
                              _serialize_schedules)
 from scheduler.utils import UnavailableTime, CourseFilter
 from scraper.models import Section, Instructor
@@ -30,45 +30,37 @@ class SchedulingAPITests(APITestCase):
         """ Tests that _parse_counter_filter works on a typical input """
 
         # Arrange
-        courses = [
-            {
-                "subject": "CSCE",
-                "courseNum": "121",
-                "sections": ["500"],
-                "honors": False,
-                "web": False,
-            },
-        ]
+        course = {
+            "subject": "CSCE",
+            "courseNum": "121",
+            "sections": ["500"],
+            "honors": False,
+            "web": False,
+        }
 
-        expected = [
-            CourseFilter(subject="CSCE", course_num="121", section_nums=["500"],
-                         honors=False, web=False)
-        ]
+        expected = CourseFilter(subject="CSCE", course_num="121", section_nums=["500"],
+                                honors=False, web=False)
 
         # Act
-        result = _parse_course_filter(courses)
+        result = _parse_course_filter(course)
 
         # Assert
         self.assertEqual(result, expected)
 
-    def test_parse_unavailable_times_is_correct(self):
+    def test_parse_unavailable_time_is_correct(self):
         """ Tests that _parse_unavailable_times works on a typical input """
 
         # Arrange
-        availabilities = [
-            {
-                "startTime": "0800",
-                "endTime": "1000",
-                "day": 0,
-            },
-        ]
+        availability = {
+            "startTime": "0800",
+            "endTime": "1000",
+            "day": 0,
+        }
 
-        expected = [
-            UnavailableTime(time(8, 00), time(10, 00), 0)
-        ]
+        expected = UnavailableTime(time(8, 00), time(10, 00), 0)
 
         # Act
-        result = _parse_unavailable_times(availabilities)
+        result = _parse_unavailable_time(availability)
 
         # Assert
         self.assertEqual(result, expected)
--- a/autoscheduler/scheduler/tests/api_tests.py
+++ b/autoscheduler/scheduler/tests/api_tests.py
@@ -2,7 +2,7 @@ from unittest.mock import patch
 from datetime import time
 from rest_framework.test import APITestCase, APIClient
 from scheduler.views import (_parse_course_filter, _parse_unavailable_times,
-                             _serialize_schedule)
+                             _serialize_schedules)
 from scheduler.utils import UnavailableTime, CourseFilter
 from scraper.models import Section, Instructor
 from scraper.serializers import SectionSerializer
@@ -73,7 +73,7 @@ class SchedulingAPITests(APITestCase):
         # Assert
         self.assertEqual(result, expected)
 
-    def test_serialize_schedule_is_correct(self):
+    def test_serialize_schedules_is_correct(self):
         """ Tests that _serialize_schedule works on a typical input """
 
         # Arrange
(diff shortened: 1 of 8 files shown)


autoscheduler/scheduler/tests/api_tests.py
```diff
@@ -0,0 +1,132 @@
...
+
+    def test_serialize_schedule_is_correct(self):
+        """ Tests that _serialize_schedule works on a typical input """
+
+        # Arrange
+        schedule = [("1", "2")]
+
+        expected = [
+            [SectionSerializer(section).data for section in self.sections]
+        ]
```
ryan-conn, line 84: This test seems useless because this is exactly the same way `_serialize_schedule` gets schedules, so explicitly declare the expected data here

autoscheduler/scheduler/views.py
```diff
@@ -1,4 +1,95 @@
...
+def _parse_unavailable_times(availabilities) -> List[UnavailableTime]:
+    """ Parses the availabilities input and converts them to UnavailableTime objects
+        to be used in create_schedules
+    """
+
+    unavailable_times = []
+    for avail in availabilities:
+        start_time = convert_meeting_time(avail["startTime"])
+        end_time = convert_meeting_time(avail["endTime"])
+        day = int(avail["day"])
```
ryan-conn, line 41: Why shouldn't we just send the day in unavailable times as an int?

autoscheduler/scheduler/views.py
```diff
@@ -1,4 +1,95 @@
...
+    sections_dict = {section.id: section for section in models.iterator()}
+
+    output = []
+    for schedule in schedules:
+        schedule_out = [SectionSerializer(sections_dict[int(section_id)]).data
+                        for _, section_id in enumerate(schedule)]
+
+        output.append(schedule_out)
+
+    return output
```
ryan-conn, line 73:
```suggestion
    def sections_for_schedule(schedule):
        sections = (sections_dict[section] for section in schedule)
        return SectionSerializer(sections, many=True).data

    return [sections_for_schedule(schedule) for schedule in schedules]
```

autoscheduler/scheduler/tests/api_tests.py
```diff
@@ -0,0 +1,132 @@
...
+        self.assertEqual(result, expected)
+
+    # Replaces the create_schedules import that's imported in scheduler.views
+    @patch('scheduler.views.create_schedules')
+    def test_route_scheduling_generate_is_correct(self, create_schedules_mock):
+        """ Tests that /scheduling/generate evalutes correctly """
+
+        # Arrange
+        # Mock create schedules so we don't have to make the meetings for the sections
+        create_schedules_mock.return_value = [("1", "2")]
```
ryan-conn, line 99: Should be a tuple of ints

autoscheduler/scheduler/views.py
```diff
@@ -1,4 +1,95 @@
...
+    Args:
+        schedules: The schedules returned from create_schedules
+
+    Returns
+        The list of given schedules with the corresponding serialized sections
+    """
+
+    # Retrieve the section models in bulk so we only do one DB query
+    # Put the section ids in a set to remove duplicates
+    section_set = set((section_id for schedule in schedules for section_id in schedule))
```
ryan-conn, line 60:
```suggestion
    section_set = set(section_id for schedule in schedules for section_id in schedule)
```

autoscheduler/scheduler/views.py
```diff
@@ -1,4 +1,95 @@
...
+
+        # List[Tuple[str, str]]
+        courses = _parse_course_filter(query["courses"])
+        unavailable_times = _parse_unavailable_times(query["availabilities"])
+
+        term = query["term"]
+        include_full = query["includeFull"]
+
+        # Might want to make num_schedules to generate a variable
+        schedules = create_schedules(courses, term, unavailable_times, include_full, 2)
```
ryan-conn, line 93: What does that comment mean, getting `num_schedules` from the request? Since that would be more flexible (and generating more of them takes hardly any more time, even if they want to generate 1000+ schedules) I agree

autoscheduler/scheduler/tests/api_tests.py
```diff
@@ -0,0 +1,132 @@
...
+        result = _parse_unavailable_times(availabilities)
+
+        # Assert
+        self.assertEqual(result, expected)
+
+    def test_serialize_schedule_is_correct(self):
+        """ Tests that _serialize_schedule works on a typical input """
+
+        # Arrange
+        schedule = [("1", "2")]
```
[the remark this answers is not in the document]
gannonprudhomme: Oooohhh I thought it returned them as strings cause that was the return type on `create_schedules`, will fix

autoscheduler/scheduler/tests/api_tests.py
gannonprudhomme: Yeah that's true, I just really didn't want to write it all out manually lol

autoscheduler/scheduler/views.py
gannonprudhomme: lol I actually meant make `num_schedules` a variable rather than have it just be `5` in the parameter, but we can do that too

autoscheduler/scheduler/views.py
gannonprudhomme: That is a great point, I'll change it with that assumption

autoscheduler/scheduler/tests/api_tests.py
gannonprudhomme: Actually I think I disagree. We shouldn't be testing the SectionSerializer's functionality (which we would be doing if we wrote it all the way out). What this is really testing is that it gets the corresponding sections id's, fetches the models, and serializes the _correct_ sections, not that it serializes the sections correctly. Hopefully this way if the `SectionSerializer` is broken, this will still pass, since the output will be the same for both (assuming it's still getting the same sections).

gannonprudhomme: That being said, for the `/schedules/generate` route, I should probably write out completely since I think the output in that case is relevant to the test (probably)

autoscheduler/scheduler/views.py
```diff
@@ -1,4 +1,89 @@
...
+        to be used in create_schedules
+    """
+
+    start_time = convert_meeting_time(avail["startTime"])
+    end_time = convert_meeting_time(avail["endTime"])
+    day = avail["day"]
+
+    return UnavailableTime(start_time, end_time, day)
+
+
```
ryan-conn, line 40: Remove extra newline

autoscheduler/scheduler/views.py
```diff
@@ -1,4 +1,89 @@
...
+        sections = (sections_dict[section] for section in schedule)
+
+        return SectionSerializer(sections, many=True).data
+
+    return [sections_for_schedule(schedule) for schedule in schedules]
+
+class ScheduleView(APIView):
+    """ Handles requests to the generate schedules algorithm  """
+
+    def post(self, request): # pylint: disable=too-many-locals
```
ryan-conn, line 70:
```suggestion
    def post(self, request):
```
There are only 7 locals so this linter rule isn't actually being broken

gannonprudhomme:
Squashed the minor fixes and rebased to be up to date with master

PR #210, opened 2020-04-17
Implement course card connection to api/sections & api/course/search

### LINKED ISSUE #144: Connect course select card to /api/sections and /api/course/search
Not sure if we'll want this to be separate issues, but once Backend V0.2 is completed, we'll want to connect the frontend to the completed routes on the backend so they're actually functioning like they should.

### LINKED ISSUE #172: Course Select Card improvements

gannonprudhomme:
This also adds the `subject`, `course_num`, `min_credits`, `max_credits`, `current_enrollment` and `max_enrollment` fields to `SectionSerializer`. The `subject` and `course_num` are just so we don't have to use the data from the search to fill in the `Section` types, but they will be needed later for the scheduling algorithm since we're not caching the Sections.

I made this a draft as I first want to check what other tests I can make for this, as I really only made one for #202 + the parseMeeting test, but I also need help from Adel on figuring out how to actually assert correctly for the #202 test. Once these are all figured out I'll open this up and also request Ryan for review

Closes #144 
Closes #172 through #202


autoscheduler/frontend/src/tests/CourseSelectCard.test.tsx
```diff
@@ -9,98 +14,173 @@ import { Provider } from 'react-redux';
...
+        act(() => { fireEvent.click(getByText('Section')); }); // Makes api/sections be called
+
+        // collapse then expand the card
+        act(() => { fireEvent.click(getByText('Collapse')); });
+        act(() => { fireEvent.click(getByLabelText('Expand')); });
+
+        // assert
+        // Need to assert that api/sections was only called once somehow?
+        //
+        // expect(fetchMock).toBeCalledTimes(1);
```
firejake308, line 182: We have a couple of options with how we can do this. The simplest is to assert `expect(fetchMock.mock.calls).toHaveLength(2)` so that it's called once for search and once for sections. 

Alternatively, if we swap out the `jest-fetch-mock` library for the `fetch-mock-jest` library, we can assert more idiomatically `expect(fetchMock).toHaveFetchedTimes(2)`. Beware that with either of these, you will have to call `fetchMock.mockClear()` at the top of the arrange phase for this test to erase data from other tests.

In addition to the choice between the two libraries, we can also choose to assert only on the calls made to `/api/sections` (sort of). To do this, we have to make a custom implementation like so:

```
fetchMock.mockImplementation((route: string) => {
          if (route.match(/api\/course.+/)) {
            return Promise.resolve(new Response(JSON.stringify({
              results: ['CSCE 121'],
            })));
          }
          if (route.match(/api\/sections.+/)) {
            sectionCalls += 1;
            return Promise.resolve(new Response('[]'));
          }
          return Promise.resolve(new Response('404 Not Found'));
        });
```
This works by making a custom variable called `sectionCalls` that is incremented whenever the URL matches api/section/whatever. We can then assert on `sectionCalls`.

gannonprudhomme: The fact that there's a `jest-fetch-mock` and `fetch-mock-jest` makes me laugh

gannonprudhomme: I'd say I like the third one the best so that's what I'm gonna go with

gannonprudhomme:
Squashed a linting fix into `Added enrollment sending from api/sections`(0c2aed8), hence the force push

gannonprudhomme:
Ok now it's fixed lol. This is why you test your commits before squashing them

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -35,6 +36,57 @@ function updateCourseCardSync(index: number, courseCard: CourseCardOptions): Upd
...
+      if (meetingData.start_time != null && meetingData.start_time.length > 0) {
+        start = meetingData.start_time.split(':');
+        end = meetingData.end_time.split(':');
+      }
+
+      const meeting = new Meeting({
+        id: Number(meetingData.id),
+        building: '',
+        meetingDays: meetingData.days,
+        startTimeHours: +start[0],
```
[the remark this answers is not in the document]
gannonprudhomme: this is due to my lack of js/ts knowledge lol, will fix

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -1,5 +1,5 @@
 .container {
-    width: calc(100% - 8px);
+    width: calc(30% - 8px);
```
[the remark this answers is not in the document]
gannonprudhomme: Yeah I made that in 4c4ee4d, which I'm going to drop before I push this, hence the `(temp)`

autoscheduler/frontend/src/tests/CourseSelectCard.test.tsx
```diff
@@ -9,98 +14,183 @@ import { Provider } from 'react-redux';
...
+            sectionsFetchCount += 1;
+            return testFetch(route);
+          }
+
+          return Promise.resolve(new Response('404 Not Found'));
+        });
+
+        const nodeProps = Object.create(Node.prototype, {});
+        // @ts-ignore
+        document.createRange = (): Range => ({
```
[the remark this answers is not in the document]
gannonprudhomme: Yeah I was thinking the same thing, especially since its not like truly integral the test / just seems to set up the page. What exactly does it do anyways?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -60,9 +119,10 @@ function updateCourseCardAsync(
         selected: false,
         meetings: [curr],
       }];
-    }, [])).then(
-      (sections) => dispatch(updateCourseCardSync(index, { ...courseCard, sections })),
-    );
+    }, []))
```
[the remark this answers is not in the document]
gannonprudhomme: I thought so too but it keeps giving me linting errors, could you make a suggestion for it?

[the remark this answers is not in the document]
firejake308: I had to change several lines to make all of them uniform, so I just kinda pushed a commit

autoscheduler/frontend/src/tests/CourseSelectCard.test.tsx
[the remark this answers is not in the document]
firejake308: We have to mock the `createRange` function because the Autocomplete calls it when it makes the popover menu, but the jest "document" doesn't have this function by default

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
```diff
@@ -28,6 +28,19 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
...
+  const [options, setOptions] = React.useState([]);
+  const [inputValue, setInputValue] = React.useState('');
+
+  function getAutocomplete(text: string): void {
+    fetch(`api/course/search?search=${text}&term=201931`).then(
+      (res) => res.json(),
+    ).then((result) => {
+      if (result.results) {
+        setOptions(result.results);
+      }
```
ryan-conn, line 40: What exactly is the point of this check?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -60,9 +108,38 @@ function updateCourseCardAsync(
...
+   * @param index the index of the course card to update in the CourseCardArray
+   * @param courseCard the options to update
+   */
+function updateCourseCardAsync(
+  index: number, courseCard: CourseCardOptions,
+): ThunkAction<void, RootState, undefined, UpdateCourseAction> {
+  return (dispatch): void => {
+    const split = courseCard.course.split(' ');
+    const subject = split[0];
+    const courseNum = split[1];
```
ryan-conn, line 127:
```suggestion
    const [subject, courseNum] = courseCard.course.split(' ');
```

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
gannonprudhomme: Uh typical null check? I think it was needed at one point but idk if it still is or not lol

gannonprudhomme:
> I noticed that it makes an extra API call after the input is cleared. Can you add in a check to prevent that (along with an appropriate test)? Incidentally, this is why it shows WGST sections when the course is empty

Will check it out. I've been trying to look into it, but could you look into the overall performance of it / check that it's not rendering/performing operations when it shouldn't? I was getting performance issues last night from it, in which searching for a course would make chrome take up like 10% CPU usage, then once I stopped it would go back down to normal.

firejake308:
I haven't tested very rigorously yet, but after cursory profiling, it seems like rendering the Material-UI Autocomplete element is one of the more expensive operations that we're doing. Virtualizing the list  could help with this if there are a lot of options, or if you want to be lazy, you could limit the autocomplete to showing only the first couple (~10?) suggestions

gannonprudhomme:
> I haven't tested very rigorously yet, but after cursory profiling, it seems like rendering the Material-UI Autocomplete element is one of the more expensive operations that we're doing. Virtualizing the list could help with this if there are a lot of options, or if you want to be lazy, you could limit the autocomplete to showing only the first couple (~10?) suggestions

I'll look into virtualizing cause I think that makes the most sense. iOS does similar stuff with their lists (only makes a certain amount of cells and just reuses the ones that are off screen) and it seems to work well for it

gannonprudhomme:
So I implemented virtualization using `react-window`, and it showed no noticeable improvement to the responsive / slowness of it. 

Turns out it was mainly due to that the autocomplete would send out an initial empty request, and `api/course/search` returned every single course (43 KB) since there was the `starts_with` filter was empty, and chrome generally took ~250ms for this to complete. Making api/course/search only send 25 objects seems to have removed this issue, as all of the other requests are generally less than 1 KB now.

autoscheduler/frontend/src/components/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
```diff
@@ -79,15 +94,27 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
...
+          }}
+          onChange={(evt: object, val: string): void => {
             dispatch(updateCourseCard(id, {
               course: val,
-            }));
+            }, term));
+          }}
+          onInputChange={(evt: object, val: string, reason: string): void => {
+            setInputValue(val);
+            if (val === '') return; // Skip empty inputs
```
firejake308, line 114: Can you `setOptions([])` before returning here? Because right now, it leaves the suggestions list after you clear the input, which feels a little off. Other than that, this PR looks good to me.

gannonprudhomme:
So I:
- Added the `setOptions([])` setting in ExpandedCourseCard and squashed it into `Prevented empty input fetch to api/course/search`.
- Rebased with master and resolved conflicts
- Dropped the (temp) commits

PR #211, opened 2020-04-17
Frontend title fixes

gannonprudhomme:
- Changes the NavBar title and window title to "Rev Registration"
- Makes the NavBar title clickable and redirect to the landing page / root upon click

Probably should've made the branch `frontend/title-fixes`, but w/e

### DIFF
--- a/autoscheduler/frontend/src/components/NavBar.tsx
+++ b/autoscheduler/frontend/src/components/NavBar.tsx
@@ -15,6 +15,9 @@ const useStyles = makeStyles((theme) => ({
   title: {
     flexGrow: 1,
   },
+  buttonText: {
+    textTransform: 'capitalize',
+  },
 }));
 
 const NavBar: React.SFC = () => {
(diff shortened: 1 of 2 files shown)


autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -26,12 +27,20 @@ const NavBar: React.SFC = () => {
...
       >
         <Toolbar>
-          <Typography
-            variant="h6"
-            className={classes.title}
-          >
-            Auto Scheduler
-          </Typography>
+          <div className={classes.title}>
+            <Button
```
firejake308, line 31:
```suggestion
            <Button
              classes={{ label: classes.buttonText }}
```

I'm not a fan of the all-uppercase text. If you add a field called `buttonText` to the `makeStyles()` above, you can tell the button to apply that class to the label of the button. In that field, set `textTransform` to 'capitalize', since the default is 'uppercase'

gannonprudhomme: Oh I didn't even notice it was all caps lmao, I like this much better

PR #217, opened 2020-04-21
Django settings fix for Cloud SQL proxy

gannonprudhomme:
This mainly adds the `SETTINGS_MODE=proxy` option, as the `STATIC_URL` has to be pointing to the actual static files rather than cloud storage. Likewise the only thing that `prod` is actually used for is collecting static files, so I changed that settings mode to `static` instead.

Also `cloud_sql_proxy` to the `.gitignore`

### DIFF
--- a/autoscheduler/autoscheduler/settings/base.py
+++ b/autoscheduler/autoscheduler/settings/base.py
@@ -3,8 +3,10 @@ from autoscheduler.config import config
 
 # What Google App Engine uses
 _IS_GCP = os.getenv('SERVER_SOFTWARE', '').startswith('gunicorn')
-# Environment variables for when we connect to Google Cloud SQL/collect static files
+# Environment variables for when we collect static files
 _IS_PROD = os.getenv('SETTINGS_MODE') == 'prod'
+# Env variable for connecting to Cloud SQL through cloud_sql_proxy
+_IS_PROXY = os.getenv('SETTINGS_MODE') == 'proxy'
 
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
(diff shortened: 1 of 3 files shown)


gannonprudhomme:
Probably gonna be force pushing this quite a few times until I can get the Django secret loading, will re-request review when its working

gannonprudhomme:
Should be good to go now. [Here's](https://github.com/aggie-coding-club/Rev-Registration/runs/607202947) the check on the previous run before I dropped the `on: push` change to make deploy-workflow.yml run on all pushes so I could test this.

autoscheduler/autoscheduler/settings/base.py
```diff
@@ -13,7 +15,15 @@
...
 
 # SECURITY WARNING: keep the secret key used in production secret!
-SECRET_KEY = 'ogxjva5h%&5c7&e#-2f1&+u5p#zygwffcy!@)k8i37#j_89xe2'
+if _IS_GCP:
+    # For production, load the secret from secret.txt and set it
+    # This is set from the GitHub actions secret DJANGO_SECRET in deploy-workflow.yml
+    _SECRET_PATH = os.path.dirname(__file__) + "/secret.txt"
+    _SECRET_FILE = open(_SECRET_PATH, "r")
+
+    SECRET_KEY = _SECRET_FILE.read()
```
ryan-conn, line 24:
```suggestion
    with open(_SECRET_PATH) as _SECRET_FILE:
        SECRET_KEY = _SECRET_FILE.read()[:-1]
```
A couple things that I changed here:
- We should always use context managers when opening files (would've helped here as you forgot to close the file)
- `"r"` is the default mode with open, so no need to use it as an argument
- `file.read()` reads the whole file, including the newline created by `echo`

gannonprudhomme: Fixed

PR #218, opened 2020-04-21
Misc Actions/CI changes

gannonprudhomme:
- Makes the frontend & backend steps run on all pushes/not ignore `backend/` and `frontend/` branches respectively.
- Removed the check in danger that gave an warning when you touch frontend related files on `backend/*` branches and vice-versa


gannonprudhomme:
> Everything changed in this PR looks fine, but I noticed something looking at `dangerfile.ts`:
> image
> `offset` gets subtracted twice from the number of changed lines, change that and this should be good

I'm also going to make it print how many lines changed there are (with the offset included) as I think it's useful

PR #220, opened 2020-04-22
Made web and honors more flexible

### LINKED ISSUE #176: Make honors/web only selection allow excluding honors/web classes
For the basic course select, right now there are only 2 options: honors only and not honors only. This is a little up to interpretation on what it should do if the boxes aren't checked: to exclude non-honors/web sections or to choose any section.

Because of this, and the fact that I see valid use cases for wanting either of these behaviors, we should make this into a three-way selection that lets the user exclude honors, have no preference, or only use honors sections. Here's a draft of what this could look like:
image
Suggestions are welcome for improvements, but we'll want this functionality in some form.

firejake308:
I have attempted to implement this change on the frontend and backend, but ngl, I have not tested the backend part at all

Closes #176


firejake308:
Replaced by #221 since this affects both frontend and backend and idk how to rename a branch without just creating a new one

Issue #193, opened 2020-04-10
Connect "Fetch Schedules" with /scheduler/generate


gannonprudhomme:
Once #145, #169, #149, and #165 are completed, we need to connect generate schedules with the route for the scheduling algorithm. 

When I implemented this on my fork, since /scheduler/generate on the backend uses (or at least should use) the `SectionSerializer`, we should be able to use the `parseMeetings` function from #144 to implement this pretty easily.

gannonprudhomme:
It'll also have to be a POST request since GET requests can't have a request body, which means we'll need to figure out CSRF authentication or w/e. In order to get it working I had to disable the CSRF middleware. As such, we'll need to pass as CSRF token from the backend to the frontend in order for use to do these POST requests to the backend.

gannonprudhomme:
After reading about it more, we might just be able to disable it for the view made in #165 , since the route doesn't actually modify anything on the server (which is what CSRF is protecting against), as it's really just a GET request in disguise. From [here](https://owasp.org/www-community/attacks/csrf):

`
CSRF attacks target functionality that causes a state change on the server, such as changing the victim’s email address or password, or purchasing something. Forcing the victim to retrieve data doesn’t benefit an attacker because the attacker doesn’t receive the response, the victim does. As such, CSRF attacks target state-changing requests.
`

firejake308:
Makes sense to me. I'm in favor if disabling CSRF for this view then.

gannonprudhomme:
We also need to make it so that if a course card has specific sections chosen, but they're on the "Basic" customization level, those sections shouldn't be sent to the `/scheduler/generate` route. Likewise if they have "web only" selected but they're on "Section", don't include the filter.

Issue #163, opened 2020-03-18
Design/implement grade distribution display for course select cards


gannonprudhomme:
Now that we have grade distributions scraped, we need to design and implement the component that'll be displayed next to professor's names in the course select card, so students can quickly see what the professor's grade distributions were for past semesters.

Should also show what the average GPA was next to it.

gannonprudhomme:
![Figma_tYyk5pfDqq](https://user-images.githubusercontent.com/7295783/80643218-52785280-8a2d-11ea-848c-73180b087b11.png)

Here's what I mocked up real quick, based off of the design Adel made awhile ago.

PR #232, opened 2020-04-28
Added grades to Section type

gannonprudhomme:
I added the grades field to the Section type, as we'll need it for the schedule GPA calculating in #193 (and I didn't want to do it in that PR), as well as for the grade distributions (#163).

I also made the `SectionSerializer` return `None` instead of `Grades` models where all of the fields were null in the cases that a prof has no previous grades out for a course, as it makes it much easier to handle on the frontend.

### DIFF
--- a/autoscheduler/frontend/src/redux/actions/courseCards.ts
+++ b/autoscheduler/frontend/src/redux/actions/courseCards.ts
@@ -8,6 +8,7 @@ import { RootState } from '../reducer';
 import Meeting, { MeetingType } from '../../types/Meeting';
 import Section from '../../types/Section';
 import Instructor from '../../types/Instructor';
+import Grades from '../../types/Grades';
 
 export function addCourseCard(): AddCourseAction {
   return {
@@ -55,6 +56,7 @@ export function parseMeetings(arr: any[]): Meeting[] {
       currentEnrollment: Number(sectionData.current_enrollment),
       maxEnrollment: Number(sectionData.max_enrollment),
       instructor: new Instructor({ name: sectionData.instructor_name }),
+      grades: sectionData.grades == null ? null : new Grades(sectionData.grades),
     });
 
     sectionData.meetings.forEach((meetingData: any) => {
--- a/autoscheduler/frontend/src/types/Grades.ts
+++ b/autoscheduler/frontend/src/types/Grades.ts
@@ -23,6 +23,18 @@ export default class Grades {
   Q: number;
   X: number;
   constructor(init: GradesInit) {
+    // None of the properties can be null
+    if (init.gpa == null) { throw Error(`Grades.gpa is invalid: ${init.gpa}`); }
+    if (init.A == null) { throw Error(`Grades.A is invalid: ${init.A}`); }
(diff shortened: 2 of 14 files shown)


gannonprudhomme:
I'm also probably gonna add null checks to the `Grades` type, as I think that's a good thing to enforce so we can assume the individual grades object are not null if the actual `Grades` object exists.

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -55,6 +56,7 @@ export function parseMeetings(arr: any[]): Meeting[] {
       currentEnrollment: Number(sectionData.current_enrollment),
       maxEnrollment: Number(sectionData.max_enrollment),
       instructor: new Instructor({ name: sectionData.instructor_name }),
+      grades: sectionData.grades == null ? null : new Grades({ ...sectionData.grades }),
```
ryan-conn, line 59: I'm pretty sure you don't have to use spreading here, since the constructor just uses `Object.assign` with the parameter passed in

gannonprudhomme: Squashed into `Added grades to section type`

autoscheduler/frontend/src/types/Grades.ts
```diff
@@ -23,6 +23,18 @@ export default class Grades {
   Q: number;
   X: number;
   constructor(init: GradesInit) {
+    // None of the properties can be null
+    if (init.gpa == null) { throw Error(`Grades.gpa is invalid: ${init.gpa}`); }
```
firejake308, line 27: I'm assuming that you're using double-equals here intentionally, to catch `undefined` as well?

gannonprudhomme: uh lol no, I honestly meant to do a triple-equals. I do think we should probably catch undefined as well, as that definitely shouldn't happen

autoscheduler/frontend/src/types/Grades.ts
```diff
@@ -23,6 +23,18 @@ export default class Grades {
...
+    if (init.A == null) { throw Error(`Grades.gpa is invalid: ${init.A}`); }
+    if (init.B == null) { throw Error(`Grades.gpa is invalid: ${init.B}`); }
+    if (init.C == null) { throw Error(`Grades.gpa is invalid: ${init.C}`); }
+    if (init.D == null) { throw Error(`Grades.gpa is invalid: ${init.D}`); }
+    if (init.F == null) { throw Error(`Grades.gpa is invalid: ${init.F}`); }
+    if (init.I == null) { throw Error(`Grades.gpa is invalid: ${init.I}`); }
+    if (init.S == null) { throw Error(`Grades.gpa is invalid: ${init.S}`); }
+    if (init.Q == null) { throw Error(`Grades.gpa is invalid: ${init.Q}`); }
+    if (init.X == null) { throw Error(`Grades.gpa is invalid: ${init.X}`); }
```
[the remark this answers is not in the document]
gannonprudhomme: Idk how the hell I missed this this is so basic lol, fixing rn

PR #233, opened 2020-04-28
Connected /api/terms to SelectTerm

### LINKED ISSUE #147: Connect /api/terms to landing page

ryan-conn:
Connects the `SelectTerm` component to /api/terms, as well as a couple other changes in order to make the data between our frontend/backend more consistent:
- `term` in the redux store is now a string, as that's what it's stored as on the backend. (If we'd rather change it on the backend to a number instead, I can do that)
- `/api/terms` now returns results in the order `description -> term code`, since descriptions map to their corresponding term codes, not the other way around

I also updated `tsconfig.json` to use es2018, as it makes translating objects into maps easier with the `Object.entries(obj)` function. Since our website already doesn't work with internet explorer, this shouldn't really be a problem, but we should consider using a polyfill library to hopefully make it compatible.

Closes #147.

### DIFF
--- a/autoscheduler/frontend/src/tests/ui/SelectTerm.test.tsx
+++ b/autoscheduler/frontend/src/tests/ui/SelectTerm.test.tsx
@@ -2,9 +2,7 @@ import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
 
 enableFetchMocks();
 /* eslint-disable import/first */ // enableFetchMocks must be called before others are imported
-import {
-  act, render, fireEvent, waitForElement,
-} from '@testing-library/react';
+import { render, fireEvent, waitFor } from '@testing-library/react';
 import * as React from 'react';
 import { createStore } from 'redux';
 import { Provider } from 'react-redux';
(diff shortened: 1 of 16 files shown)


autoscheduler/frontend/src/tests/ui/SelectTerm.test.tsx
```diff
@@ -12,60 +18,85 @@ jest.mock('@reach/router', () => ({
...
       const store = createStore(autoSchedulerReducer);
 
-      const { getByText } = render(
-        <Provider store={store}>
-          <SelectTerm />
-        </Provider>,
-      );
-
+      let getByText: Function;
+      await act(async () => {
```
firejake308, line 86: The way that I've been doing this is, instead of wrapping the render in `act`, I swap `getByText` for `findByText`, which you can use below with an `await`. I think that reads a little better, if it works

autoscheduler/frontend/src/tsconfig.json
```diff
@@ -4,7 +4,7 @@
         "sourceMap": true,
         "noImplicitAny": true,
         "module": "commonjs",
-        "target": "es6",
+        "target": "es2018",
```
firejake308, line 7: Just to clarify, this was needed for the `new Map()` you're using for the termMap?

ryan-conn: It's needed for the function `Object.entries(obj)`, technically we only need to use es2017 for that function but I chose to use 2018 because that's what our `eslintrc` was already configured to

ryan-conn:
Implemented suggestions, I also noticed looking at the API for react testing library that `waitForX` functions have been deprecated in favor of a generic `waitFor` function, so I updated our `package.json` and tests to use `waitFor`. If you get peer dependency warnings, update your npm installation, as these are optional and npm 6.11.0 adds optional peer dependencies (if you have a version below that, they'll be treated as required).

autoscheduler/frontend/src/tests/ui/SelectTerm.test.tsx
```diff
@@ -12,60 +18,85 @@ jest.mock('@reach/router', () => ({
...
+    'Summer 2019': '201921',
+    'Spring 2019': '201911',
+  }));
+};
+
+
 describe('SelectTerm', () => {
+  beforeEach(mockTermsAPI);
+
+  afterEach(fetchMock.mockRestore);
```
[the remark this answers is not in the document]
ryan-conn: Yeah you're right

autoscheduler/frontend/src/tests/ui/SelectTerm.test.tsx
```diff
@@ -12,60 +18,85 @@ jest.mock('@reach/router', () => ({
...
+          <Provider store={store}>
+            <SelectTerm />
+          </Provider>,
+        ));
+      });
 
       // act
       const button = getByText('Select Term');
-      fireEvent.click(button);
+      act(() => { fireEvent.click(button); });
```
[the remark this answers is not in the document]
ryan-conn: Yeah, I did that because I saw other test wrap `fireEvent` in act (and I was getting warnings about changing state outside of `act`) so I had assumed it was still necessary, but the warnings were only coming from the `render` calls

PR #234, opened 2020-04-28
Parse instructor fix

gannonprudhomme:
Small bug, but `parse_instructor` would return `None` anytime an instructor was already visited, which resulted in lot of sections having `None` instructors when they shouldn't (i.e. for `201531`). This fixes that bug and adds a test for it.

### DIFF
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -120,7 +120,7 @@ def parse_instructor(course_data, instructors_set) -> Instructor:
 
         if name not in instructors_set:
             instructors_set.add(name)
-            return instructor_model
+        return instructor_model
 
     return None
 
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -98,7 +98,7 @@ def parse_meeting(meetings_data, section: Section, meeting_count: int) -> Meetin
                             meeting_type=class_type, section=section)
     return meeting_model
 
-def parse_instructor(course_data, instructors_set) -> Instructor:
+def parse_instructor(course_data) -> Instructor:
     """ Parses the instructor data and saves it as a Instructor model.
         Called from parse_course.
     """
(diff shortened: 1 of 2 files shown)


gannonprudhomme:
Oh wait I just realized that this is going to be saving extra/duplicate instructors, so I'm going to move that `instructors_set` check outside of it probably

autoscheduler/scraper/tests/commands/scrape_courses_tests.py
```diff
@@ -124,14 +124,31 @@ def test_parse_instructor_does_save_model(self):
...
+        # Both sections have John M. Moore as the professor
+        # Parse the professor for the first time, which adds it to the instructor_set
+        instructor1 = parse_instructor(self.csce_section_json)
+
+        # Act
+        instructor2 = parse_instructor(self.csce_web_section_json)
+
+        # Assert
+        self.assertEqual(instructor1, instructor2)
+        self.assertNotEqual(instructor2, None)
```
ryan-conn, line 150:
```suggestion
        self.assertIsNotNone(instructor2)

```

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -147,9 +143,15 @@ def parse_course(course_data: List,
     credit_hours = course_data['creditHourLow']
 
     # Parse the instructor, then send the returned Instructor model to parse_section
-    instructor_model = parse_instructor(course_data, instructors_set)
+    instructor_model = parse_instructor(course_data)
     section_data = parse_section(course_data, instructor_model)
 
+    if not instructor_model is None and instructor_model.id not in instructors_set:
+        instructors_set.add(instructor_model.id)
```
ryan-conn, line 150:
```suggestion
    if instructor_model is not None:
        instructors_set.add(instructor_model)
```
Hashing a model just hashes its primary key so it's fine to pass in the instructor directly, and there's no reason to check if something is in a set before adding it: if it's already a member, it'll just do nothing.
Also `is not` is an operator

gannonprudhomme: Will change the `is not` operator, but we want to set `instructor_model` to None whenever the `instructor_model` is already in the `instructor_set`, so I'm leaving that check. Otherwise it would set `instructor_model` to None whenever it is None, which doesn't make much sense.

gannonprudhomme:
Squashed the `self.assertIsNotNone` into the first commit, and the other comment into the second commit.

gannonprudhomme:
@ryan-conn look at this when you get the time I wanted it meeerrrgggeeeddddd

PR #244, opened 2020-05-03
Refactored parseMeetings to parseSections

### LINKED ISSUE #241: Optimize updateCourseCardAsync
The way we calculate sections for meetings right now seems unnecessarily complicated: first we create a section for each section in the response, then we find all meetings belonging to that section and only return the meetings, then sort the meetings to determine which section they were originally in and create `SectionSelected` objects based on that. I propose that we do the following:
- Instead of giving each `Meeting` a `Section`, give the `Section` type an array of its meetings
- Directly return an array of sections with their meetings in `parseMeetings` and completely remove `meetingsToSectionSelected`

Also, since the only property exclusive to `SectionSelected` is `selected`, we could either move it to the `Section` type (and just give it a default value of `null`), or keep a boolean array of whether sections are selected in the state for each course card. It seems a little overkill to make another class just to add one attribute to `Section` when we can just ignore the property when we don't need it.

Let me know what you guys think @gannonprudhomme @firejake308, I don't see any reason we shouldn't be able to make these changes but perhaps I'm missing something that you guys noticed when working on those functions

ryan-conn:
Optimizes the way that `SectionSelected` objects are created by parsing them directly from the section data we get from the API. I also added `hasHonors` and `hasWeb` as parts of the course card state, so that they only have to be calculated when sections are parsed again.

I realized that it's far more effort than it's worth to remove the `SectionSelected` class (as it would create a circular dependency between `Section` and `Meeting` and even if I make `Meeting` just take section nums/ids instead of a section object, it would involve a huge amount of refactoring), so I didn't change it at all.

Closes #241.

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicSelect.tsx
@@ -3,7 +3,6 @@ import { Typography, FormLabel } from '@material-ui/core';
 import { useSelector } from 'react-redux';
 import { RootState } from '../../../../../../redux/reducer';
 import * as styles from './BasicSelect.css';
-import { SectionSelected } from '../../../../../../types/CourseCardOptions';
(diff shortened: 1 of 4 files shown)


ryan-conn:
Rebased to squash changing the name of `parseSections` tests into the commit with other changes to the file

Issue #238, opened 2020-04-29
Align the section type/day/time columns in SectionSelect


gannonprudhomme:
After working on #184 I realized that the section type / day / time columns weren't aligned with each other

![chrome_m3t01jk59P.png](https://images.zenhubusercontent.com/5da358de33f74500019c3b7f/e1267165-9995-4671-90f6-607923ae15fe)

We should align them so they look something like this (ignore the periods, just did that so I could get my point across lol). I don't think each actual section's columns need to align with different ones, but the meetings within a single section should align with each other. 
![chrome_ADcMwU3f1h.png](https://images.zenhubusercontent.com/5da358de33f74500019c3b7f/24ed24ab-aa0f-406d-9c55-d3e6936d0a4b)

firejake308:
I think using an HTML `<table>` would help us with this. I can make a PR to convert the current layout to use `<table>` if you'd like

ryan-conn:
Sure, that's what I was thinking we would have to do to align columns so go ahead

gannonprudhomme:
Yeah feel free to do this, I tried doing it with `display: grid` and stuff then got too frustrated and stopped lol

PR #248, opened 2020-05-06
Show ONLINE for online meetings in SectionSelect

### LINKED ISSUE #184: Make course cards display Web/Online for online courses
If a section is online, the times should show "Online" or "Web" instead of "00:00" like it currently does.

~~This issue could also include making the "web only" and "honors only" check boxes functional, and have it make the sections only show web only or honors only (as well as sending it to `scheduler/generate`)~~

gannonprudhomme:
This adds the showing of ONLINE for meeting times that are `00:00`. I used the assumption that if the startTime for a meeting is 0, then it's safe to assume that the rest of them are also 0, meaning that it's an online meeting time.

Closes #184

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -46,7 +46,8 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
 
   const getMeetingTimeText = (mtg: Meeting): string => {
     if (mtg.startTimeHours === 0) {
-      return 'ONLINE';
+      // If the time is 00:00 but it's not honors, then showing nothing for the meeting time
+      return mtg.section.web ? 'ONLINE' : '';
     }
 
     // Returns it in the format 12:00 - 1:00
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -44,6 +44,16 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
     return meeting.meetingDays.reduce((acc, curr, idx) => (curr ? acc + DAYS_OF_WEEK[idx] : acc), '');
   };
 
+  const getMeetingTimeText = (mtg: Meeting): string => {
+    if (mtg.startTimeHours === 0) {
+      return 'ONLINE';
+    }
+
+    // Returns it in the format 12:00 - 1:00
+    return `${formatTime(mtg.startTimeHours, mtg.startTimeMinutes)}
+      - ${formatTime(mtg.endTimeHours, mtg.endTimeMinutes)}`;
+  };
+
   const renderMeeting = (mtg: Meeting, showSectionNum: boolean): JSX.Element => (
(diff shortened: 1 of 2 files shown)


gannonprudhomme:
Doing this made me realize the alignment of the times/days/sections are off, so I made #238

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -56,8 +73,7 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
         {formatMeetingDays(mtg)}
       </span>
       <span className={styles.meetingDetailsText}>
-        {`${formatTime(mtg.startTimeHours, mtg.startTimeMinutes)}
-        - ${formatTime(mtg.endTimeHours, mtg.endTimeMinutes)}`}
+        {getMeetingTimeText(mtg)}
```
firejake308, line 76: I think you could just have `getMeetingTimeText` return the text instead of a span with the text, to reduce unnecessary clutter in the DOM. The way it is right now, there's a `<span>` inside of a `<span>`

firejake308:
Seems to work fine in browser. Only issue is that there might be a better way to get the alignment. Have you considered deleting the meetingDays span if the section is online?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
gannonprudhomme: Yeah that makes a lot more sense, squashed into the first commit

gannonprudhomme:
> Seems to work fine in browser. Only issue is that there might be a better way to get the alignment. Have you considered deleting the meetingDays span if the section is online?

Honestly I'm not going to worry about the alignment too much in this, and instead we can just deal with it in #238 . However I did this (by making the `meetingDetails` span be null when `startTimeHours === 0`), which resulted in this:

chrome_fgn6u2T5V9

So not sure if we want it like this or not? Personally I like the ONLINE text being where the times are, but I think this just comes down to what we think looks the best.

firejake308:
Yeah, ONLINE text looks better wher the times are

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -44,6 +44,16 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
     return meeting.meetingDays.reduce((acc, curr, idx) => (curr ? acc + DAYS_OF_WEEK[idx] : acc), '');
   };
 
+  const getMeetingTimeText = (mtg: Meeting): string => {
+    if (mtg.startTimeHours === 0) {
+      return 'ONLINE';
+    }
```
ryan-conn, line 50: My only problem with this is that checking if the start time is 0 includes all meetings with null start times, which aren't exclusively online classes (for example, I'm taking JAPN 491 next semester, and it has a null meeting time since it's a research class but isn't online, despite showing as online on our website). Most research classes are going to be this way, and there are probably other exceptions. I believe we should check if the section for meeting times of 0 are online—if so, print "ONLINE", and if not, just don't print anything (if we want, we could change it to another message but I think leaving it empty should be fine).

gannonprudhomme: Good point, will add

gannonprudhomme:
Made it show nothing when a section has null meeting times but it's not an online section, then rebased to be up to date with master

PR #251, opened 2020-05-08
Added icon to honors sections

### LINKED ISSUE #230: Show honors sections in Course Cards
We need to show that a specific section is honors so the users can figure out which sections they want. In the 4/26 meeting we decided that we're going to 

We need to make the course card's section displaying to split the sections-instructor grouping to put honors and non-honors sections into different "instructor-groups". Below is what I have mocked up for it, albeit with the out-of-date designs. I personally like the one that explicitly says `Honors` rather than use an icon as I think it's much more explicit. For the icon I was thinking we'd also have it just show a tooltip that said "honors"/"honors sections" on hover.

firejake308:
Added an icon for honors sections and appropriate tests. Currently, the icon is gray to match th professor's name, and there isn't a whole lot of space between the text and the icon. Seems fine to me, but if y'all don't like it, it's an easy fix.
image

Closes #230

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -75,8 +75,11 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
 
   const makeList = (): JSX.Element[] => {
     let lastProf: string = null;
+    let lastHonors = false;
     return sections.map(({ section, selected, meetings }, secIdx) => {
-      const instructorLabel = lastProf !== section.instructor.name
+      if (section.subject === 'MATH' && section.instructor.name === 'Aakash Tyagi') debugger;
+      const makeNewGroup = lastProf !== section.instructor.name || lastHonors !== section.honors;
+      const instructorLabel = makeNewGroup
         ? (
           <ListSubheader disableGutters className={styles.listSubheaderDense}>
             {section.instructor.name}
@@ -89,6 +92,7 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
         )
         : null;
       lastProf = section.instructor.name;
+      lastHonors = section.honors;
 
       // get the meetings that match this section
       const sectionDetails = (
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import {
-  List, ListItemText, ListItem, Checkbox, ListItemIcon, Typography, ListSubheader,
+  List, ListItemText, ListItem, Checkbox, ListItemIcon, Typography, ListSubheader, Tooltip,
 } from '@material-ui/core';
 import HonorsIcon from '@material-ui/icons/School';
 import { useSelector, useDispatch } from 'react-redux';
@@ -80,7 +80,11 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
         ? (
           <ListSubheader disableGutters className={styles.listSubheaderDense}>
             {section.instructor.name}
-            {section.honors ? <HonorsIcon titleAccess="honors" /> : null}
+            {section.honors ? (
+              <Tooltip title="Honors" placement="right">
+                <HonorsIcon titleAccess="honors" />
+              </Tooltip>
+            ) : null}
           </ListSubheader>
         )
         : null;
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -84,7 +84,7 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
             {section.instructor.name}
             {section.honors ? (
               <Tooltip title="Honors" placement="right">
-                <HonorsIcon titleAccess="honors" />
+                <HonorsIcon data-testid="honors" />
               </Tooltip>
             ) : null}
           </ListSubheader>
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -77,7 +77,6 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
     let lastProf: string = null;
     let lastHonors = false;
     return sections.map(({ section, selected, meetings }, secIdx) => {
-      if (section.subject === 'MATH' && section.instructor.name === 'Aakash Tyagi') debugger;
       const makeNewGroup = lastProf !== section.instructor.name || lastHonors !== section.honors;
       const instructorLabel = makeNewGroup
         ? (
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import {
   List, ListItemText, ListItem, Checkbox, ListItemIcon, Typography, ListSubheader,
 } from '@material-ui/core';
+import HonorsIcon from '@material-ui/icons/School';
 import { useSelector, useDispatch } from 'react-redux';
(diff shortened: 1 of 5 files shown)


ryan-conn:
I think that hovering over the honors icon should make help text appear that says "Honors" or "Honors Section" or something like that, just to make it a little easier for users. Luckily MaterialUI includes the `Tooltip` component so this shouldn't be too difficult to implement

firejake308:
Good idea

firejake308:
Got the tooltip and also separated regular vs honors sections of the same professor.

gannonprudhomme:
Just a thought, but instead of grouping them "in order", we might want to globally group all of the sections for a professor together, regardless of whether the sections they have for a given course are in order or not. For instance, for PHYS 201 for Fall 2020, sections 506, 511-514, and 519-524 are all TBA sections. We might want to group all of these together so they're shown in one block, then we can sort all of the instructors group by the first section number they have, so the TBA block would show after section 505 and before section 510 (which isn't a TBA setion). That way no matter what sections a prof has (other than one having honors and non-honors sections) they'll only be one grouping of that professor. 

For instance, PHYS 201 - Fall 2020 would look like:
```
Lewis Ford:
501 - 505

TBA:
506, 
511 - 514, 
519 - 524

Vladimir V. Khmelenko:
507 - 510

Katrin Becker:
515 - 518

Junaid M. Farooqi:
M01
```

Thoughts?

firejake308:
Force push was done after rebasing with master to resolve a conflict

firejake308:
Two questions:
1. Where in the list would the TBAs end up? With the lowest TBA section number, or should TBA always be at the bottom? 
2. Also, how would you deal with a professor who teaches both honors and non-honors, but the honors is like 200 and non-honors is like 599? Would they be together or separated?

gannonprudhomme:
> 1. Where in the list would the TBAs end up? With the lowest TBA section number, or should TBA always be at the bottom?

It'd honestly probably be best to have TBA's be a the bottom, but I think it'd be kind of weird to hard code that in so I think the lowest TBA section number should be fine

> 2. Also, how would you deal with a professor who teaches both honors and non-honors, but the honors is like 200 and non-honors is like 599? Would they be together or separated?

I'd say make them separate

firejake308:
Should we separate out the tests for grouping by instructor and sorting by section number? I think an integration test would be better at catching errors, but I'm open to adding more tests

firejake308:
Also, if a professor has honors sections, their other sections will also be sorted to the top. I think the sorting algorithm is complicated enough as it is, so I didn't bother with fixing this relatively minor issue

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -93,6 +93,45 @@ export function parseSections(arr: any[]): SectionSelected[] {
...
+  // sort professor groups by section number
+  profs = profs.sort((a, b) => a.sectionNum.localeCompare(b.sectionNum));
+
+  // rebuild sections
+  const sorted: SectionSelected[] = [];
+  profs.forEach((prof) => sorted.push(...sections.slice(prof.start, prof.end)));
+
+  return sorted;
+}
+
```
ryan-conn, line 134: A couple things about this sort:
- I don't think it works exactly as intended, for example when I try to see the results for MATH 151 in 202031, it shows TBA honors, then TBA non-honors, then more TBA honors, then more TBA non-honors sections
- I don't really like how much custom code this uses to sort, there are a lot of points it could potentially make mistakes and it isn't very easy to read

I've tested this code and it seems to work fine (it also makes it so we don't need to separately sort by instructor names like we do on line 155), another advantage is that it's easy to adapt if we want to change the sort order (for example, have `sectionToArr` check if the prof name is TBA, and if so set the field to ZZZZ or something so it shows up last):
```js
function sortSections(sections: SectionSelected[]): SectionSelected[] {
  // converts a section to an array of relevant fields for sort
  const sectionToArr = (section: SectionSelected, idx: number): any[] => (
    [section.section.instructor.name, section.section.sectionNum, idx]);
  const sortedArrs = sections.map(sectionToArr).sort();
  // sort sections based on their positions in the sorted array of arrays
  const sorted = sortedArrs.map((section) => sections[section[2]]);
  return sorted;
}
```

firejake308: Wow, that is a really clean way to write it! Thanks for the suggestion. I'll add a test for the TBA thing, then rewrite the sort using this version

ryan-conn: I came up with a way to do it that displays professors ordered by their lowest section number (which I believe is how we wanted to do it) but is a little more complicated:
```js
function sortSections(sections: SectionSelected[]): SectionSelected[] {
  // sort sections by sectionNum
  const sorted = sections.sort((a: SectionSelected, b: SectionSelected) => {
    if (a.section.sectionNum < b.section.sectionNum) return -1;
    return 1;
  });
  const sectionsForProfs: Map<string, SectionSelected[]> = new Map();
  // maps maintain key insertion order, so add all sections to map and remember order of professors
  sorted.forEach((section) => {
    const instructorName = section.section.instructor.name;
    if (sectionsForProfs.has(instructorName)) sectionsForProfs.get(instructorName).push(section);
    else sectionsForProfs.set(instructorName, [section]);
  });
  // sections are now grouped by professor and sorted by section num
  return [].concat(...sectionsForProfs.values());
}
```
I think the following change correctly puts TBA sections last, but I haven't found any courses that I can test this with/made any test data for it:
```js
function sortSections(sections: SectionSelected[]): SectionSelected[] {
  // sort sections by sectionNum
  const sorted = sections.sort((a: SectionSelected, b: SectionSelected) => {
    if (a.section.sectionNum < b.section.sectionNum) return -1;
    return 1;
  });
  const sectionsForProfs: Map<string, SectionSelected[]> = new Map();
  // maps maintain key insertion order, so add all sections to map and remember order of professors
  const TBASections: SectionSelected[] = [];
  sorted.forEach((section) => {
    const instructorName = section.section.instructor.name;
    if (instructorName === 'TBA') {
      TBASections.push(section);
    } else if (sectionsForProfs.has(instructorName)) {
      sectionsForProfs.get(instructorName).push(section);
    } else sectionsForProfs.set(instructorName, [section]);
  });
  // sections are now grouped by professor and sorted by section num
  return [].concat(...sectionsForProfs.values(), TBASections);
}
```

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -64,15 +65,23 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
...
     return sections.map(({ section, selected, meetings }, secIdx) => {
-      const instructorLabel = lastProf !== section.instructor.name
+      const makeNewGroup = lastProf !== section.instructor.name || lastHonors !== section.honors;
+      const instructorLabel = makeNewGroup
         ? (
           <ListSubheader disableGutters className={styles.listSubheaderDense}>
             {section.instructor.name}
+            {section.honors ? (
+              <Tooltip title="Honors" placement="right">
+                <HonorsIcon titleAccess="honors" />
```
ryan-conn, line 77: I don't think we want this `titleAccess` now that we have an actual tooltip, since now there are two of them:
image

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -93,6 +93,26 @@ export function parseSections(arr: any[]): SectionSelected[] {
   return sections;
 }
 
+function sortSections(sections: SectionSelected[]): SectionSelected[] {
```
gannonprudhomme, line 96: I'd add doc comments to this describing the general sorting idea (i.e. grouping, sort groups by smallest section number, TBA sections last)

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
firejake308: Unfortunately, I was using the built-in tooltip during testing, so I had to replace that with a `testId`. However, Material's tooltip looks better than the built-in, so I think that this is fine

gannonprudhomme:
We're changing the sorting for this again, correct? Just make sure cause I forgot

Also really just wanna get this merged soon since it's been open for almost a month at this point

firejake308:
Yeah, we're changing it to sort all honors sections to the top IIRC

gannonprudhomme:
> Yeah, we're changing it to sort all honors sections to the top IIRC

Yeah I uh forgot about the meeting notes lmao

gannonprudhomme:
ackkkk I did not mean to do that, whatta day

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -107,8 +107,9 @@ function sortSections(sections: SectionSelected[]): SectionSelected[] {
   // maps maintain key insertion order, so add all sections to map and remember order of professors
   const TBASections: SectionSelected[] = [];
   sorted.forEach((section) => {
-    const instructorName = section.section.instructor.name;
-    if (instructorName === 'TBA') {
+    // H stands for honors, R stands for regular
+    const instructorName = section.section.instructor.name + (section.section.honors ? 'H' : 'R');
+    if (instructorName.startsWith('TBA')) {
```
ryan-conn, line 112: Is there a case where checking this instead of equality will make a difference? I don't really see a reason to change it back to how it was seeing as both of these should take pretty much the same amount of time, but curious why you made the change

firejake308: The actual name could either be `TBAH` or `TBAR`

ryan-conn: Oh yeah that's right, I don't know how I didn't notice the order it's in (change name then check if it's TBA), I'd like if you could change it so it checks if the section is TBA first since there isn't really a reason to change the string first, whenever you do I can reapprove this real quick

firejake308: If I were using the site, I would want to still see honors sections at the top, even if the professor was TBA, so I changed it to behave that way. Lmk if y'all like that

firejake308:
Rebased with master, should be ready for review

gannonprudhomme:
Lets merged this mf 🚀

PR #253, opened 2020-05-08
Sort sections by section number

### LINKED ISSUE #252: Sort sections by section number
While working on #251, I noticed that the order of sections is random. This creates a problem because all of one professor's sections should be together, but they end up getting split. To fix the problem, sections should be presented in order of increasing section number.

firejake308:
Ez fix. ~I'm lazy, so I'll un-draft when I add tests later~ Added tests. Ready for review.

Closes #252


autoscheduler/frontend/src/tests/ui/dummyData.ts
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export const dummyMeetingArgs = {
+  id: 12345,
+  building: 'HRBB',
+  days: [true, false, true, false, true, false, false],
+  start: '08:00',
+  end: '08:50',
+  type: 'LEC',
+};
```
gannonprudhomme, line 23: Not used, so can be removed

autoscheduler/frontend/src/tests/ui/dummyData.ts
```diff
@@ -0,0 +1,23 @@
+/* eslint-disable @typescript-eslint/camelcase */
```
gannonprudhomme, line 1: Add above the `eslint-disable` saying like `// Disable so we can represent the data returned from api/sections/ with snake case` or whatever. Just wanna document the ignores generally, as there's been a few times that I've looked at old ones and am not positive why they're needed

PR #254, opened 2020-05-16
Show remaining seats in SectionSelect

### LINKED ISSUE #186: Add displaying of empty/filled seats to course cards

ryan-conn:
`SectionSelect` now shows the available seats for each section. I changed the section number to show on its own line so that this can look better on low resolutions (specifically, I wanted to make it still look good on 1366x768 since some people might still have laptops with that terrible resolution), and still be self-explanatory, rather than just having it say "0/50" or something like that.

This is a draft for now since I haven't made tests yet, and this is based off of Adel's `frontend/show-honors` branch.

A couple design choices I made, but feel free to argue the other way if you feel there's a reason to do so:
- Shows max seats instead of just available seats so that people can easily tell if a section is really full, or if it just hasn't had all of its seats opened yet
- Seats are right-aligned which might look a little weird on high resolutions (for example, this is what it looks like on my 1440p monitor): 
image

Closes #186.


gannonprudhomme:
Not necessarily related to this PR, but I also think if you manually selected a full section, you shouldn't have to press `Include full sections` in order for it to show up on the schedule, since you obviously want it to be included. Maybe we should move the `Include full sections` option to the `BASIC` customization level for each course cards? idk just a thought

gannonprudhomme:
Imo the `0 of 10 seats remaining` format is too long/wordy and doesn't seem to scale too well, maybe `0/10 seats left`? I do think that what you have now is probably the most clear, but I think the reduced space usage is worth losing some verbose-ness. What do y'all think @firejake308 @ThatJuanGuy 

Here's what it looks like on my 1080p screen for reference
chrome_4rT3wdmDI6

gannonprudhomme:
At this point, I'd just add the tests and make this a full PR so we can review the rest of the code, and we can discuss my above comments in the meeting tomorrow.

ryan-conn:
Yeah I just started working on them actually, so this should stop being a draft later today

ThatJuanGuy:
I think using left is also better than using available/remaining.

ryan-conn:
Tests created and I reformatted how available seats display to what Gannon suggested

autoscheduler/frontend/src/tests/ui/CourseSelectCard.test.tsx
```diff
@@ -463,4 +569,93 @@ describe('Course Select Card UI', () => {
...
+
+      fireEvent.click(getByText('Section'));
+
+      // assert
+      // verify text for seats is correct
+      const sectionText = (await findAllByText(/ seats left/))[1];
+      expect(sectionText).toHaveTextContent('0/25 seats left');
+    });
+
+    test('when a section has over max enrollement', async () => {
```
ryan-conn, line 632: Oops I spelled this wrong, will fix when I rebase this on master

autoscheduler/frontend/src/tests/ui/CourseSelectCard.test.tsx
```diff
@@ -463,4 +569,93 @@ describe('Course Select Card UI', () => {
...
+      const courseEntry = getByLabelText('Course') as HTMLInputElement;
+      fireEvent.change(courseEntry, { target: { value: 'CSCE ' } });
+      fireEvent.click(await findByText('CSCE 121'));
+
+      fireEvent.click(getByText('Section'));
+
+      // assert
+      // verify text for seats is correct
+      const sectionText = (await findAllByText(/ seats left/))[2];
+      expect(sectionText).toHaveTextContent('-1/25 seats left');
```
ryan-conn, line 658: Part of the reason I made this test is that I was wondering how we wanted to display sections that have more seats taken than there are total (which I've noticed happens pretty frequently when trying to register for classes in the past), and I can see arguments for either way. Showing negative seats gives more information and is technically correct, while showing "0/25 seats left" instead would look nicer. Feel free to share your thoughts if you think I should change this

ryan-conn:
> Not sure what changed, but the section number isn't on it's own row anymore.

That happened because I changed a `div` to a `React.Fragment` since I assumed the only reason it was there was as a container for multiple elements, when it was being used for the line break as well. Should be fixed now

firejake308:
image
I think it would be helpful to either bold the section number or underline the first row to separate it from the meeting list

gannonprudhomme:
Rebase this with master when you get the chance @ryan-conn

ryan-conn:
Rebased with master and I decided to put a line underneath section number rather than bold it, as bold looks a bit odd imo
Here's with bold text:
image
And here's with a material ui divider:
image
I also had to change how SectionSelect is laid out to silence some warnings about having a `<div>` as a descendant of `<p>` since the whole thing was wrapped in typography, but as far as I can tell it looks the same.

gannonprudhomme:
Just as a reminder, here's what we decided on changing for this:

- Add a line under the professor instead of under the section number
- Gray out the section text (or make it lighter)

ryan-conn:
> Just as a reminder, here's what we decided on changing for this:
> 
>     * Add a line under the professor instead of under the section number
> 
>     * Gray out the section text (or make it lighter)

I implemented those changes locally, but making the section text grey instead of the meeting text looks a little weird so I'm not going to push yet, here's a comparison of both:
image

firejake308:
> I implemented those changes locally, but making the section text grey instead of the meeting text looks a little weird so I'm not going to push yet, here's a comparison of both:

I like the one on the right better, because the section number feels like it should be more important than the meeting times and consequently have darker text

gannonprudhomme:
> > I implemented those changes locally, but making the section text grey instead of the meeting text looks a little weird so I'm not going to push yet, here's a comparison of both:
> 
> I like the one on the right better, because the section number feels like it should be more important than the meeting times and consequently have darker text

Also definitely like the one on the right better for the same reasons as Adel

gannonprudhomme:
@ryan-conn Can you rebase this with master? Dunno what caused this conflict tbh

gannonprudhomme:
Also this has been open for two months :(

ryan-conn:
Rebased with master, I also had all the changes that were requested (moving section header + changing text color) in a stash that I thought I had committed, so this should hopefully be final now

PR #255, opened 2020-05-17
Connected generate schedules

### LINKED ISSUE #193: Connect "Fetch Schedules" with /scheduler/generate

gannonprudhomme:
Connects the `Generate schedules` button to /scheduler/generate.

Currently I handle the BASIC/SECTIONS customization level by if the SECTIONS option is selected, then the options for the honors/web only are set to the default. And if the BASIC option is selected, then the sections aren't sent. We might want to modify this for sessions so we *always* have the data in order to save it, and we can instead just send the current customization level for each section, and `create_schedules` can then choose which one it wants to ignore. Should be a relatively easy fix.

Also think we might want to add some tests for this that I missed, as the availabilities time converting was messed up and I didn't catch it for awhile, so lmk if you have any ideas for tests to add.

I also had to change the course cards actions which sort of (not really) reverts the changes Ryan made in #244 since I actually do need the original `parseMeetings` functionality for this, but I did my best extracting-out the common functionality of it all.

Closes #193

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -80,15 +80,9 @@ const ConfigureCard: React.FC = () => {
       }),
     }).then(
       (res) => res.json(),
-    ).then((data: any[][]) => {
-      const ret: Meeting[][] = [];
-
-      data.forEach((schedule) => {
-        ret.push(parseAllMeetings(schedule));
-      });
-
-      return ret;
-    }).then((schedules: Meeting[][]) => {
+    ).then((generatedSchedules: any[][]) => generatedSchedules.map(
+      (schedule) => parseAllMeetings(schedule),
+    )).then((schedules: Meeting[][]) => {
       dispatch(replaceSchedules(schedules));
       dispatch(selectSchedule(0));
       setLoading(false);
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -9,7 +9,7 @@ import * as styles from './ConfigureCard.css';
 import { replaceSchedules } from '../../../redux/actions/schedules';
 import selectSchedule from '../../../redux/actions/selectedSchedule';
 import Meeting from '../../../types/Meeting';
-import { parseMeetings } from '../../../redux/actions/courseCards';
+import { parseAllMeetings } from '../../../redux/actions/courseCards';
 // DEBUG
 import { RootState } from '../../../redux/reducer';
 import { CourseCardArray, CustomizationLevel } from '../../../types/CourseCardOptions';
@@ -24,7 +24,7 @@ const ConfigureCard: React.FC = () => {
   const [includeFull, setIncludeFull] = React.useState(false);
   const [loading, setLoading] = React.useState(false);
   const courseCards = useSelector<RootState, CourseCardArray>((state) => state.courseCards);
-  const term = useSelector<RootState, number>((state) => state.term);
+  const term = useSelector<RootState, string>((state) => state.term);
   const avsList = useSelector<RootState, Availability[]>((state) => state.availability);
   const dispatch = useDispatch();
 
@@ -50,12 +50,16 @@ const ConfigureCard: React.FC = () => {
         const [subject, courseNum] = courseCard.course.split(' ');
         const isBasic = courseCard.customizationLevel === CustomizationLevel.BASIC;
 
+        // The default option for honors and web when the Section customization level is selected
+        const filterDefault = 'no_preference';
+
         courses.push({
           subject,
           courseNum,
           sections: isBasic ? [] : selectedSections, // Only send if "Section" customization level
-          honors: isBasic ? courseCard.honors ?? null : null, // Only send if "Basic" level
-          web: isBasic ? courseCard.web ?? null : null,
+          // Only send if "Basic" level
+          honors: isBasic ? courseCard.honors ?? filterDefault : filterDefault,
+          web: isBasic ? courseCard.web ?? filterDefault : filterDefault,
         });
       }
     }
@@ -80,7 +84,7 @@ const ConfigureCard: React.FC = () => {
       const ret: Meeting[][] = [];
 
       data.forEach((schedule) => {
-        ret.push(parseMeetings(schedule));
+        ret.push(parseAllMeetings(schedule));
       });
 
       return ret;
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -35,11 +35,20 @@ const ConfigureCard: React.FC = () => {
     const courses = [];
     for (let i = 0; i < courseCards.numCardsCreated; i++) {
       if (courseCards[i] && courseCards[i].course) {
+        const selectedSections: string[] = []; // the section nums of the selected sections
+
+        // Iterate through the sections and only choose the ones that are selected
+        courseCards[i].sections.forEach((sectionSel) => {
+          if (sectionSel.selected) {
+            selectedSections.push(sectionSel.section.sectionNum);
+          }
+        });
+
         const [subject, courseNum] = courseCards[i].course.split(' ');
         courses.push({
           subject,
           courseNum,
-          sections: courseCards[i].sections,
+          sections: selectedSections,
           honors: courseCards[i].honors,
           web: courseCards[i].web,
         });
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -65,8 +65,8 @@ const ConfigureCard: React.FC = () => {
     }
     // make availabilities object
     const availabilities = avsList.map((avl) => ({
-      startTime: formatTime(avl.startTimeHours, avl.startTimeMinutes).replace(':', ''),
-      endTime: formatTime(avl.endTimeHours, avl.endTimeMinutes).replace(':', ''),
+      startTime: formatTime(avl.startTimeHours, avl.startTimeMinutes, true, true).replace(':', ''),
+      endTime: formatTime(avl.endTimeHours, avl.endTimeMinutes, true, true).replace(':', ''),
       day: avl.dayOfWeek,
     }));
 
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -12,7 +12,7 @@ import Meeting from '../../../types/Meeting';
 import { parseMeetings } from '../../../redux/actions/courseCards';
 // DEBUG
 import { RootState } from '../../../redux/reducer';
-import { CourseCardArray } from '../../../types/CourseCardOptions';
+import { CourseCardArray, CustomizationLevel } from '../../../types/CourseCardOptions';
 import Availability from '../../../types/Availability';
 import { formatTime } from '../../../timeUtil';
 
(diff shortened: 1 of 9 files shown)


autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -36,57 +36,98 @@ function updateCourseCardSync(index: number, courseCard: CourseCardOptions): Upd
...
+    });
+
+    meetings.push(meeting);
+  });
+
+  return meetings;
+}
+
+/**
+ * Helper function that parses all of the meetings in the section data and returns them.
```
firejake308, line 100: Based on the docstrings, `parseAllMeetings` and `parseMeetings` seem to do the same thing. Could you clarify that a bit?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -36,57 +36,98 @@ function updateCourseCardSync(index: number, courseCard: CourseCardOptions): Upd
...
+  return ret;
+}
+
 /**
  *  Helper function to parse the serialized Section model from the backend convert it into an
  *  array of Sections
  *  @param arr The array of sections returned from the backend, such as from api/sections
  */
-export function parseSections(arr: any[]): SectionSelected[] {
+export function parseSections(arr: any[]): SectionSelected[] { // Rename to parseSectionSelected?
```
firejake308, line 123: I think we probably should rename it, just so that no one is surprised by the return type

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -36,57 +36,98 @@ function updateCourseCardSync(index: number, courseCard: CourseCardOptions): Upd
...
-      honors: sectionData.honors,
-      web: sectionData.web,
-      instructor: new Instructor({ name: sectionData.instructor_name }),
-      grades: sectionData.grades == null ? null : new Grades(sectionData.grades),
-    });
-
-    sectionData.meetings.forEach((meetingData: any) => {
-      let start: string[] = ['0', '0'];
-      let end: string[] = ['0', '0'];
+    const section = parseSection(sectionData);
```
firejake308, line 127: So basically, these two functions have been separated out into helper functions, and those helper functions are re-used between `parseSections` and `parseAllMeetings`, right?

autoscheduler/frontend/src/tests/ui/ConfigureCard.test.tsx
```diff
@@ -1,19 +1,42 @@
...
+import { createStore, applyMiddleware } from 'redux';
 import { Provider } from 'react-redux';
+import thunk from 'redux-thunk';
 import ConfigureCard from '../../components/SchedulingPage/ConfigureCard/ConfigureCard';
-import fetch from '../../components/SchedulingPage/ConfigureCard/generateSchedulesMock';
 import autoSchedulerReducer from '../../redux/reducer';
+import { updateCourseCard } from '../../redux/actions/courseCards';
+import { CustomizationLevel } from '../../types/CourseCardOptions';
 
 jest.mock('../../components/SchedulingPage/ConfigureCard/generateSchedulesMock', () => ({
```
firejake308, line 17: We can probably remove this mock

autoscheduler/frontend/src/tests/ui/ConfigureCard.test.tsx
```diff
@@ -1,19 +1,42 @@
...
 
-describe('OptionsCard component', () => {
+describe('ConfigureCard component', () => {
+  beforeEach(fetchMock.mockReset);
+
   describe('makes an API call', () => {
     test('when the user clicks Fetch Schedules', () => {
+      let fetchCount = 0;
+
+      fetchMock.mockImplementation((route: string): Promise<Response> => {
```
firejake308, line 29: If there are no other calls to `fetch` in this test, we might be able to `expect(fetchMock).toBeCalledWith('/scheduler/generate')`

autoscheduler/scheduler/views.py
```diff
@@ -18,10 +18,10 @@ def _parse_course_filter(course) -> CourseFilter:
     subject = course["subject"]
     course_num = course["courseNum"]
 
-    sections = course["sections"]
+    sections = course.get("sections", [])
 
-    honors = BasicFilter(course["honors"])
-    web = BasicFilter(course["web"])
+    honors = BasicFilter(course.get("honors"))
```
firejake308, line 23: Is `.get()` more reliable, or are you just using it to provide a default value?

autoscheduler/frontend/src/tests/ui/ConfigureCard.test.tsx
```diff
@@ -26,7 +49,74 @@ describe('OptionsCard component', () => {
...
+
+      // act
+      fireEvent.click(getByText('Generate Schedules'));
+
+      const { body } = fetchMock.mock.calls[0][1]; // Body is returned as a "blob"
+      // Convert the body into a string, parse it into an object, then get the sections field
+      const { sections } = JSON.parse(body.toString());
+
+      // assert
+      expect(sections).toBeUndefined();
```
firejake308, line 119: Based on my reading of `ConfigureCard.tsx`, shouldn't `sections` be an empty array rather than undefined?

autoscheduler/scheduler/views.py
gannonprudhomme: So the `[key]` accessing will throw an error if the key doesn't exist in the data, whereas `.get()` returns `None` if it doesn't exist. This way if `honors`/`web` isn't in the sent data it will still return stuff. This case shouldn't ever actually occur, but it's probably a better practice to use `.get()`

autoscheduler/frontend/src/redux/actions/courseCards.ts
gannonprudhomme: Yeah exactly. They're only very slightly different and since they're both needed for different situations, I thought extracting out the common functionality was the best move

autoscheduler/frontend/src/tests/ui/ConfigureCard.test.tsx
gannonprudhomme: Yeah that's a great point. I think I ran into trouble when I tried to make it an empty array that it would be `undefined` anyways, but I'll double check and get back to you

gannonprudhomme:
- Rebased with `master` and resolved conflicts
- Squashed the `parseAllMeetings` and `parseMeeting` docstring updates and the `sectionDAta` type into `Refactored parseSections and added parseAllMeetings`
- Renamed `parseSection` to `parseSectionSelected`
- Added loading indicator mocking & removed unused mock

I'll look into the `undefined` section assertions then will re-request review afterwards

autoscheduler/frontend/src/tests/ui/ConfigureCard.test.tsx
gannonprudhomme: So I thought this would work too, but since it's a POST request with a JSON body, we'd also have to assert the body of it:

```
 FAIL  tests/ui/ConfigureCard.test.tsx (6.81s)
  ● ConfigureCard component › makes an API call › when the user clicks Fetch Schedules

    expect(jest.fn()).toHaveBeenCalledWith(...expected)

    Expected: "scheduler/generate"
    Received: "scheduler/generate", {"body": "{\"term\":\"\",\"includeFull\":false,\"courses\":[],\"availabilities\":[]}", "method": "POST"}

    Number of calls: 1
```

firejake308: In that case, try `expect(fetchMock).toBeCalledWith('/scheduler/generate', expect.any())`

gannonprudhomme:
Okay so turns out I had my test set up like completely wrong. Will try to get them fixed before the meeting tomorrow, but I think I'm going to go insane if I work on them for any longer right now

autoscheduler/frontend/src/timeUtil.ts
```diff
@@ -8,7 +8,13 @@ export const LAST_HOUR = 21;
...
-  return `${use24Hour ? h : (h - 1) % 12 + 1}:${
+export function formatTime(h: number, m: number, use24Hour = false, padZeroes = false): string {
+  const ret = `${use24Hour ? h : (h - 1) % 12 + 1}:${
     new Intl.NumberFormat('en-US', { minimumIntegerDigits: 2 }).format(m)}`;
+
+  if (padZeroes && ret.length < 4) {
+    return `0${ret}`;
+  }
+
+  return ret;
```
ryan-conn, line 19: This function is a bit messy right now, if you reuse the NumberFormat for hours it's more straightforward:
```suggestion
  const formattedHours = use24Hour ? h : (h - 1) % 12 + 1;
  const padZero = new Intl.NumberFormat('en-US', { minimumIntegerDigits: 2 });
  return `${padZeroes ? padZero.format(formattedHours) : formattedHours}:${padZero.format(m)}`;
```

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -36,57 +36,99 @@ function updateCourseCardSync(index: number, courseCard: CourseCardOptions): Upd
...
+      endTimeMinutes: Number(end[1]),
+      // Converts string to MeetingType enum
+      meetingType: MeetingType[meetingData.type as keyof typeof MeetingType],
+      section,
+    });
+
+    meetings.push(meeting);
+  });
+
+  return meetings;
```
ryan-conn, line 96: Since all the `forEach` is doing is appending a single item to an array, you can use `sectionData.meetings.map()` instead

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -36,57 +36,99 @@ function updateCourseCardSync(index: number, courseCard: CourseCardOptions): Upd
...
 /**
  *  Helper function to parse the serialized Section model from the backend convert it into an
  *  array of Sections
  *  @param arr The array of sections returned from the backend, such as from api/sections
  */
-export function parseSections(arr: any[]): SectionSelected[] {
+export function parseSectionSelected(arr: any[]): SectionSelected[] {
   const sections: SectionSelected[] = [];
 
   arr.forEach((sectionData) => {
```
ryan-conn, line 127: Change this to use `arr.map()`

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -35,39 +36,58 @@ const ConfigureCard: React.FC = () => {
     const courses = [];
     for (let i = 0; i < courseCards.numCardsCreated; i++) {
       if (courseCards[i] && courseCards[i].course) {
-        const [subject, courseNum] = courseCards[i].course.split(' ');
+        const courseCard = courseCards[i];
+
+        const selectedSections: string[] = []; // the section nums of the selected sections
+
+        // Iterate through the sections and only choose the ones that are selected
+        courseCard.sections.forEach((sectionSel) => {
```
firejake308, line 44: This could probably be implemented as a `.filter()`

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -35,39 +36,58 @@ const ConfigureCard: React.FC = () => {
...
+
         courses.push({
           subject,
           courseNum,
-          sections: courseCards[i].sections,
-          honors: courseCards[i].honors,
-          web: courseCards[i].web,
+          sections: isBasic ? [] : selectedSections, // Only send if "Section" customization level
+          // Only send if "Basic" level
+          honors: isBasic ? courseCard.honors ?? filterDefault : filterDefault,
```
firejake308, line 61: Can I get some parentheses around this to make it easier to read? Side note, nice use of the new `??` operator

gannonprudhomme:
> The first time I ran it, I forgot to check "include full sections" since most sections are full now, so I'd like to have an error message when there are no schedules. We could consider a [snackbar](https://material-ui.com/components/snackbars/), a dialog, or just a message in the `SchedulePreview`, but something owuld be nice.

Yeah good call, I think a snackbar would work well. Would rather do this in a separate PR though honestly, as I just want this merged.

> Also, this would be a backend change I guess, but maybe the API route should have a query/body parameter for the number of schedules to return in case we want more than 5 later?

It's already an option for the actual `create_schedules` function so this would be really easy. Would like to do this in a separate PR though whenever we actually want to, as again it'd be a really small change. We can discuss during the meeting though

PR #262, opened 2020-05-20
Meeting times are no longer covered up by adjacent meetings

### LINKED ISSUE #219: Overlapping Meeting Times
When one meeting ends at the same time as another, the end time for the first is covered up by the second. See the picture below:
image

firejake308:
Before:
image
After:
image

I don't know if this is something that Jest can test, since it pretty much requires looking at the screen. I also wasn't sure if JSDom would mock Z-indices, but I could try comparing those if y'all think it would be worth it.

Closes #219

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/Schedule/ScheduleCard/ScheduleCard.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/Schedule/ScheduleCard/ScheduleCard.css
@@ -14,6 +14,10 @@
     z-index: 3;
 }
 
+.schedule-card:hover {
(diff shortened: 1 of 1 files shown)


gannonprudhomme:
Do you have any example courses we can test this with?

firejake308:
@gannonprudhomme I discovered the bug using test data, but it looks like you figured it out since you approved it. Did you want me to do anything else before merging?

gannonprudhomme:
> @gannonprudhomme I discovered the bug using test data, but it looks like you figured it out since you approved it. Did you want me to do anything else before merging?

Nope, all good!

PR #270, opened 2020-05-31
Fixed bug when dragging outside of schedule component

### LINKED ISSUE #268: Availability blocks change start time to 0:00 when dragging outside of schedule component
When making an availability block, dragging out of the left/right side of the Schedule component causes the start time to reset to 0:00.
leave schedule bug
This is due to `handleMouseLeaveCalendarBounds` sets `time2` to `null` by default (you can verify this by changing it to something else), although I'm not sure what the best way to fix it is.

firejake308:
Added a test to replicate and a null-check to fix. Closes #268


firejake308:
I changed it to have the functionality we discussed at our 5/31 meeting, where the drag continues if the mouse leaves the calendar bounds to the side

PR #273, opened 2020-06-06
Frontend/concat exams

### LINKED ISSUE #256: Concatenate duplicate EXAM meeting times in SectionSelect
We need to combine the duplicate EXAM meeting times in the SectionSelect so that they're only one entry.

I.e. for CSCE 121 - term 202031:

firejake308:
## Description

Some classes have multiple meetings listed for common exams, but each of these meetings meet on the same day of the week, at the same time, and have the same meeting type. This PR filters out those duplicates and condenses them into one meeting in `SectionSelect`.

## Rationale

At the moment, I have elected to also condense meetings that start and end at the same hour, even if the minutes are different. However, this change is implemented in `SectionSelect`, so in the `Schedule`, we'll still see overlapping exam times. I think I'll probably change that as this PR progresses.

Furthermore, the current changes affect all meeting types, not just exams. This is probably desired, even though we have no use cases at the moment.

Finally, exams that meet on different days but the same time will currently not be concatenated. This can be changed if desired.

## How to test

Tests have been implemented for this feature. To see the change in the UI, check out PHYS 207 - 501, term 202031

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #256

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -58,16 +58,35 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
   /**
    * Accepts an array of meetings and returns a filtered array without duplicate meetings.
    * Meetings are considered to be duplicates if they are of the same type, meet on the same days,
-   * and start and end at the same hour. Meetings that are the same by all of these criteria but
-   * differ only in the exact minutes of the meeting will still be considered duplicates
+   * and start at the same time. Meetings that are the same by all of these criteria but
+   * differ only in the end times will still be considered duplicates
    * @param arr
    */
   const filterDuplicateMeetings = (arr: Meeting[]): Meeting[] => {
+    // helper function to merge two meetings
+    const mergeMeetings = (mtg1: Meeting, mtg2: Meeting): Meeting => {
+      if (!mtg2) return mtg1;
+
+      // choose the later end time
+      const [laterEndHours, laterEndMinutes] = mtg2.endTimeHours > mtg1.endTimeHours
+        ? [mtg2.endTimeHours, mtg2.endTimeMinutes]
+        : [mtg1.endTimeHours, mtg1.endTimeMinutes];
+      // merge the days array by logical OR of each element
+      const days = mtg1.meetingDays.map((hasMeeting, idx) => hasMeeting || mtg2.meetingDays[idx]);
+      return {
+        ...mtg1,
+        endTimeHours: laterEndHours,
+        endTimeMinutes: laterEndMinutes,
+        meetingDays: days,
+      };
+    };
+
     // add all meetings to a map, then get the values of the map
     const map = new Map<string, Meeting>();
-    arr.forEach((mtg) => map.set(
-      `${mtg.meetingType}${mtg.meetingDays}${mtg.startTimeHours}${mtg.endTimeHours}`, mtg,
-    ));
+    arr.forEach((mtg) => {
+      const key = `${mtg.meetingType}${mtg.startTimeHours}${mtg.startTimeMinutes}`;
+      map.set(key, mergeMeetings(mtg, map.get(key)));
+    });
     return [...map.values()];
   };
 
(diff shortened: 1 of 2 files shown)


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -55,6 +55,22 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
...
+  /**
+   * Accepts an array of meetings and returns a filtered array without duplicate meetings.
+   * Meetings are considered to be duplicates if they are of the same type, meet on the same days,
+   * and start and end at the same hour. Meetings that are the same by all of these criteria but
+   * differ only in the exact minutes of the meeting will still be considered duplicates
+   * @param arr
+   */
+  const filterDuplicateMeetings = (arr: Meeting[]): Meeting[] => {
+    // add all meetings to a map, then get the values of the map
+    const map = new Map<string, Meeting>();
```
ryan-conn, line 67: This map could use a more descriptive name, like `uniqueMeetings`

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -55,6 +55,22 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
       - ${formatTime(mtg.endTimeHours, mtg.endTimeMinutes)}`;
   };
 
+  /**
+   * Accepts an array of meetings and returns a filtered array without duplicate meetings.
+   * Meetings are considered to be duplicates if they are of the same type, meet on the same days,
+   * and start and end at the same hour. Meetings that are the same by all of these criteria but
+   * differ only in the exact minutes of the meeting will still be considered duplicates
```
ryan-conn, line 62: I'm not sure exactly how I feel about this since it only depends on hours and will end up being inaccurate in some cases (for example if there are 2 exams, one is 8:00-9:30 and the other is 8:00-10:00, they won't be considered the same, if there is instead one from 8:00-9:30 and another from 8:00-9:00, the time will show as 8:00-9:00 despite one of the meetings going later), I think a better way of doing this (although it'll make your pretty clever code a little more verbose) would be making the start time in hours and minutes the key, and when going through meetings if the key already exists, update the end time to the later of the existing value and the meeting being added

firejake308:
> Other than my comments, here's something I noticed:
> 
> ... another way we can merge meetings is to merge meetings that are identical other than the day

That's actually one of the things I was considering, but I didn't want to make the changes without discussing first. It wouldn't change the displayed schedule at all, but it would help us condense the `CourseSelectCard`s, so I think it would be helpful

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
firejake308: Makes sense, but the algorithm you described would miss the case where one of the exams had a slightly different start time than the others, if it were ever to occur. That's fine by me, but I just wanted to double check before implementing it.

firejake308: As per meeting on 6/14, I have implemented the requested behavior

firejake308:
Yeah, I think this is good

gannonprudhomme:
@ryan-conn Can you review this when you get the chance? This looks good to me

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -55,6 +55,41 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
...
+        endTimeMinutes: laterEndMinutes,
+        meetingDays: days,
+      };
+    };
+
+    // add all meetings to a map, then get the values of the map
+    const uniqueMeetings = new Map<string, Meeting>();
+    arr.forEach((mtg) => {
+      const key = `${mtg.meetingType}${mtg.startTimeHours}${mtg.startTimeMinutes}`;
+      uniqueMeetings.set(key, mergeMeetings(mtg, uniqueMeetings.get(key)));
```
ryan-conn, line 88: This is a pretty clever way to avoid having to use `uniqueMeetings.has()` since it checks if mtg2 is falsy

PR #276, opened 2020-06-13
Frontend/save schedules

### LINKED ISSUE #228: Add save button to the generated schedules
Need to add a button on the schedules which would allow the users to save a schedule. Then anytime a generate schedules is pressed, it will still delete all of the old schedules, but keep the saved ones.

Also need to make a design for it

ryan-conn:
## Description

Adds functionality to save/unsave and remove schedules.

## Rationale

A couple of potentially controversial/questionable changes I made:
- Colors for save/delete buttons are the same because I wasn't sure what color if any to change them to, if anyone has suggestions then go ahead
- A couple tests are causing a minor error I'm not sure how to fix, I'll post a comment describing this in more detail

## How to test

Just generate schedules and save them, you can see that when you click generate schedules again it keeps your saved schedules.

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #228.


autoscheduler/frontend/src/tests/ui/SchedulePreview.test.tsx
```diff
@@ -145,4 +147,94 @@ describe('SchedulePreview component', () => {
...
+      const saveScheduleButton = (await findAllByText('Save'))[1];
+      fireEvent.click(saveScheduleButton);
+      await waitFor(() => getByText('Unsave'));
+      fireEvent.click(saveScheduleButton);
+
+      // assert
+      await waitFor(() => (
+        expect(store.getState().schedules.savedSchedules).toHaveLength(0)
+      ));
+    });
```
ryan-conn, line 238: These tests cause the following error:
image
This is because of the line `await waitFor(() => getByText('Unsave'));` (which I realize could just be `await findByText('Unsave')`, not sure why I didn't do that in hindsight but it still results in the same error). This error doesn't occur when actually using the UI so it's pretty minor and the tests still work correctly despite it but still, it would be nice to fix.

autoscheduler/frontend/src/redux/reducers/schedules.ts
```diff
@@ -22,18 +24,77 @@ export interface ReplaceSchedulesAction {
...
+export interface UnsaveScheduleAction {
+    type: 'UNSAVE_SCHEDULE';
+    index: number;
+}
+export type ScheduleAction = AddScheduleAction | RemoveScheduleAction
+  | ReplaceSchedulesAction | SaveScheduleAction | UnsaveScheduleAction;
+
+interface Schedules {
+  schedules: Meeting[][];
+  savedSchedules: Meeting[][];
```
firejake308, line 40: Doesn't this result in duplication of data? I'm thinking `savedSchedules` should be like an array of indices

autoscheduler/frontend/src/redux/reducers/schedules.ts
```diff
@@ -22,18 +24,77 @@ export interface ReplaceSchedulesAction {
...
+}
+
+const initialSchedules: Schedules = {
+  schedules: [],
```
firejake308, line 44: Maybe rename this to `allSchedules` or something to be more specific about what it's doing

autoscheduler/frontend/src/redux/reducers/schedules.ts
```diff
@@ -22,18 +24,77 @@ export interface ReplaceSchedulesAction {
...
+      return {
+        ...state,
+        savedSchedules: state.savedSchedules
+          .concat(containsSchedule(state.savedSchedules, state.schedules[action.index])
+            ? [] : [state.schedules[action.index]]),
+      };
+    case UNSAVE_SCHEDULE:
+      return {
+        ...state,
+        savedSchedules: state.savedSchedules.filter((schedule) => (
```
firejake308, line 92: Why can we use the built-in `filter` here if we have to use a custom `contains` function at line 86?

autoscheduler/frontend/src/tests/ui/SchedulePreview.test.tsx
firejake308: I would move the `sleep` function to `testUtil.ts` and mock it to return immediately during the test

autoscheduler/frontend/src/redux/reducers/schedules.ts
ryan-conn: The custom `contains` function makes it so if 2 schedules are generated with the same classes in them, only one of them will be saved instead of having 2 identical schedules (which should be a pretty rare edge case), whereas elements in `savedSchedules` are references to elements in `allSchedules`, so checking shallow equality will work

autoscheduler/frontend/src/redux/reducers/schedules.ts
ryan-conn: No, objects in arrays are references to the objects, not the actual data so storing it as a number or a reference take up the same memory, as both are 8 bytes

autoscheduler/frontend/src/redux/reducers/schedules.ts
ryan-conn: Good idea, referencing `schedules.schedules` is a little confusing

ryan-conn:
Now up to date with master, I also changed the UI to the new design that Adel made + addressed his review comments

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -83,7 +87,7 @@ const SchedulePreview: React.FC = () => {
         <div id={styles.cardHeader}>Schedules</div>
       }
     >
-      <List className={styles.list}>
+      <List className={styles.list} disablePadding>
```
ryan-conn, line 90: Not really related to this PR but the padding on the top and bottom of the list bugged me so I changed it, here's before:
image
and after:
image
Let me know if you guys don't like this change, but I think it looks nicer

autoscheduler/frontend/src/components/SmallFastProgress.tsx
```diff
@@ -1,8 +1,18 @@
 import * as React from 'react';
 import { CircularProgress } from '@material-ui/core';
 
-const SmallFastProgress: React.FC = () => (
-  <CircularProgress size={24} style={{ animationDuration: '550ms' }} disableShrink />
-);
+interface SmallFastProgressProps {
```
gannonprudhomme, line 4: Since this is just going to be a POST/PUT request and the user won't be receiving any data from it (and as such nothing in the view will actually change), I don't think this needs a progress indicator at all & I think it's safe to assume that the request we do worked successfully. Worst case scenario we can show a snackbar in case of an error (although I don't think it's necessary)

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/Buttons/SaveScheduleButton.tsx
```diff
@@ -0,0 +1,70 @@
...
+  let tooltipText = saved ? 'Unsave' : 'Save';
+
+  if (loading) {
+    icon = <SmallFastProgress size="small" />;
+    tooltipText = saved ? 'Saving...' : 'Unsaving...';
+  }
+
+  return (
+    <Tooltip title={tooltipText} placement="top">
+      <IconButton className={styles.scheduleButton} size="small" onClick={handleClick} data-testid="save-schedule">
```
gannonprudhomme, line 63: Ik `small` is the smallest size MUI provides, but I think the lock & delete icons are still a bit too large - I think they should be shorter than or equal to the schedule text itself

autoscheduler/frontend/src/components/SmallFastProgress.tsx
ryan-conn: I disagree (at least about removing it entirely), since the most useful part of saving schedules is that after refreshing they'll still be there we want to make sure request to save actually goes through. Imagine that you click save on your perfect schedule, see "saved", refresh, and then it disappears when you refresh. Also because the overhead on the save/remove schedule API calls will probably be very small, I think it's acceptable to have the loading indicator

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/Buttons/SaveScheduleButton.tsx
ryan-conn: So apparently icons have a weird behavior where the icon size isn't actually changed by putting `size="small"` on the `IconButton` unless you also put `size="inherit"` on the icon itself, which I didn't notice until you pointed this out. Here's what it looks like now:
image
And here's what it looks like with `size="inherit"` on the buttons:
image
Let me know which you like better, or any suggestions for other options

ryan-conn:
> ...my main thing that I think is weird about this is the deleting of schedules & how we handle the titles of the schedules. If you delete `Schedule 1`, `Schedule 2` becomes `Schedule 1`, which frankly is pretty confusing. Not sure the best way to deal with this as we might have to completely change how we handle the schedules, but I think if possible we should maintain the indices the schedules received whenever they were generated

I think if we do this, we should just give users the option to change the name of their schedules manually (which shouldn't be too much work, we should be able to just add a `name` prop to schedule and change the schedule name with whatever is in that field).

We should also put that functionality into a separate PR if we decide to implement that functionality, imo it would be a pretty useful feature so it's worth implementing.

Thoughts? @gannonprudhomme @firejake308 @ThatJuanGuy

ThatJuanGuy:
I think changing the names of schedules would be a pretty useful feature to have as well. Regarding the indices, I agree with not renaming the remaining schedules when deleting one. It would be pretty confusing, especially if multiple schedules are already saved and they all get renamed.

autoscheduler/frontend/src/tests/ui/SchedulePreview.test.tsx
ryan-conn: I just removed the sleep function since everyone's seen how it looks and we won't want it in master anyways

ryan-conn:
Merged in changes from master since we're going to squash and merge PRs now anyways, and addressed feedback so this should be ready for another round of review.

ryan-conn:
I removed the loading indicator like we discussed in the previous meeting and made this branch up-to-date with master

PR #281, opened 2020-06-27
Shows snackbar when no schedules can be generated

### LINKED ISSUE #278: Add snackbar when no schedules are generated to frontend
Per Adel's comment [here](https://github.com/aggie-coding-club/Rev-Registration/pull/255#pullrequestreview-430249486), we should add a snackbar popup whenever no schedules are generated (when `/scheduler/generate` returns an empty array)

firejake308:
## Description

I often hit "Generate Schedules" and then sit for a very confusing minute wondering why it's taking so long, until I realize that I didn't check "Include full sections". Users will almost certainly make similar mistakes, so we need an error handling method to let them know that they need to change their filters. To this end, I have added a snackbar that shows when the Generate Schedules  button returns no schedules.

***tests coming soon***

## Rationale

I think the only possible controversial thing here would be the text in the error bar. It currently reads "No schedules found. Try widening your criteria." Let me know if you'd like something else.

## How to test

Hitting Generate Schedules without adding any courses is the easiest way to test this.

## Screenshots

image

## Related tasks

Closes #278

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import {
-  Button, Checkbox, ListItem, ListItemIcon, ListItemText,
+  Button, Checkbox, ListItem, ListItemIcon, ListItemText, Snackbar, IconButton,
 } from '@material-ui/core';
+import CloseIcon from '@material-ui/icons/Close';
 import { useDispatch, useSelector } from 'react-redux';
 import GenericCard from '../../GenericCard/GenericCard';
 import SmallFastProgress from '../../SmallFastProgress';
@@ -23,10 +24,16 @@ import { formatTime } from '../../../timeUtil';
 const ConfigureCard: React.FC = () => {
   const [includeFull, setIncludeFull] = React.useState(false);
   const [loading, setLoading] = React.useState(false);
+  const [showSnackbar, setShowSnackbar] = React.useState(false);
   const courseCards = useSelector<RootState, CourseCardArray>((state) => state.courseCards);
   const term = useSelector<RootState, string>((state) => state.term);
   const avsList = useSelector<RootState, Availability[]>((state) => state.availability);
   const dispatch = useDispatch();
+
+  const checkIfEmpty = (schedules: Meeting[][]): Meeting[][] => {
+    if (schedules.length === 0) setShowSnackbar(true);
+    return schedules;
+  };
 
   const fetchSchedules = React.useCallback(() => {
     // show loading indicator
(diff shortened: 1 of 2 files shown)


autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -117,6 +129,17 @@ const ConfigureCard: React.FC = () => {
             : 'Generate Schedules'}
         </Button>
       </div>
+      <Snackbar
+        open={showSnackbar}
+        autoHideDuration={10000}
+        message="No schedules found. Try widening your criteria."
+        onClose={(): void => setShowSnackbar(false)}
```
ryan-conn, line 136: Might want to change this callback to the following:
```ts
  const handleSnackbarClose = (_event: any, reason: string): void => {
    if (reason === 'clickaway') return;
    setShowSnackbar(false);
  };
```
Right now the snackbar closes if you click anywhere outside of it, which makes it easy to ignore if you're not paying close attention

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -117,6 +129,17 @@ const ConfigureCard: React.FC = () => {
             : 'Generate Schedules'}
         </Button>
       </div>
+      <Snackbar
+        open={showSnackbar}
+        autoHideDuration={10000}
```
ryan-conn, line 134: Like Gannon said in the meeting, we'll probably want to change this to ~5-6 seconds

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
firejake308: Hmm, I know this was the recommended behavior in the docs, but I kinda liked having it go away if you clicked outside. Can I get a 3rd opinion from @gannonprudhomme or @ThatJuanGuy ?

ThatJuanGuy: I pulled the branch but I can't really check how obvious the snack bar is when it pops up because I still generate the same 3 preset schedules regardless of what courses I have selected. Regardless, I don't think it should disappear if you click outside of it. This is because if no schedules are showing up the user might be randomly clicking to pass the time or repeatedly clicking the generate button to try and make it go faster. Both of which would make it much easier for them to miss the message letting them know no schedules can be generated with their current settings.

autoscheduler/frontend/src/tests/ui/ConfigureCard.test.tsx
```diff
@@ -153,4 +153,47 @@ describe('ConfigureCard component', () => {
       expect(sections.length).toEqual(0);
     });
   });
+
+  describe('shows an error message', () => {
```
gannonprudhomme, line 157: nit: This should probably be `shows an error snackbar`. Same with the `does not show an error message`.

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -89,7 +107,7 @@ const ConfigureCard: React.FC = () => {
         <div id={styles.cardHeader}>Configure</div>
       }
     >
-      <div className={styles.buttonContainer}>
+      <div className={styles.buttonContainer} ref={ref}>
```
gannonprudhomme, line 110: Just for my understanding, what does adding this ref do for us?

firejake308: I'll add a comment above the initial `useRef` because you're right that it's not immediately clear what its purpose is. Basically, when I was running tests, I kept getting an error about React dispatching a `setState` in an async callback after un-mounting a component. My solution to this was to obtain a reference to the DOM element and check if it's still mounted before calling any asynchronous `setState()`s.

TL;DR the ref fixes async setState

PR #288, opened 2020-06-29
Google social auth now works

### LINKED ISSUE #156: Add Google Authentication
This will most likely involve adding some form of Student/User model, as we'll most likely need to save their access & refresh token that is returned from OAuth

This is both a frontend and backend task since this might as well including adding the login button on the frontend (since it's not that involved)

ThatJuanGuy:
## Description

This pull request allows for google authentication when the google login button is pressed in the top right corner of the nav bar. In order for this to work without leaking our client and secret key, y'all are going to have to create a .env file. The instructions are as follows.
1) Navigate to the settings directory inside autoscheduler
2) Create a new file called ".env"
3) Within the .env file, type/copy paste the following.
`GOOGLE_OAUTH2_CLIENT_ID =`
`GOOGLE_OAUTH2_SECRET = `
This picture shows the contents of the .env file.
tempsnip
4) Set the client_id and secret values equal to what is in the discord using single quotes.
ex) `GOOGLE_OAUTH2_CLIENT_ID = 'MAGICVALUE'`
5) Once both values are set, save the changes to your .env file. It should now be properly set up.
## How to test

To test, first run `python manage.py migrate` because the social auth app needs to create some tables. After this is done, click the login button and see if the google login page pops up. Login in with a valid google account. If the authentication is successful, you should be redirected back to the homepage. Any other behavior likely means this isn't working right.

## Screenshots

This is what should appear after pressing the google login button.

googleauth page

## Related tasks

Closes #156


autoscheduler/requirements.txt
```diff
@@ -13,3 +13,4 @@ pytest-django==3.6.0
 PyPDF2==1.26.0
 requests==2.21.0
 requests-mock==1.7.0
+social-auth-app-django==3.1.0
```
gannonprudhomme, line 16: You need to add `dotenv` here, which is why the tests are failing (at least probably)

ThatJuanGuy: Do I add the instructions for the .env file to the read me or just this PR?

gannonprudhomme: I think just the PR is fine

gannonprudhomme:
Ack I resolved the conflicts online which dismissed my approval and made that merge commit - but I'm going to be squashing this due to that `temp` commit so it shouldn't matter

PR #290, opened 2020-07-08
Got rid of getOptionSelected warning in tests

firejake308:
## Description

So I fixed the annoying warning we got with the `CourseSelectCard` tests about no option being selected. Basically , I told it to select the first option. In the browser, however, this only works after you type at least two letters, which is good enough for me.
¯\\\_(ツ)\_/¯

## Rationale

It's like a one-line change, just read the code.

I did also add some underscores for unused arguments on the functions below it, just for better linting

## How to test

Run `npm run test`. You hsould get no yellow warnings. There are still some red warnings about an update not wrapped inside `act`, but I believe I fixed those in #278 

## Screenshots

N/A

## Related tasks

I acutally don't think we ever made one for this

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
@@ -115,16 +115,17 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
           value={course}
           multiple={false}
           filterOptions={(): any[] => options} // Options are not filtered
+          getOptionSelected={(option): boolean => option === options[0]}
           onClose={(): void => {
             if (!options.find((val) => val === inputValue)) setInputValue('');
           }}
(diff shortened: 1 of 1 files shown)


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
```diff
@@ -115,16 +115,17 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
           value={course}
           multiple={false}
           filterOptions={(): any[] => options} // Options are not filtered
+          getOptionSelected={(option): boolean => option === options[0]}
```
ryan-conn, line 118:
```suggestion
          getOptionSelected={() => true}
```
Shouldn't this do the same thing

firejake308: Yeah, you're right. I thought it called this function on all of the options, but it turns out it only applies it to the current value of the Autocomplete. Fixed as of 7/25

Issue #260, opened 2020-05-20
Add sessions functionality


gannonprudhomme:
This will entail:

- Creating the `user_sessions` django app
- Storing the sessions on `scheduler/generate` calls
- Retrieving the sessions & setting the data on the frontend for:
  - Navigating to /schedule from landing page
  - Refreshing when on the /schedule page

Need to figure out how to:
- Connect a session to a user

The route for saving schedules will be added in a later issue, but this should also provide the groundwork for it

gannonprudhomme:
We already have separate issues for saved schedules (#267) and availabilities (#316), so I'm going to close this.

PR #292, opened 2020-07-12
Store courses in sessions

ryan-conn:
## Description

Saves courses when generate schedules is clicked, and sets courses on loading the page or choosing a term.

## Rationale

I had to change the `courseCards` redux pretty substantially to make everything work, so if there are any changes you guys want to suggest then I'm open to it. The main thing that's a bit strange with it is that `replaceCourseCards` doesn't actually create an action, it just dispatches several different actions depending on the arguments. The reason for this is so that all the intermediate states can be shown instead of having to calculate everything and then updating state, but I can change it to do everything simultaneously.

I also changed the formatting of the `sessions/get_last_term` API to return empty json instead of an empty string if a last term doesn't exist since it's simpler, didn't require any frontend changes, and is consistent with how the `get_saved_courses` API works.

Last thing is that I fetch updated sections and update the saved course cards since the seats/available sections, etc might have changed since the schedule was generated, I tried to keep as much information as possible from the saved schedules though. It might be worthwhile to simplify how schedules are represented on the backend (instead of serializing the entire `courseCards` redux slice, only store certain parts to reduce storage requirements) because not all of the information is used, but for now I'm keeping it simple. 

## How to test

Do whatever you want with course cards, click "generate schedules", and refresh/change terms, etc., and the state of your course cards should be exactly the same.

## Related tasks

Related to #260, still need to save availabilities but courses are pretty complicated as it involves fetching/modifying data rather than direct serialization/deserialization so it's not happening in this PR


ryan-conn:
Should be ready for review now, tests have been created

I also noticed that github actions wasn't running tests from the `user_sessions` folder, so I removed `testpaths` from our `pytest.ini` because it can find all tests ending in `test[s].py` by itself, I opted to do that instead of adding the `user_sessions/tests` folder to prevent this from happening again if we add another folder in the future

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -212,3 +234,71 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
...
+    for (let i = 0; i < courseCards.numCardsCreated; i++) {
+      const card = courseCards[i];
+      if (card) cards.push(card);
+    }
+
+    // create promise for each updated course card
+    const newCourseCards = cards.map((card) => (updateSectionsForCourseCard(card, term)));
+
+    // clear all course cards
+    new Promise((resolve) => { dispatch(clearCourseCards()); resolve(); }).then(async () => {
```
firejake308, line 290: Do you need to call `resolve` manually here?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -212,3 +234,71 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
...
+      // add empty course card if there are no course cards for the current term
+      if (cards.length === 0) {
+        dispatch(addCourseCard());
+      } else {
+        // update information from all course cards and add them in order
+        for (let i = 0; i < newCourseCards.length; i++) {
+          // eslint-disable-next-line no-await-in-loop
+          const newCourseCard = await newCourseCards[i];
```
firejake308, line 298: Why do we need an `await` here? Array access isn't asynchronous, is it?

autoscheduler/frontend/src/redux/reducers/courseCards.ts
```diff
@@ -61,6 +63,8 @@ export default function courseCards(
         ...state,
         [action.index]: { ...state[action.index], ...action.courseCard },
       };
+    case CLEAR_COURSE_CARDS:
+      return { numCardsCreated: 0 };
```
firejake308, line 67: This might need a `...state`, let me see if I can check it

ryan-conn:
> As far as functionality is concerned, does this only save when you press Generate Schedules?

Yeah for now that's the only time they're sent, but during the meeting I was planning on talking about what other situations we'll want to save courses

autoscheduler/frontend/src/redux/reducers/courseCards.ts
firejake308: Ok nvm it's fine

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: Yes, if you don't do it manually then it doesn't wait for the dispatch to complete

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: We want the courses to be in the same order as they're stored in when they're recreated, so here it has to wait for the previous card to be generated before dispatching to add the next card

autoscheduler/frontend/src/redux/actions/courseCards.ts
gannonprudhomme: In that case, I would mention that in a comment as it's not very intuitive to me

ryan-conn: I changed the name of the array to `courseCardPromises` after Adel's comment, do you still think the comment is necessary? Also since I refactored this code to call `.then()` on each of them I thought it would be clearer

autoscheduler/frontend/src/tests/ui/CourseSelectColumn.test.tsx
```diff
@@ -19,16 +23,20 @@ function ignoreInvisible(content: string, element: HTMLElement, query: string |
...
 
+      // sessions/get_saved_courses
+      fetchMock.mockResponseOnce(JSON.stringify([]));
+
       const { getByText, getAllByText } = render(
         <Provider store={store}>
           <CourseSelectColumn />
         </Provider>,
       );
+      await new Promise(setImmediate);
```
gannonprudhomme, line 39: What does this do? And why do we need it

autoscheduler/user_sessions/tests/courses_api_tests.py
```diff
@@ -0,0 +1,108 @@
...
+        session = self.client.session
+
+        # Assert
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(expected, session[term]['courses'])
+
+    def test_save_courses_correctly_parses_course_data_json(self):
+        """ Tests that /sessions/save_courses can parse a complicated json object.
+            Note that this doesn't test anything related to the actual formatting:
+            it should simply trust that the frontend gives it data that can it can
```
gannonprudhomme, line 33:
```suggestion
            it should simply trust that the frontend gives it data that it can
```

autoscheduler/user_sessions/tests/courses_api_tests.py
```diff
@@ -0,0 +1,108 @@
...
+        """ Tests that /sessions/get_saved_courses doesn't allow requests with no term
+        """
+        # Arrange
+        term = ''
+
+        # Act
+        response = self.client.get(f'/sessions/get_saved_courses?term={term}')
+
+        # Assert
+        self.assertNotEqual(response.status_code, 200)
```
gannonprudhomme, line 68: Do you think asserting that it's not 200 is better than asserting that it is 400?

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -83,6 +84,29 @@ const ConfigureCard: React.FC = () => {
...
     });
+
+    // make request to save course cards
+    const courseData = [];
+    for (let i = 0; i < courseCards.numCardsCreated; i++) {
+      const course = courseCards[i];
+      if (course) {
+        const sections = course.sections.filter(({ selected }) => selected).map((section) => (
+          section.section.id
+        ));
```
gannonprudhomme, line 95:
```suggestion
        const sections = course.sections.filter(({ selected }) => selected).map((sectionSel) => (
          sectionSel.section.id
        ));
```

nit: Frankly I always get confused by `SectionSelected`'s, so I would prefer specifying that its not just a normal `Section` object (the `section.section` looks pretty weird too lol)

autoscheduler/user_sessions/views.py
```diff
@@ -1,22 +1,61 @@
...
     # empty string term is valid (used to unset term), so explicitly check for None
     if term is None:
         return Response(status=400)
     request.session['term'] = term
     return Response()
+
+@api_view(['PUT'])
+@parser_classes([JSONParser])
+def save_courses(request):
+    """ API endpoint that saves course cards for user's current term. """
```
gannonprudhomme, line 30: I think it would be worth mentioning that this/both route(s) can store any format of data, since really the only thing related to courses is that it saves the data with the `courses` key in the session.

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -274,4 +284,109 @@ describe('Course Cards Redux', () => {
       });
     });
   });
+
+  describe('deletes all course cards and resets numCardsCreated', () => {
+    test('when clearCourseCards is dispatched', () => {
```
gannonprudhomme, line 289:
```suggestion
  describe('clearCourseCards', () => {
    test('deletes all course cards and resets numCardsCreated, () => {
```

So while this does sort of invert our typical test structure, I think I prefer having like the "category" as the out-most `describe` as it groups all the test for that action under the same block (i.e. it specifies the system under test more directly). Also similar to how the first word/phase after `test_` in our Python tests are the system under test. Same thing with the rests of the tests

autoscheduler/frontend/src/redux/actions/courseCards.ts
gannonprudhomme: I think really what isn't clear to me is that you have to call `resolve()` to wait for the `dispatch(clearCourseCards())` to finish (still don't understand that tbh) - the `courseCardPromises` itself doesn't confuse me as much

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -153,6 +171,33 @@ function sortSections(sections: SectionSelected[]): SectionSelected[] {
   return [].concat(...sectionsForProfs.values(), ...TBASections);
 }
 
+/**
+ * Fetches sections for course in courseCard, then updates courseCard with new sections
+ * @param courseCard course to get sections for
+ * @param term term to fetch sections for
+ */
+async function createCourseCardFrom(
```
gannonprudhomme, line 179: I would change the name of this to mention that it's doing a fetch (e.g. add a `fetch` or `retrieve` to the name)

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -274,4 +284,109 @@ describe('Course Cards Redux', () => {
...
+
+      // act
+      store.dispatch(clearCourseCards());
+
+      // assert
+      expect(store.getState().courseCards).toEqual({ numCardsCreated: 0 });
+    });
+  });
+
+  describe('replaces all course cards and keeps attributes not related to sections', () => {
```
gannonprudhomme, line 301: What does "keep attributes" mean here?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -153,6 +171,33 @@ function sortSections(sections: SectionSelected[]): SectionSelected[] {
...
+  courseCard: CourseCardOptions,
+  term: string,
+): Promise<CourseCardOptions> {
+  // const { course } = courseCard;
```
gannonprudhomme, line 183: Remove this commented out line

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -163,31 +208,12 @@ function updateCourseCardAsync(
...
-      )
-      .catch((error) => {
-        // eslint-disable-next-line no-console
-        console.log(error);
-      });
+    createCourseCardFrom(courseCard, term).then((updatedCourseCard) => {
+      dispatch(updateCourseCardSync(index, updatedCourseCard));
+    }).catch((error) => {
+      // eslint-disable-next-line no-console
+      console.log(error);
```
gannonprudhomme, line 215: I'm fine leaving this for now, but tbh we probably _shouldn't_ be outputting to the console - at least not in production. We can always add replace this with actual logging that would send this error somewhere we could view though (i.e. for crash analytics) / use a logging library so when we're debugging we can have it print to the console, but hide it when in prod.

autoscheduler/frontend/src/tests/ui/CourseSelectColumn.test.tsx
ryan-conn: This waits for all pending promises to be resolved, since we have to wait for all of the actions called by `replaceCourseCards` to finish if this isn't here then the lines after the dispatch will run before it's completed

autoscheduler/user_sessions/tests/courses_api_tests.py
ryan-conn: I might as well do it that way since it's more specific, when I made this test though I was thinking just to test that it doesn't give a successful response

autoscheduler/user_sessions/views.py
ryan-conn: Yeah I put a comment about it in a test instead of the actual API for some reason, will change

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
ryan-conn: This should probably say "properties", I mean all of the data for the card

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: If you look at the diff I just moved this into a function so this console log was there before, I agree but I didn't want to change what was already there. Should I just remove it?

autoscheduler/frontend/src/tests/ui/CourseSelectColumn.test.tsx
ryan-conn: Actually I don't think it's needed, I've been testing without it and it seems to work fine. Since the action is synchronous it should have to wait until after it finishes to continue

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
ryan-conn: I refactored this so it's no longer relevant

ryan-conn:
> Are the selected sections intentionally not being saved / reloaded? It remembers the customization level we had selected, but not the actual sections I selected.
This is unintentional, when I serialized course data I forgot to change the way I calculate what sections are checked (instead of an array it was still expecting `CourseCardOptions`) and didn't update the tests.

To fix this in the future, I made an interface called `SerializedCourseCard` that is used when creating the serialized data and when creating course cards from it so they should be guaranteed to have the same data now

> Also, do we want to save whether a card was collapsed/expanded?

We can do this in a future PR, but since it would involve moving expanded/collapsed into redux instead of the `CourseSelectCard` state I don't want to mess with it now

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: Actually I don't think it's needed, I've been testing without it and it seems to work fine. Since the action is synchronous the function should have to wait until after it finishes to continue

ryan-conn:
I addressed all your comments that have a 👍 which is most of them, so this should be ready for another round of review

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -17,6 +17,22 @@ const CourseSelectColumn: React.FC = () => {
...
+  React.useEffect(() => {
+    if (!term) return;
+    fetch(`sessions/get_saved_courses?term=${term}`).then((res) => (
+      res.json()
+    )).then((courses: SerializedCourseCardOptions[]) => {
+      dispatch(replaceCourseCards(courses, term));
+    }).catch(() => {
+      // If unable to parse saved course cards or request fails, go back to default
+      // of one empty course card
+      dispatch(replaceCourseCards([], term));
```
firejake308, line 32: Is there a difference between this and `clearCourseCards()`?

autoscheduler/frontend/src/redux/actions/courseCards.ts
firejake308: I'm thinking, yeah, we can remove it for now. I only put it in there as like a "just in case" sorta thing, but we can figure out a better crash-handling system later

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -212,3 +238,66 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
...
+  term: string,
+): Promise<CourseCardOptions> {
+  return fetchCourseCardFrom(courseCard, term).then((newCourseCard) => {
+    // course is now updated with new sections, re-select sections from original courseCard
+
+
+    const selectedSections = new Set(courseCard.sections);
+
+    newCourseCard.sections.forEach((section) => {
+      // eslint-disable-next-line no-param-reassign
```
firejake308, line 262: Instead of disabling the rule, we should be able to just set the `props` option to false in the config, since we're reassigning a property of the param, not the param itself

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -212,3 +238,66 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
...
+    });
+
+    return newCourseCard;
+  });
+}
+
+/**
+ * Creates an updated course card from courseCard by making an API call for updated
+ * section information
+ * @param courseCard Course card to update
```
firejake308, line 273: Can you specify here that `courseCard` contains data saved from the last visit?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -212,3 +238,66 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
...
+ */
+export function replaceCourseCards(
+  courseCards: SerializedCourseCardOptions[],
+  term: string,
+): ThunkAction<void, RootState, undefined, CourseCardAction> {
+  // data for sections might have changed since last visit, so create a new course card
+  // for each one asynchronously and update them with data from courseCards
+  return (dispatch): void => {
+    // if courseCards is improperly formatted or no cards are saved, create default
+    if (!Array.isArray(courseCards) || courseCards.length === 0) {
```
firejake308, line 284: What else would `courseCards` be besides an array? Is this different from null-checking?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -212,3 +238,66 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
...
+      dispatch(addCourseCard());
+      return;
+    }
+
+    // create promise for each updated course card
+    const courseCardPromises = courseCards.map((card) => (updateSectionsForCourseCard(card, term)));
```
firejake308, line 290: I feel like you have one extra set of parentheses here

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -274,4 +284,94 @@ describe('Course Cards Redux', () => {
       });
     });
   });
+
+  describe('clearCourseCards', () => {
+    test('deletes all course cards and resets numCardsCreated', () => {
+      // arrange
```
firejake308, line 290: I mean, there was no data to clear in this case, so this test would've passed even without dispatching any actions. I'd probably add a card during the arrange phase just to make sure that clearing it actually does something.

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -274,4 +284,94 @@ describe('Course Cards Redux', () => {
...
+      // act
+      store.dispatch<any>(replaceCourseCards(courseCards, '202031'));
+      // wait for all actions to finish
+      await new Promise(setImmediate);
+
+      // assert
+      expect(store.getState().courseCards).toMatchObject(expectedCourseCards);
+    });
+
+    test('adds non-existing sections for course cards', async () => {
```
firejake308, line 331:
```suggestion
    test('adds non-pre-existing sections for course cards', async () => {
```

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -274,4 +284,94 @@ describe('Course Cards Redux', () => {
...
+    });
+
+    test('keeps selected sections from courseCards', async () => {
+      // arrange
+      const store = createStore(autoSchedulerReducer, applyMiddleware(thunk));
+      // section is the id of the section returned by testFetch, should be checked
+      const section = 830262;
+      fetchMock.mockImplementationOnce(testFetch);
+
+      const courseCards = [
```
firejake308, line 359: This is the serialized version, right?

ryan-conn: Yeah, I should declare this as `SerializedCourseCardOptions[]` to make it more clear

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: The default response is `{}` instead of `[]` since I didn't want to change the expected format, I could just change it to `[]` instead to make this easier to understand

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
ryan-conn: `replaceCourseCards` with an empty array creates the initial empty card, now that I look back at this I think I could just make the initial state a single course card again, and then if there are saved course cards it should be able to correctly update the initial blank one and avoid ambiguity like this

ryan-conn:
Fixed the issue where cards don't get replaced when browser navigation is used (thanks for the suggestion, I forgot you could add cleanup to `useEffect`), and everything else should be addressed now, I changed the default back to 1 course card again since `addCourseCard` can overwrite existing cards now

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -274,4 +286,108 @@ describe('Course Cards Redux', () => {
...
+      ];
+      fetchMock.mockImplementationOnce(testFetch);
+
+      // act
+      store.dispatch<any>(replaceCourseCards(courseCards, '202031'));
+      // wait for all actions to finish
+      await new Promise(setImmediate);
+
+      // assert
+      // testFetch with a MATH course has 1 section, which should be added
```
firejake308, line 364: Do y'all think we need to add one more test to make sure that the `replaceCourseCards` action works if the `fetch`es return out of order? So, for example, if the `/api/sections` call returned card #1 before card #0, they stay in order? Intuitively, it feels like something that could go wrong if someone messed with it, but Ryan's implementation right now is solid enough to where it feels like common sense, so I'm not sure.

ryan-conn: I added one just so that we have more thorough tests, and just in case we change the implementation later we can still know that this works

ryan-conn:
> Just had one idea for a potential test, but I don't have many more suggestions besides some of the ones from before that are still unresolved.

I added the test you suggested and fixed a couple of suggestions that I missed the first time around, everything should be addressed now

PR #304, opened 2020-07-20
User sessions/get first and last name api

### LINKED ISSUE #295: Add API route & displaying for retrieving a user's name when logged in
Need to add an API route (to `user_sessions` / the `sessions/` route) so we can retrieve a users full name, and retrieve & display this name on the frontend to indicate that a user is logged in.

ThatJuanGuy:
## Description

This pull request creates an API that allows for the logged in User's name to be retrieved. It is used to replace the login text where the login option used to let the user know their log in was successful. Should not be merged until after Social Auth.

## Rationale

Honestly kinda hard to do at the moment because it's not done and the name showing is kinda implemented poorly as it overwrites the login button text and ideally it would replace the button with a logout button or something that displayed the name but that would require a logout API (or clearing cookies) I think and that might be a separate PR.

## How to test

Despite the fact this shouldn't be merged until after Social Auth, I kinda needed code from that branch in order to get this one to work so I added that code to this branch. If you just boot up the localhost and login you should be able to see if the Google Login text was overwritten. Also, I still need to add tests for the API route.

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|

## Related tasks

Closes #295

### DIFF
--- a/autoscheduler/frontend/src/components/NavBar.tsx
+++ b/autoscheduler/frontend/src/components/NavBar.tsx
@@ -29,8 +29,8 @@ const NavBar: React.SFC = () => {
     fetch('sessions/get_full_name').then(
       (res) => res.json(),
     ).then((result) => {
-      if (result.full_name) {
-        setUsersName(result.full_name);
+      if (result.fullName) {
+        setUsersName(result.fullName);
       }
     });
     if (usersName === '') {
--- a/autoscheduler/frontend/src/components/NavBar.tsx
+++ b/autoscheduler/frontend/src/components/NavBar.tsx
@@ -23,22 +23,14 @@ const useStyles = makeStyles((theme) => ({
 const NavBar: React.SFC = () => {
   const classes = useStyles(appTheme);
 
-  const [usersName, setUsersName] = React.useState('');
-
-  function getUsersName(): void {
+  const [usersName, setUsersName] = React.useState('Google Login');
+  React.useEffect(() => {
     fetch('sessions/get_full_name').then(
       (res) => res.json(),
-    ).then((result) => {
-      if (result.fullName) {
-        setUsersName(result.fullName);
-      }
+    ).then(({ fullName }) => {
+      if (fullName) setUsersName(fullName);
     });
-    if (usersName === '') {
-      setUsersName('Google Login');
-    }
-  }
-
-  getUsersName();
+  }, []);
 
   return (
     <div className={classes.root}>
--- a/autoscheduler/frontend/src/components/NavBar.tsx
+++ b/autoscheduler/frontend/src/components/NavBar.tsx
@@ -23,6 +23,23 @@ const useStyles = makeStyles((theme) => ({
 const NavBar: React.SFC = () => {
   const classes = useStyles(appTheme);
 
+  const [usersName, setUsersName] = React.useState('');
+
+  function getUsersName(): void {
+    fetch('sessions/get_full_name').then(
+      (res) => res.json(),
+    ).then((result) => {
+      if (result.full_name) {
+        setUsersName(result.full_name);
+      }
+    });
+    if (usersName === '') {
+      setUsersName('Google Login');
+    }
+  }
+
+  getUsersName();
+
   return (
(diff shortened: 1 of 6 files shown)


gannonprudhomme:
Just FYI probs a better strategy of basing this off of social auth (#288 ) would be to create the branch when on that branch (or by doing `git rebase socialAuth`), and have the base branch of this PR be `socialAuth`. That way we only see changes for the sessions part of this.

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -23,6 +23,23 @@ const useStyles = makeStyles((theme) => ({
...
+      if (result.full_name) {
+        setUsersName(result.full_name);
+      }
+    });
+    if (usersName === '') {
+      setUsersName('Google Login');
+    }
+  }
+
+  getUsersName();
```
ryan-conn, line 41: You can change the initial value of `usersName` to `Google Login` and instead of calling `getUsersName` directly, use `React.useEffect` to make this a little cleaner (also notice how I destructure the json to simplify it a little bit):
```suggestion
  const [usersName, setUsersName] = React.useState('Google Login');
  React.useEffect(() => {
    fetch('sessions/get_full_name').then(
      (res) => res.json(),
    ).then(({ full_name }) => {
      if (full_name) setUsersName(full_name};
    });
  }, []);
```

autoscheduler/user_sessions/views.py
```diff
@@ -20,3 +21,14 @@ def set_last_term(request):
...
+
+@api_view(['GET'])
+def get_full_name(request):
+    """ View that retrieves the first and last name seperated by a space for the user
+    of the current session """
+    user_id = request.session.get('_auth_user_id')
+    if user_id is None:
+        return Response(status=400)
+    user = User.objects.get(pk=user_id)
+    response = {'full_name': user.get_full_name()}
```
ryan-conn, line 33: Really nitpicky but since this is JSON, you should use camelCase instead of snake_case for the key
```suggestion
    response = {'fullName': user.get_full_name()}
```

ThatJuanGuy: Yeah. You're right. Mocking fetch requests also apparently requires camelCase or it causes an error.

autoscheduler/user_sessions/tests/name_api_tests.py
```diff
@@ -0,0 +1,64 @@
...
+
+class UsersFullNameAPITests(APITestCase):
+    """ Tests the functionality of the get_full_name API """
+    def setUp(self):
+        """ Delete sessions table before each test to ensure new session """
+        Session.objects.all().delete()
+
+    def test_get_full_name_returns_correct_value_bob(self):
+        """ Tests that /sessions/get_full_name response is equal to the user's name """
+        #Arrange
```
ryan-conn, line 13: Put spaces after the `#`s in comments

autoscheduler/user_sessions/tests/name_api_tests.py
```diff
@@ -0,0 +1,64 @@
...
+                                   last_name='Bobbins')
+        user.set_password(password)
+        user.save()
+
+        expected_name = 'Bob Bobbins'
+        expected = {'fullName': expected_name}
+
+        #Act
+        self.client.login(username=username, password=password)
+        response = self.client.get(f'/sessions/get_full_name')
```
ryan-conn, line 26: Since these strings don't actually need formatting, use normal strings instead of f-strings

autoscheduler/user_sessions/views.py
```diff
@@ -62,3 +63,14 @@ def get_saved_courses(request):
         if courses:
             response = courses
     return Response(response)
+
+@api_view(['GET'])
+def get_full_name(request):
+    """ View that retrieves the first and last name seperated by a space for the user
+    of the current session """
```
gannonprudhomme, line 70:
```suggestion
    """ View that retrieves the first and last name separated by a space for the user
          of the current session
    """
```

Note you also misspelled "separated"

PR #308, opened 2020-07-25
Made scrape_courses override previous data

### LINKED ISSUE #293: Change scrape_courses to bulk override/update duplicate entries
Currently, since `scrape_courses` uses `bulk_create`, when it hits a duplicate entry it skips updating/overwriting the values. As such, you can't run `scrape_courses` to update the previous data - you would have to nuke the DB first before running it again to get updated data.

gannonprudhomme:
## Description

Previously `scrape_courses` wouldn't override the previous data in the database since `bulk_create` doesn't overwrite on conflicts. 
This makes it overwrite previous entries by deleting them & uploading them. Hopefully adding the `transaction.atomic()` makes all of these commit at once, so there isn't a window where the models won't exist.

## Rationale

I deleted `slice_every` since it's really not necessary, as `batch_size` for `bulk_create` provides the same functionality. Sorry Ryan 😢 

## How to test

Run `scrape_courses` for a given term and check if it updates it. A quick way to check if it updates it (assuming the last time you ran it was before the Fall 2020 courses were updated) is to run the server & check (for `202031`) if `CSCE 121 - 200` exists. Before the fall semester shift it existed, but afterwards it doesn't. 

## Related tasks

Closes #293

### DIFF
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -182,7 +182,9 @@ def parse_all_courses(course_list, term: str, courses_set: set,
 
     return (parse_course(course, courses_set, instructors_set) for course in course_list)
 
-def get_course_data(depts_terms) -> Tuple[List[Instructor], List[Section], List[Meeting], List[Course]]: # pylint: disable=too-many-locals
+def get_course_data(  # pylint: disable=too-many-locals
+        depts_terms,
+) -> Tuple[List[Instructor], List[Section], List[Meeting], List[Course]]:
     """ Retrieves all of the course data from Banner """
     # This limit is artifical for speed at this point,
     concurrent_limit = 50
@@ -214,14 +216,13 @@ def get_course_data(depts_terms) -> Tuple[List[Instructor], List[Section], List[
 
     return (instructors, sections, meetings, courses)
 
-def save_models(models_tuple: Tuple[List[Instructor], List[Section], List[Meeting], List[Course]],
-                term: int, terms: List[int], options):
+def save_models(instructors: List[Instructor], sections: List[Section], # pylint: disable=too-many-arguments
+                meetings: List[Meeting], courses: List[Course], term: int,
+                terms: List[int], options):
     """ Takes in a tuple of the models and attempts to save them
         "bulk updates" the models by deleting the according models then re-saving them
         in a single transaction.
     """
-    instructors, sections, meetings, courses = models_tuple
-
     start_save = time.time()
     start = time.time()
     Instructor.objects.bulk_create(instructors,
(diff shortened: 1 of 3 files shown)


autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -242,26 +243,48 @@ def handle(self, *args, **options): # pylint: disable=too-many-locals, too-many-
                 meetings.extend(meetings_list)
 
         start = time.time()
-        for slc in slice_every(instructors, 50_000):
-            Instructor.objects.bulk_create(slc, ignore_conflicts=True)
+        Instructor.objects.bulk_create(instructors,
+                                       ignore_conflicts=True, batch_size=50_000)
```
ryan-conn, line 247: Instructors only store email addresses now (which we don't actually use anywhere) so it doesn't matter right now, but it's worth pointing out that this doesn't actually update existing instructors, do you think it's worth updating them

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -190,7 +191,7 @@ def add_arguments(self, parser):
         parser.add_argument('--year', '-y', type=int,
                             help="A year to scrape all courses for, such as 2019")
 
-    def handle(self, *args, **options): # pylint: disable=too-many-locals, too-many-statements
+    def handle(self, *args, **options): # pylint: disable=too-many-locals, too-many-statements, too-many-branches
```
ryan-conn, line 194: I feel like we should split up the logic here into a few different functions instead of ignoring the linter since this function is getting really long, `get_course_data()` for the asyncio loop + making the model arrays and `save_models()` for all of the logic for saving seems like a good way to do it

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -21,18 +21,3 @@ def get_all_terms(year: int = -1) -> List[str]:
 
     return [f"{year}{semester}{location}"
             for year, semester, location in product(years, semesters, locations)]
-
-
-def slice_every(iterable: Iterable, n: int) -> Iterable[Iterable]:
```
ryan-conn, line 26: rest in peace my sweet child

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Yeah I intentionally didn't update them as I didn't think it really mattered since we don't use the emails. I can if you'd like but I don't think it's necessary really

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Yeah I figured as much, I think I was just being lazy at the time

autoscheduler/scraper/management/commands/utils/scraper_utils.py
gannonprudhomme: gif

autoscheduler/scraper/management/commands/scrape_courses.py
ryan-conn: You can leave it as is, we can change it if we actually start using emails for something but for now this works

PR #309, opened 2020-07-26
Organized SectionSelect tests

firejake308:
## Description

I separated out the `<SectionSelect>` tests from the `<CourseSelectCard>` tests. In doing so, I also removed the `fetch` from a lot of the SectionSelect tests, instead just pre-populating the Redux store with the appropriate data.

## Rationale

I originally expected this to reduce test runtime, since there would be less asynchronous fetching and less stuff to render, but it turned out to be about the same, as far as I can tell. However, I think this does still improve test readability, so I'm still posting it as a draft PR. Lmk what y'all think and we can make it a full PR if you think it's useful.

## How to test

`npm run test`

## Screenshots

Doesn't affect UI

## Related tasks

None, just felt like it


autoscheduler/frontend/src/tests/ui/CourseSelectCard.test.tsx
```diff
@@ -806,438 +697,4 @@ describe('Course Select Card UI', () => {
...
-      fireEvent.click(await findByText('CSCE 121'));
-
-      fireEvent.click(getByText('Section'));
-
-      // assert
-      // verify text for seats is correct
-      const sectionText = (await findAllByText(/ seats left/))[2];
-      expect(sectionText).toHaveTextContent('-1/25 seats left');
-    });
-  });
```
gannonprudhomme, line 1242: my god I never realized how absolutely _massive_ this file was, great call splitting this up

gannonprudhomme:
Honestly just given how large the course card tests file is, I definitely this is worth doing

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -25,15 +25,20 @@ const ConfigureCard: React.FC = () => {
...
   const [showSnackbar, setShowSnackbar] = React.useState(false);
-  // Holds a reference to the DOM element to check if the component is still mounted
-  const ref = React.useRef();
+
   const courseCards = useSelector<RootState, CourseCardArray>((state) => state.courseCards);
   const term = useSelector<RootState, string>((state) => state.term);
   const avsList = useSelector<RootState, Availability[]>((state) => state.availability);
   const dispatch = useDispatch();
 
+  // Holds a reference to the DOM element to check if the component is still mounted
```
gannonprudhomme, line 34: So what does mounted mean again?

firejake308: Basically, mounting happends when React intializes the component, and unmounting happens when the component's about to be destroyed and garbage-collected or whatever

PR #319, opened 2020-08-09
Update saved courses when course cards are updated

### LINKED ISSUE #299: Periodically save courses when course cards are changed
Instead of making an API call to save course cards when schedules are generated, we should create a function that takes 2 parameters: the first should be arguments for fetch as an array, and the second should be a max frequency (something like 15 seconds). Then, whenever a course card is changed, it should call this function. When it's called, it should immediately call fetch with the provided arguments if the provided time has passed, and otherwise queue a fetch. If it's called a second time before the max frequency has passed, it should discard the information from the first fetch (as it's now outdated) and instead when the timer expires call the fetch with the new arguments.

This function should also be used when we store availabilities (and perhaps when saving schedules), since it's generic enough to take in any fetch parameters.

ryan-conn:
## Description
Saved courses now continuously update as the user enters data. To do this, I created a `throttle` function to limit the number of API calls being made.

## Rationale
Couple of things I want to point out:
- if you close the page, your courses will be saved immediately in order to prevent data loss. However, this doesn't work as intended if you refresh, since the call to save courses will more than likely still be going on before the page is refreshed, and it will fetch slightly outdated courses. I'm not sure how we could fix this without compromising responsiveness, so any suggestions are welcome
- I didn't add any tests to make sure courses are saved as expected yet since I feel like we should all discuss when we should be saving in more detail first
- I made the timeout for throttling 15 seconds since that seemed reasonable, if we can't find a way around the issue with refreshing we might want to make the timer a bit shorter

## How to test
Change some course cards, and either close the page or change to a new term, your course cards should always be completely up-to-date. If you refresh, they revert to how they were when the throttle timer finished last (due to the issue I pointed out in [Rationale](#rationale).

## Related tasks
Closes #299


gannonprudhomme:
Need to add tests to this (I think?) and make it a full PR

ryan-conn:
A few changes:
- Throttle function uses intervals rather than timeouts now, so even when you're updating courses they'll be saved every 15 seconds
- Courses are saved immediately when you change terms so that if you switch back to the original term, the updated courses will be saved instead of ignored
- Added tests
- Merged master branch into this PR to fix the backend linting errors

I'm happy with how the saving functionality works now and there are tests, so I'm going to make this an actual PR now instead of a draft

autoscheduler/frontend/src/utils/throttle.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Function to throttles the execution of other functions
+ * @param id ID to use, each ID operates on a separate timer
+ * @param callback Callback to run after delay expires
+ * @param delay Delay (in ms) before running effect
+ * @param runOtherIDs Whether to immediately run pending callbacks for IDs other than specified id
```
firejake308, line 6: What is the purpose of this parameter?

autoscheduler/frontend/src/utils/throttle.ts
```diff
@@ -0,0 +1,54 @@
...
+ */
+const throttle = ((): (id: string, callback: any, delay: number, runOtherIDs: boolean) => void => {
+  const intervals = new Map<string, number>();
+  const intervalTimes = new Map<string, number>();
+  const functions = new Map<string, any>();
+
+  return (id: string, callback: any, delay: number, runOtherIDs = false): void => {
+    // Update interval frequency if needed
+    if (delay !== intervalTimes.get(id)) {
```
firejake308, line 15: When would the delay stored in `intervalTimes` not match the `delay` passed as a parameter?

autoscheduler/frontend/src/tests/ui/CourseSelectColumn.test.tsx
```diff
@@ -150,11 +154,78 @@ describe('CourseSelectColumn', () => {
...
+        </Provider>,
+      );
+
+      // act
+      store.dispatch(setTerm('202031'));
+      store.dispatch(setTerm('201931'));
+      await new Promise(setImmediate);
+
+      // assert
+      // console.log(fetchMock.mock.calls)
```
firejake308, line 206: Remember to remove this console log before merging

autoscheduler/frontend/src/utils/throttle.ts
ryan-conn: This happens in a couple situations:
- The first time it's called, the get returns `undefined` and this block is run to set the initial interval
- If you change the delay in the code that calls `throttle`, this will update the interval to the correct timer (this isn't being done right now, but doing it this way gives more flexibility)

autoscheduler/frontend/src/utils/throttle.ts
ryan-conn: It makes it so only one ID can be active at once, since it runs all of the others immediately if the parameter is true. Right now it's being used so that whenever you switch terms, the courses are saved for other terms immediately rather than having to wait for their intervals to run out (and if you select the old timer before the interval is over, lose your last batch of changes).

autoscheduler/frontend/src/utils/throttle.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Function to throttles the execution of other functions
+ * @param id ID to use, each ID operates on a separate timer
+ * @param callback Callback to run after delay expires
+ * @param delay Delay (in ms) before running effect
+ * @param runOtherIDs Whether to immediately run pending callbacks for IDs other than specified id
+ */
+const throttle = ((): (id: string, callback: any, delay: number, runOtherIDs: boolean) => void => {
```
gannonprudhomme, line 8: Do you think this function deserves tests of its own? It does have a decent amount of branching/logic to it

autoscheduler/frontend/src/utils/throttle.ts
```diff
@@ -0,0 +1,54 @@
+/**
+ * Function to throttles the execution of other functions
```
gannonprudhomme, line 2: I think this is due for more comments - the starting syntax of this is _weird_.

So from how I understand it (just spit-balling to double check my understanding):, `throttle` is a function that calls itself immediately and then in turn returns a function (represented by the inner `return (id: string, ....`). So the first 3 lines in it (defining the variables, e.g. `const interval`) is only called _once_, but the inside of it is called whenever you call the function (like in `CourseSelectColumn`). Javascript is so weird

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -33,6 +34,44 @@ const CourseSelectColumn: React.FC = () => {
     return (): void => { dispatch(clearCourseCards()); };
   }, [term, dispatch]);
 
+  // When courseCards is changed, save courses (and throttle API calls)
```
gannonprudhomme, line 37: I'd change this to something like `When courseCards is changed, set off a timer that will save the courses in 15 seconds. Previous timer is invalidated if this gets called again (when course cards are further updated)` (but more succinctly lol)

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -33,6 +34,44 @@ const CourseSelectColumn: React.FC = () => {
...
+
+      const body = JSON.stringify({
+        courses,
+        term,
+      });
+
+      fetch('sessions/save_courses', {
+        method: 'PUT',
+        body,
+      });
```
gannonprudhomme, line 68:
```suggestion
      fetch('sessions/save_courses', {
        method: 'PUT',
        JSON.stringify({ courses, term, }),
      });
```

I think I'd prefer if you shortened it to this form, but really just a nitpick so feel free to disregard

gannonprudhomme (commented):
Most of these are really just minor comments/nitpicks - this looks great otherwise. Took a bit for it to click for me, but the `throttle` function is really clever, good job.

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -108,29 +108,6 @@ const ConfigureCard: React.FC = () => {
         dispatch(selectSchedule(0));
         if (isMounted.current) setLoading(false);
       });
-
-    // make request to save course cards
-    const courseData: SerializedCourseCardOptions[] = [];
```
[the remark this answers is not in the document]
ryan-conn: Yeah, I think just having the 15-second autosave seems like a better solution since when you click generate schedules, it'll automatically save your courses in the next 5-15 seconds anyways

autoscheduler/frontend/src/utils/throttle.ts
ryan-conn: Yeah, the technical term is an "immediately invoked function expression", and I used it for the reason you described (keeping an internal state for the function). When the module is imported, it immediately creates the data structures used to keep track of functions, and calling it from somewhere will use the function that's returned by it.

gannonprudhomme:
Need to update this with master to resolve some conflicts

autoscheduler/frontend/src/utils/throttle.ts
ryan-conn: I tested all of the functionality in the `saves course cards` tests for `CourseSelectColumn` except for the timer (since having any tests using a timer even in a different file was causing a lot of other tests to fail), I can see why we'd want to have tests for it to be more thorough but I feel like as long as the behavior from our components is correct it's fine. I can add them if you want though

ryan-conn:
All the changes suggested (other than throttle tests) have been implemented, so this should be ready for review again

gannonprudhomme (approved):

PR #320, opened 2020-08-12
Multi day availability

### LINKED ISSUE #227: Add dragging availabilities over multiple days
Should be able to start dragging an availability for one day and move into the next column, and it should create an availability for both days.

firejake308:
## Description

Allows users to drag across multiple days in order to create an availability that occurs at the same time on every day.

## Rationale

I had to make substantial changes to the Redux for availabilities. First, there can now be multiple selected availabilties, so `setSelectedAvailability` was replaced with `addSelectedAvailability` and `removeSelectedAvailability`. Second, I didn't like how the old system merged availabilities as soon as they were added, so merging now occurs only when the `MERGE_AVAILABILITY` action is dispatched and not when `ADD_AVAILABILITY` is dispatched. Third, the reducer for merging can now merge multiple new availabilities using the `numNewAvs` argument, which is necessary when 5 availabilities are created at once. Finally, the `mergeSelectedAvailability` action was changed to allow users to drag upwards across multiple days. How is that related? I don't know, but it worked ¯\\\_(ツ)\_/¯.

## How to test

The availability system has a lot of edge cases. Try dragging on one day and multiple days, toward the beginning and end of the day, creating small availabilities, and dragging inside of other availabilities. All of these functionalities have been buggy at some point, and I've done my best to smooth out any bugs that I found, but there are likely to be a few more that I missed.

## Screenshots

Not really necessary. It looks exactly how you think it would look. (multiple availability blocks, one on each day).

## Related tasks

Closes #227


ThatJuanGuy:
Nice job on the feature! Making availabilties feels so much more streamlined now.

I did my best to break it. Here's what I got so far. When creating multiple availablities and dragging them past the upper bounds of the calendar, it causes the website to crash for some reason. By this I mean everything is replaced with the gray screen. Also, occasionally instead of crashing the website or immediately before crashing the website, the bars shoot out past the top of the calendar. I included some pictures. These bugs are easiest to replicate when dragging the mouse outside of the calendar to the left while still making the availabilities.

Gray Screen of Death pic:
GrayScreenOfDeath

Extending past top pic:
Availabilities Past Top

gannonprudhomme:
Can you update this with master (rebase or merge, up to you)? Has a conflict currently

firejake308:
Ahahahaha the bug you pointed out was literally caused by dragging backwards. I suck at math, fixed.

ryan-conn (changes requested):
I found a pretty minor bug, steps to reproduce:
1. Create an availability
2. Create another availability with the same start or end time and drag it across multiple days

mdavbug

I also noticed that as you keep dragging between days, the page can eventually get pretty laggy (try holding click and flailing your mouse around on the schedule for 5-10 seconds). Looking at redux devtools, it seems that selected availabilities aren't always being added/removed properly:
image

I see duplicate key errors in the console when I make multi-day availabilities, try adding the day to the key:
image


Other than that, I don't want to

firejake308:
Bugs found:

- Start on Tuesday. Move mouse to Monday. Now jump to Friday. Expect Tues - Fri. Get Mon - Fri.
- Create 2-4 availability on Tuesday. Create 4-6 availability on Wednesday. Drag to 6 on Tuesday. Expect 2- 6 on Tuesday and 4-6 on Wednesday. Get 4-6 on both days.

autoscheduler/frontend/src/tests/ui/Availability.test.tsx
```diff
@@ -276,6 +346,570 @@ describe('Availability UI', () => {
       expect(queryByLabelText('Adjust End Time'))
         .toHaveAttribute('aria-valuetext', expectedEnd);
     });
+
```
gannonprudhomme, line 349: This is almost 600 lines of tests for multi-day availabilities - I think it's worth making this it's own file

autoscheduler/frontend/src/tests/ui/Availability.test.tsx
```diff
@@ -276,6 +346,570 @@ describe('Availability UI', () => {
...
+        fireEvent.mouseUp(monday, endEventProps);
+
+        // assert that there are 3 availabilities, all with the same times
+        expect(getAllByText('BUSY')).toHaveLength(3);
+        expect(queryByText(/NaN/)).not.toBeInTheDocument();
+
+        const startTimes = queryAllByLabelText('Adjust Start Time');
+        expect(startTimes[0]).toHaveAttribute('aria-valuetext', expectedStart);
+        expect(startTimes[1]).toHaveAttribute('aria-valuetext', expectedStart);
+        expect(startTimes[2]).toHaveAttribute('aria-valuetext', expectedStart);
```
gannonprudhomme, line 539: These might be nicer to look at if you did this in a for-loop imo

gannonprudhomme (commented):
Few minor comments

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -228,7 +260,8 @@ const Schedule: React.FC = () => {
           time1,
           time2,
         }));
-        dispatch(mergeThenSelectAvailability({
+        dispatch(mergeAvailability());
+        dispatch(selectMergedAvailability({
           dayOfWeek: startDay,
           available: availabilityMode,
           time1,
```
ryan-conn, line 267: What does this achieve? I deleted it thinking it might have something to do with the same start/end time bug (it didn't lol) but it didn't seem to change any of the functionality, and it still passed all the tests. Also this is the only time `selectMergedAvailability` is called, so I'm curious why it's necessary

ryan-conn (changes requested):
Just so we have this documented somewhere for the future, this is what happens when you create a multi-day availability that starts or ends at the same time as an existing one. I know we said we'd ignore it for this PR, so I'm not going to tell you to fix it:
new bug

I spent a lot of time thinking about how we could fix this, but then I came up with a pretty good (imo) solution: make the drag handle a dynamic size based on how long the calendar is, so that it's literally impossible to click on the same time as another availability block without resizing it, and make it so drag handles can only create single-day availabilities. I think this makes more sense than the way it's currently being done (being able to click the same time and having 2 different potential actions is a bit strange, and the drag handles make it seem like they only resize a single availability anyways)

Wouldn't this solve the issue? It seems like by far the easiest solution to the bug, and shouldn't involve too much refactoring.

If not, I'd like for us to at least propose a way to fix it before this gets merged, so that we have an action plan moving forward. I know you proposed making availabilities based on IDs, but if we're going to go that far in refactoring this then I think completely separating already existing availabilities from selected ones, then just adding the selected availabilities (then merging) would be a lot cleaner and easier to maintain.

Other than that I don't have too much to comment on, other than that one bug this works fine and you added plenty of tests.

autoscheduler/frontend/src/tests/redux/Availability.test.ts
```diff
@@ -209,6 +242,141 @@ describe('Availabilities', () => {
...
+        dayOfWeek: DayOfWeek.MON,
+        ...unmergedAvailability,
+      });
+      expect(finalAvailabiltiies).toContainEqual<Availability>({
+        dayOfWeek: DayOfWeek.TUE,
+        ...mergedAvailability,
+      });
+    });
+
+    test('if multiple new availabiltiies overlap with multiple old availabilities', () => {
```
gannonprudhomme, line 302:
```suggestion
    test('if multiple new availabilties overlap with multiple old availabilities', () => {
```

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
firejake308: The `selectMergedAvailability` action exists to deal with a very specific edge case. When updating selected availabilities, the correct av to update is identified by the `time1` fed into the action, as this is where the drag originated and should be unchanged from the original av that is currently in the Redux. As the selected av is continuously dragged, `time2` is always changing, but its `time1` is always the same, and can therefore be used as a unique identifier. 

Unless. Unless the av is merged immediately after being created, in which case, its `time1` might actually become the start/end time of the av it was merged with. Now when would such a thing occur? Pretty much the only situation I can think of is the one that you mentioned earlier, where you click outside of an av but pick a start time where another av already exists. In that case, a new av is created shortly, then immediately merged with the other one so that you actually end up dragging that one.

This oculd be resolved by applying either of the suggested changes of identifying availabilities by ID rather than time, and/or making drag handles size themselves dynamically. In either case, we could theoretically remove this funky `selectMergedAvailability` action.

As for why this isn't included in tests, I remember trying to test this but giving up earlier. I would be willing to give it another shot if y'all think it's worth.

autoscheduler/frontend/src/tests/ui/Availability.test.tsx
```diff
@@ -276,6 +346,570 @@ describe('Availability UI', () => {
       expect(queryByLabelText('Adjust End Time'))
         .toHaveAttribute('aria-valuetext', expectedEnd);
     });
+
+    describe('across multiple days', () => {
+      test('if the user starts on Monday, drags downward, then releases on Tuesday', () => {
```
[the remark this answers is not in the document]
firejake308: Attempted to fix, lmk if there's still room for improvement

gannonprudhomme (changes requested):
Idk why it's not failing, but there are a few ESLint errors you have in here (guess it's b/c they're "warnings" not "errors"?. Otherwise I think this should be good.

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
gannonprudhomme: Given that it's a very specific edge case, then I think it would be worth adding tests for (assuming it doesn't take like an hour to add them lol) since a "very specific edge case" is probably something that's like to be missed when we're QA testing anytime we change this going forward

firejake308: Btw, turns out the bug was actually fixed by lines 240-253 and this function appears to be unnecessary. Removing the ridiculously clunky `selectMergedAvailability` action

firejake308: Also, I realized that we do have a test for the bug that I was thinking of, and it's on line 336 of `Availability.test.tsx`

gannonprudhomme (approved):

PR #321, opened 2020-08-13
Page layout

### LINKED ISSUE #187: Figure out how to explain the website
We first need to figure out what text to add to the landing page in order to the landing page.

Second, I think it might be beneficial if we had a gif or video on the homepage showing what the website actually does.

Also, I think we need some visual that would show them that you can drag on the schedule to make availabilities, such as through an interactive tutorial or a gif 

Might also be worth explaining that the schedules generate random combinations, so if you don't like the combinations it gives you you can try to regenerate it (assuming we don't set a consistent seed in #167)

firejake308:
## Description

Split up the schedule page into many different pages. The key benefit to this layout is that it allows the user to see one task at a time instead of being overloaded with the entire functionality of the website at once.

*also, PR is drafted because I haven't made any tests yet

## Rationale

The current step is stored in context and accessed via a custom hook. We could alternatively make a new Redux slice for this.

## How to test

Click around between the pages and see how it feels for yourself

## Screenshots

image

## Related tasks

Closes #187


autoscheduler/frontend/src/components/App/App.tsx
```diff
@@ -1,23 +1,43 @@
...
-import SchedulingPage from '../SchedulingPage/SchedulingPage';
+import CourseSelectPage from '../CourseSelectPage/CourseSelectPage';
+import AvailabilityPage from '../AvailabilityPage/AvailabilityPage';
+import SchedulesPage from '../SchedulesPage/SchedulesPage';
+import CustomizeSchedulePage from '../CustomizeSchedulePage/CustomizeSchedulePage';
+import { StepContext } from '../NavBar/stepManager';
+import steps from '../NavBar/steps';
 
 const App: React.SFC = function App() {
+  const [activeStep, _setActiveStep] = React.useState(
```
gannonprudhomme, line 17: Instead of changing App.tsx, I think this should be its separate file & route so we can basically A/B test it, as I think we should really think and get a good feel w/ it over a bit of time before switching completely to a different layout if that makes sense

firejake308: Okay, I put back in the old routes and set up for A/B testing. After picking a term, the site will randomly decide whether to take them to the single-page layout or the multi-page layout. A cookie will then remember that selection so that they continue to see the same layout the next time they return.

gannonprudhomme:
Should we close this since it's going to be remade into the tutorial?

firejake308:
Yes

PR #326, opened 2020-08-30
Delete schedule dialog

ryan-conn:
## Description

Creates a dialog when the user attempts to delete a saved schedule, also fixes the touch ripple of schedules playing whenever a schedule button is clicked

## Rationale

I used `useRef()` on the text component to determine the position of the save/delete buttons, since they have to have `position: absolute` to not trigger the touch ripple from the schedule list item.

## How to test
Save a schedule and try to delete it, the dialog will appear. You can use the buttons or press escape/enter to control it.

## Screenshots
image

## Related tasks
We never made a task for this but I was supposed to work on it ¯\\_(ツ)_/¯


ryan-conn:
Good catch, the `ListItemSecondaryAction` was automatically creating space for itself even though it was absolutely positioned, so I just made the `MiniSchedule` the last child

autoscheduler/frontend/src/tests/ui/SchedulePreview.test.tsx
```diff
@@ -237,4 +238,54 @@ describe('SchedulePreview component', () => {
...
+      fireEvent.click(deleteScheduleButton);
+
+      // assert
+      const schedules = store.getState().schedules.allSchedules;
+      expect(schedules).toHaveLength(1);
+      expect(schedules[0]).toEqual(testSchedule1);
+    });
+
+    test('when deleted from the dialog from a saved schedule', async () => {
+      // TODO: complete
```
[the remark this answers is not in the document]
ryan-conn: fixed

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,95 +1,19 @@
...
-            </span>
-          ))
-        }
-      />
-      <MiniSchedule schedule={schedule} />
-    </ListItem>
-  );
+  const scheduleListItems = schedules.length === 0
+    ? <p className={styles.noSchedules}>No schedules available.</p>
+    // eslint-disable-next-line react/no-array-index-key
```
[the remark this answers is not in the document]
ryan-conn: Yeah that was my thought as well, for now we don't have anything deterministic to use for the key anyways

ryan-conn:
Here's what I made for a maroon header:
image

Let me know if you guys want it changed in any way @firejake308 @gannonprudhomme @ThatJuanGuy 

I'd also like to point out that there are a couple ways of doing this:
- Use `createStyles` to style the `DialogTitle`
- Use a theme override for `DialogTitle` to make it so all dialog titles will be themed this way

I think using a theme override seems like a better idea, since if we make any more dialogs in the future they'll automatically be styled the same way, thoughts?

ThatJuanGuy:
It looks pretty good imo. I think right aligning the buttons with the text would look slightly nicer but that's pretty nit picky.
Original
image
Right Aligned
image

gannonprudhomme:
> It looks pretty good imo. ~I think right aligning the buttons with the text would look slightly nicer but that's pretty nit picky.~
> Original
> image
> Right Aligned
> image
> 
> Edit: Actually never mind. the corner spacing looks worse the more I look at it. The original was superior

I think this is on the right track actually - I think the right edge of the delete button should align with the text in the dialog (like it is with the right-aligned screenshot). Or rather, the dialog should only be as wide as it needs to fit the text on it (probably). I do agree that in that particular picture that extra gap isn't ideal, but otherwise I like it.

I also think we should decrease the font size of the header to something closer to what Adel had in his mockup from above. Lastly, the margins seem to be inconsistent: there appears to be much more margin on the left side than on the right side.

> I think using a theme override seems like a better idea, since if we make any more dialogs in the future they'll automatically be styled the same way, thoughts?

Yeah I think theme override is definitely the move - I can't imagine us wanting dialogs to appear any different than how we configure them in this PR

ryan-conn:
> I think the right edge of the delete button should align with the text in the dialog (like it is with the right-aligned screenshot). Or rather, the dialog should only be as wide as it needs to fit the text on it (probably). I do agree that in that particular picture that extra gap isn't ideal, but otherwise I like it.

This is actually what the dialog already does (if you change the text to something else you can see that it changes the size of the whole dialog), it's just that using `<DialogActions>` doesn't use the same margins as the `<DialogContent>`. I can move the buttons to a flexbox that's right-aligned if that's what we'd rather go with, but here's an example from the official material design website:
image
The padding is smaller for the buttons, and tbh I think I like it more since the it makes the buttons stand out more from the text (also look at the image Carlos posted for a comparison)

> I also think we should decrease the font size of the header to something closer to what Adel had in his mockup from above.

The current size is `1.25rem`, here's a comparison:
`1.25rem`
image
`1.2rem`
image
`1.1rem`
image
`1rem`
image
imo the best looking one is `1.2rem` since it's not as thick but is still big enough to distinguish it from the normal text

> Lastly, the margins seem to be inconsistent: there appears to be much more margin on the left side than on the right side.

They're the same, unless you mean the buttons going past the padding of the content

gannonprudhomme:
> > Lastly, the margins seem to be inconsistent: there appears to be much more margin on the left side than on the right side.
> 
> They're the same, unless you mean the buttons going past the padding of the content

I don't think they are (see the blue H looking things at top left & bottom right): 
Inked91888516-0bf06a00-ec52-11ea-96e8-ff67317a98fb_LI

ryan-conn:
> I don't think they are (see the blue H looking things at top left & bottom right): 

Like I said the buttons have different padding because they're part of the `<DialogActions>` component, but the text itself has 24px padding on both sides

gannonprudhomme:
> > I don't think they are (see the blue H looking things at top left & bottom right):
> 
> Like I said the buttons have different padding because they're part of the `<DialogActions>` component, but the text itself has 24px padding on both sides

Then we should reduce the 24px padding imo, I don't think padding should ever be greater than 16px honestly

ryan-conn:
> Then we should reduce the 24px padding imo, I don't think padding should ever be greater than 16px honestly

| 16px | 8px |
|-|-|
| image | image |

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/Buttons/DeleteScheduleButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+  const handleKeyPress = (e: React.KeyboardEvent<HTMLElement>): void => {
+    // Allow deleting by pressing enter instead of clicking the confirm button
+    if (dialogOpen && e.key === 'Enter') handleConfirmDelete();
+  };
+
+  return (
+    <>
+      <Tooltip title="Delete" placement="top">
+        <IconButton size="small" onClick={handleDeleteButtonClick} data-testid="delete-schedule">
```
firejake308, line 63: Do you think you could use `aria-label` instead of `data-testid`? That way, we would be able to access the element during tests, but we'd also be improving our accessibility instead of having props that are useless in production

firejake308 (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/Buttons/DeleteScheduleButton.tsx
ryan-conn: Yeah I agree with this, it should be addressed now

ryan-conn:
Changes should all be addressed

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/Buttons/DeleteScheduleButton.tsx
```diff
@@ -0,0 +1,94 @@
...
+}
+
+const DeleteScheduleButton: React.FC<DeleteScheduleButtonProps> = ({ index }) => {
+  const dispatch = useDispatch();
+
+  const schedule = useSelector<RootState, Meeting[]>((state) => (
+    state.schedules.allSchedules[index]
+  ));
+  const savedSchedules = useSelector<RootState, Meeting[][]>((state) => (
+    state.schedules.savedSchedules
```
firejake308, line 22: I think I recall you saying that you had changed to making `saved` a field for each schedule rather than storing a separate array in Redux. Was that in a different PR?

firejake308 (approved):
🚀

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/Buttons/DeleteScheduleButton.tsx
ryan-conn: Yeah that's in the rename schedules PR, I decided to make all the redux changes there so it's more clear what exactly I changed for it

PR #338, opened 2020-09-03
Logout button

### LINKED ISSUE #329: Logout Button Icon
Replace "Logout" with a door icon or something. 
Also make name font pretty because times new roman or whatever it is currently looks bad

ThatJuanGuy:
## Description
This branch adds a working logout button to the website
## Rationale
Seeing as how it was possible to sign in, signing out seemed like the logical next step

## How to test
Login to the website. You should be logged in and your name and a logout button should appear where the login button used to be. Logout of the website, the login button should reappear and you should be logged out.

## Screenshots

|Before|After|
|--|--|
|LoggedOut|LogoutIconOnlyArrow|

## Related tasks

Closes #329 #313


ThatJuanGuy:
Material UI only had the door and arrow logo as logout icons. Here's all the combinations of Logout and an icon. Personally, I like either Only Arrow and Door Then Logout. Of those two, I have a slight preference for Door then Logout. However, I wanted to get your opinions as to what yall think looks best. 

Only Door
LogoutIconOnlyDoor

Only Arrow
LogoutIconOnlyArrow

Logout Then Arrow
LogoutIconWordFirstArrow

Logout Then Door
LogoutIconWordFirstDoor

Arrow Then Logout
LogoutIconWordLastArrow

Door Then Logout
LogoutIconWordLastDoor

firejake308:
I prefer Only Arrow. For accessibility, we could add a label/alt text/tooltip to identify it as a logout button

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -23,6 +24,51 @@ const useStyles = makeStyles((theme) => ({
 const NavBar: React.SFC = () => {
   const classes = useStyles(appTheme);
 
+  const [usersName, setUsersName] = React.useState('');
+  React.useEffect(() => {
```
ryan-conn, line 28: Add a comment explaining that what this does (check if the user is logged in and sets name if so)

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -23,6 +24,51 @@ const useStyles = makeStyles((theme) => ({
...
+        title="Logout"
+        onClick={(): void => {
+          window.open('/sessions/logout', '_self');
+        }}
+      >
+        <ExitToAppOutlinedIcon />
+
+      </Button>
+    );
+  }
```
ryan-conn, line 69: These don't have to be functions, you can do this:
```jsx
const userInformation = usersName ? (
  // component to render when logged in
) : (
  // component to render when logged out
);
```

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -44,15 +90,20 @@ const NavBar: React.SFC = () => {
...
+              </div>
+            )
+            : (
+              <div>
+                <Typography variant="subtitle1">
+                  {usersName/* make font prettier */}
+                  {LogoutButton()}
+                </Typography>
+              </div>
+            )}
```
ryan-conn, line 106: I like to avoid putting as little logic as possible into the component that gets rendered, so that you tell at a glance how the components are nested rather than looking at the business logic. You can do a couple things to help this:
- Move the ternary logic outside of the JSX
- Make the login/logout buttons include the typography, username, etc. instead of just the `Button` component

and for this part of the code, just put the variable that has the correct components to render:
```suggestion
          {userInformation}
```

autoscheduler/user_sessions/views.py
```diff
@@ -20,3 +23,19 @@ def set_last_term(request):
...
+    if user_id is None:
+        return Response(status=400)
+    user = User.objects.get(pk=user_id)
+    response = {'fullName': user.get_full_name()}
+    return Response(response)
+
+def logout_view(request):
+    """ Logs out the user and redirects to index"""
+    logout(request)
+    return redirect('/')
```
ryan-conn, line 41: At the top of the file if you import logout like this:
```python
from django.contrib import auth
```
we can avoid having to call this `logout_view`, since none of our other views have `view` in the name:
```suggestion
def logout(request):
    """ Logs out the user and redirects to index"""
    auth.logout(request)
    return redirect('/')
```

ryan-conn (changes requested):
I agree with Adel's suggestion to just use the arrow icon, and also have a few ideas for making the NavBar more readable

gannonprudhomme:
Oh yeah when you log out it should give a toast that says like "You've successfully logged out" or something like that

ThatJuanGuy:
That’s a pretty good idea. I’ll add that.

ryan-conn:
Also, as we talked about in our last meeting can you create a `LoginButton` component and move all of the login/logout functionality there instead of `NavBar`? It would be nice to isolate all the fetching, redirection, and state updates

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
```diff
@@ -0,0 +1,74 @@
+import * as React from 'react';
+import { Button } from '@material-ui/core';
+import ExitToAppOutlinedIcon from '@material-ui/icons/ExitToAppOutlined';
```
firejake308, line 3: Let's just rename it to `LogoutIcon` for our purposes. Saves a little space, and it also explains what it's doing better imo

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+// This file contains the login button. When a user logs in, it is replaced by their name
+// followed by the logout button.
+const LoginButton: React.FC = () => {
+  // Used to track whether the API call has returned anything yet
+  const [loading, setLoading] = React.useState(true);
+
+  // used to logout the user
+  function logout(): void {
+    fetch('sessions/logout').then(
```
firejake308, line 12: `sessions/logout` should definitely be a POST, not a GET

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+      (res) => {
+        if (res.ok) {
+          window.location.reload();
+          return;
+        }
+        throw new Error(res.status.toString());
```
firejake308, line 18: Why throw an error if you're going to catch it without handling it 2 lines later? Is that for future updates maybe?

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+      },
+    ).catch(() => { });
+  }
+
+  // This checks for the logged in user's name and sets it if found.
+  // Otherwise, it throws an error and catches it so nothing breaks.
+  const [usersName, setUsersName] = React.useState('');
```
firejake308, line 25: Although it technically doesn't matter, I like declaring all of the state at the top of the component because it makes it easier to find later

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+        throw new Error(res.status.toString());
+      },
+    ).then(({ fullName }) => {
+      if (fullName) setUsersName(fullName);
+    }).catch(() => { })
+      .finally(() => { setLoading(false); });
+  }, []);
+
+  // Checks to make sure fetch call has finished before showing anything
+  if (loading) {
```
firejake308, line 42: Could simplify to one line

autoscheduler/frontend/src/components/App/App.tsx
```diff
@@ -4,7 +4,7 @@ import { ThemeProvider } from '@material-ui/styles';
 
 import theme from '../../theme';
 import * as styles from './App.css';
-import NavBar from '../NavBar';
+import NavBar from '../NavBar/NavBar';
```
firejake308, line 7: Interesting, it looks like the file should exist but linting says it doesn't. Weird.

autoscheduler/frontend/src/components/Navbar/NavBar.tsx
```diff
@@ -53,15 +45,9 @@ const NavBar: React.SFC = () => {
...
-
-          <Button
-            color="inherit"
-            onClick={(): void => {
-              window.open('/login/google-oauth2/', '_self');
-            }}
-          >
-            {usersName}
-          </Button>
+          <Typography variant="subtitle1">
```
firejake308, line 48: Why wrap it in a `<Typography>`?

autoscheduler/frontend/src/components/App/App.tsx
firejake308: Oh wait, it's the caps. Linux is case-sensitive, so we'll need to rename the folder from `Navbar` to `NavBar`

autoscheduler/frontend/src/tests/ui/NavBar.test.tsx
```diff
@@ -0,0 +1,153 @@
+import fetchMock, { enableFetchMocks } from 'jest-fetch-mock';
+// TODO ADD TEST FOR LOADING RETURN NULL
```
firejake308, line 2: Reminder to address this TODO

autoscheduler/frontend/src/tests/ui/NavBar.test.tsx
```diff
@@ -0,0 +1,153 @@
...
+        <NavBar />
+      </Provider>,
+    );
+
+    // act
+    const logoutButton = await findByRole('button', { name: 'Logout' });
+    act(() => { fireEvent.click(logoutButton); });
+
+    // assert
+    expect(window.open).toHaveBeenCalledWith('/sessions/logout', '_self');
```
firejake308, line 151: These tests look good to me. Nice work.

autoscheduler/user_sessions/views.py
```diff
@@ -76,3 +77,9 @@ def get_full_name(request):
     user = User.objects.get(pk=user_id)
     response = {'fullName': user.get_full_name()}
     return Response(response)
+
+@api_view(['GET'])
```
firejake308, line 81: Mentioned it earlier, but logout should be `POST` imo since it's not idempotent

firejake308 (changes requested):

autoscheduler/frontend/src/components/App/App.tsx
ThatJuanGuy: I thought I fixed this by renaming the folder before I pushed but apparently not. How do I make sure it gets renamed correctly? I'm thinking of deleting the folder, pushing that, adding it again with the correct casing, and pushing that but that seems pretty roundabout.

autoscheduler/frontend/src/components/Navbar/NavBar.tsx
ThatJuanGuy: I believe I was doing that to control the text size. I could move the typography into the button component if it looks weird there.

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
ThatJuanGuy: do you mean like `if (loading) return null;` ?

ThatJuanGuy: If so, I am aware. I just left it the way I did because I like having the curly braces. I can change it no problem though if you want.

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
ThatJuanGuy: The catch is meant to stop the website from crashing if the API gives a bad response. It was in the getName UseEffect because when the response didn't have JSON it caused the website to crash. I made this one by copy pasting the getName one and then modifying it. Since I'm not actually processing any of the response other than checking `if(res.ok)` I can probably get rid of it here without any consequences.

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
ThatJuanGuy: I had no idea I could rename this. Regardless it is done.

autoscheduler/frontend/src/components/Navbar/NavBar.tsx
firejake308: Ok, just move it into the `LogoutButton` component

autoscheduler/frontend/src/components/App/App.tsx
ryan-conn: If you delete a file and add another identical one in the same commit, git will recognize that you renamed the file. That should be enough to fix this

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
```diff
@@ -0,0 +1,74 @@
...
+    return null;
+  }
+  // Determines whether to show the login button  or the logout button
+  // and user's name depending on whether a user is logged in or not
+  return (
+    usersName ? (
+      <div>
+        {usersName}
```
ryan-conn, line 50: Can you add `text-transform: uppercase` to this? Since the login text is uppercase, this should match it for consistency

ryan-conn (changes requested):
Just a small change other than the ones Adel suggested, overall good work on this (especially the tests, they're pretty thorough and you didn't ask for help afaik despite not having much experience with them)

autoscheduler/frontend/src/components/Navbar/LoginButton.tsx
ryan-conn: Oh yeah also I realized when messing around with the text that the text and button aren't centered, also add `display: flex;` and `align-items: center;`

gannonprudhomme:
@ThatJuanGuy mind addressing these changes before the meeting? Would love to get this merged soon!

gannonprudhomme:
Could you edit this to make it pass the linting + testing? It's currently failing every frontend step

ThatJuanGuy:
> Could you edit this to make it pass the linting + testing? It's currently failing every frontend step. Been open for over a month so I'd like to get this + #366 merged before the website launch.

Just looked into it and it's because of the Navbar folder name. I deleted it before my last push and added it back with the correct casing (NavBar) but apparently the changes didn't take effect when I pushed. I'm just going to delete the folder, commit and push, add it back in a different commit and push again to see if that works.

ThatJuanGuy:
Still figuring out how to make the frontend tests for the refresh on logout and the button returning null if loading is true but website functionality wise it should be good to go.

autoscheduler/frontend/src/components/NavBar/LoginButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+        </Button>
+      </div>
+    ) : (
+      <Button
+        color="inherit"
+        onClick={(): void => {
+          window.open('/login/google-oauth2/', '_self');
+        }}
+      >
+    Login With Google
```
firejake308, line 86: Indent this line a little

autoscheduler/frontend/src/components/NavBar/NavBar.css
```diff
@@ -0,0 +1,4 @@
+#NameAndButton {
+    display: flex;
+    align-items: center;
+}
```
firejake308, line 4: Add new line

autoscheduler/frontend/src/components/NavBar/reloadPage.tsx
```diff
@@ -0,0 +1,15 @@
...
+// or replace with just window.location.reload() without reading the
+// related note in that file first at least. just ctrl f 'reload' to find it quickly
+
+function reloadPage(): void {
+  window.location.reload();
+}
+
+// This is written like this because the tests won't let me
+// replace the function with jest.fn() otherwise
+const reloadPageFunctions = {
```
firejake308, line 11: This file could be simplified to just `export default function reloadPage...`
When you mock it, you'll have to do something like this:
```
jest.mock('../reloadPage', () => ({
  default: jest.fn(),
}));
```

autoscheduler/frontend/src/tests/ui/NavBar.test.tsx
```diff
@@ -0,0 +1,166 @@
...
+      <Provider store={store}>
+        <NavBar />
+      </Provider>,
+    );
+
+    // act
+    // as of time of creation of this test, mocking window.location.reload is impossible.
+    // Navigation is not supported by jsdom. Thus I am placing reload inside a function
+    // called reloadPage in reloadPage.tsx that holds window.location.reload. If the
+    // reloadPageFunctions.reloadPage was called, window.location.refresh is guranteed
```
firejake308, line 157:
```suggestion
    // reloadPageFunctions.reloadPage was called, window.location.refresh is guaranteed
```

firejake308 (changes requested):
So I can't test this in browser because I get this error:

image

But I had a few suggestions to clean up the code

autoscheduler/frontend/src/components/NavBar/NavBar.css
```diff
@@ -0,0 +1,4 @@
+#NameAndButton {
```
ryan-conn, line 1:
```suggestion
.name-and-button {
```
To be consistent with our other CSS rules, also since we're using classes instead of ids for like everything else on the site might as well conform to it

autoscheduler/frontend/src/components/NavBar/LoginButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+  // This checks for the logged in user's name and sets it if found.
+  // Otherwise, it throws an error and catches it so nothing breaks.
+  React.useEffect(() => {
+    fetch('sessions/get_full_name').then(
+      (res) => {
+        if (res.ok) {
+          return res.json();
+        }
+
+        throw new Error(res.status.toString());
```
ryan-conn, line 48:
```suggestion
        throw new Error(String(res.status));
```
We should really be using `String()` instead of `.toString()` in places that use `.toString()` since it better fits typescript and can't throw errors, same for the other `.toString()` earlier in this file

autoscheduler/frontend/src/components/NavBar/LoginButton.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+// This file contains the login button. When a user logs in, it is replaced by their name
+// followed by the logout button.
+const LoginButton: React.FC = () => {
+  // Used to track whether the API call has returned anything yet
+  const [loading, setLoading] = React.useState(true);
+  // default set to defaultTestUserName instead of empty string becuase I thought
+  // an empty string might cause the refreshes on logout test to report false positives
+  // very easily depending on future changes.
+  const [usersName, setUsersName] = React.useState('defaultTestUserName');
```
ryan-conn, line 16:
```suggestion
  const [usersName, setUsersName] = React.useState<string>();
```
Use undefined as a placeholder value instead

ryan-conn (changes requested):
Couple of small changes and I'm ready to approve this

PR #340, opened 2020-09-07
Moved availability instructions to dialog

firejake308:
## Description

The instructions for how to create availabilities would never be read by users if they stayed in the `ConfigureCard`, where they were before this PR. This PR moves the instructions into a dialog that obscures the schedule until the user dismisses it.

The new, schedule-covering dialog is set to appear by default, and can be removed if the user has saved schedules, if the user has generated schedules, or if the user presses OK. However, it might be worth considering setting a cookie so that it's only shown once per user, because I could easily see this becoming a nuisance if you don't have any saved availabilities.

## Rationale

Placed in a separate component because the `Schedule` component is already a little thicc.

## How to test

Look at it: http://localhost:8080

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|
|image|image|

## Related tasks

None

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.css
@@ -2,7 +2,8 @@
     display: flex;
     flex-direction: column;
     justify-content: space-between;
-    width: calc(100% - 16px);
+    width: 100%;
+    box-sizing: border-box;
     padding: 8px;
     user-select: none;
 }
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.css
@@ -1,8 +1,9 @@
 .button-container {
     display: flex;
     flex-direction: column;
-    padding: 8px;
     justify-content: space-between;
+    width: calc(100% - 16px);
+    padding: 8px;
     user-select: none;
 }
 
@@ -13,9 +14,4 @@
 #card-header {
     text-align: center;
     width: 100%;
-}
-
-#instructions {
-    margin-left: 8px;
-    margin-right: 8px;
 }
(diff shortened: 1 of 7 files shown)


autoscheduler/frontend/src/components/SchedulingPage/Schedule/InstructionsDialog/InstructionsDialog.tsx
```diff
@@ -0,0 +1,51 @@
...
+const InstructionsDialog: React.FC = () => {
+  const [open, setOpen] = React.useState(true);
+  const hasCurrentAv = useSelector<RootState, boolean>(
+    (state) => state.availability.length > 0,
+  );
+  // can have 1 of 3 values:
+  // null -> user's first time on site
+  // ''   -> user has not yet made an av
+  // 'Y'  -> user has made an av or has manually dismissed dialog in the past
+  const hasPastAv = localStorage.getItem('has-created-availability');
```
ryan-conn, line 18: We should be able to just `return null` here if this is true, right? It would make renders very marginally faster since the dialog is no longer on the page, and clean up the logic later when you have to check if `hasPastAv`

firejake308: I tried it, but it looks ugly because currently, the `Fade` component will gradually fade out the dialog, whereas returning `null` makes the dialog disappear suddenly. Will keep as is.

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.css
```diff
@@ -1,8 +1,9 @@
 .button-container {
     display: flex;
     flex-direction: column;
-    padding: 8px;
     justify-content: space-between;
+    width: calc(100% - 16px);
```
ryan-conn, line 5:
```suggestion
    width: 100%;
    box-sizing: border-box;
```
`border-box` includes padding and border in the box size calculation

ryan-conn (dismissed):
Looks good, I noticed a way to make the css a bit cleaner but other than that this PR is fine

ryan-conn (approved):

PR #341, opened 2020-09-09
GCP: Scrape courses on a schedule

### LINKED ISSUE #192: Setup Google Cloud Tasks for scraping every X minutes
We need to look into setting up Google Cloud Tasks for scraping the current term's seats every X minutes, that way the current filled seats is as up to date as possible during the registration period.

They should really only run every 15 minutes when registration is actually live, and could probably just run once a day after registration is over (or even later).

This probably won't need to be done for awhile, but I'm creating the tasks so that it's on our mind. Frankly this won't actually need to be done until the Fall semester, but it would be nice if we're ready for it.

You can read more about Google Cloud Tasks [here](https://cloud.google.com/tasks/docs/how-to-index)

gannonprudhomme:
## Description

So this is mainly two parts: getting scrape_courses to connect to Google Cloud SQL and running scrape courses in Docker, and adding `--recent` flag to `scrape_courses`. I'll also discuss how the scheduling works in GCP, but since it's all just configs you really can't see the changes in this PR lol.

### Docker containers

So this actually won't work if you run it locally since the cloud sql proxy won't be authenticated. However if you run it on Compute Engine, it uses Google Default App Credentials that are automatically on the Compute Engine instance, so we don't need to pass any credentials.

Since I added new docker containers to the `docker-compose.yml` file, so we run the according container by specifying it, e.g. `docker-compose up --build scrape-courses` (`cloud-sql-proxy` will automatically start since `scrape-courses` depends on it). Note that we have to do the same thing when we run the tests for CI, i.e. `docker-compose up --build web`.

For the settings changes, I opted to use the docker settings file rather than add another to `base.py` just to reduce confusion with the environment variables. The `_IS_SCHEDULE` flag db settings is the same as the `_IS_PROXY` one, except that the host is `cloud-sql-proxy` (corresponds to the docker container name).

### Scrape_courses --recent

Pretty straight-forward, just adds a `--recent` flag to scrape_courses. I basically follow the logic that if it's before the registration date for that term, scrape the one before it.

### GCP Scheduling setup

It follows [this tutorial](https://cloud.google.com/scheduler/docs/start-and-stop-compute-engine-instances-on-a-schedule) on the GCP website.

It starts with the [Cloud Scheduler](https://console.cloud.google.com/cloudscheduler), which I have configured to run every 3 hours. Cloud Scheduler triggers a [Cloud Pub/Sub](https://console.cloud.google.com/cloudpubsub/topic/list) `start-instance-event`. Then there is a [Cloud Function](https://console.cloud.google.com/functions) that responds to the `start-instance-event` (called `startInstancePubSub`). The Cloud Function runs a JS script (copy/pasted from the tutorial) that starts the specified [Compute Engine](https://console.cloud.google.com/compute/instances) instances. Next, when the Compute Engine starts up, it runs a startup script that fetches the repo and updates itself with `origin/master`, runs `docker-compose up --build scrape-courses`, then runs `sudo terminate` at the end of it to stop the instance. 

Succinctly: 
1. Cloud Scheduler runs every 3 hours, sending a Pub/Sub `start-instance-event` event
2. The Pub/Sub event triggers a Cloud Function
3. The Cloud Function starts the Compute Engine instance
4. Compute Engine instance runs the startup script and updates & runs scrape_courses
5. Compute Engine script finishes and runs `sudo terminate` to stop the instance.

## How to test

You can test the `--recent` change just by running `scrape_courses --recent` and ensuring it scrapes for the correct terms (fall+summer 2020 for all locations).

Testing the docker change is a bit harder since we have to change it to pass in the credentials file. As such, you can just check that Compute Engine is running correctly by checking out the logs for it [here](https://console.cloud.google.com/sql/instances)

## Related tasks

Closes #192

### DIFF
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -11,7 +11,9 @@ from scraper.banner_requests import BannerRequests
 from scraper.models import Course, Instructor, Section, Meeting, Department
 from scraper.models.course import generate_course_id
 from scraper.models.section import generate_meeting_id
-from scraper.management.commands.utils.scraper_utils import get_all_terms
+from scraper.management.commands.utils.scraper_utils import (
+    get_all_terms, determine_terms_to_scrape,
+)
 
 # Set of the courses' ID's
 def convert_meeting_time(string_time: str) -> datetime.time:
@@ -283,6 +285,8 @@ class Command(base.BaseCommand):
                             help="A valid term code, such as 201931")
         parser.add_argument('--year', '-y', type=int,
                             help="A year to scrape all courses for, such as 2019")
+        parser.add_argument('--recent', '-r', action='store_true',
+                            help="Scrapes the most recent semester(s) for all locations")
 
     def handle(self, *args, **options):
         depts_terms = []
@@ -291,17 +295,23 @@ class Command(base.BaseCommand):
         terms = None
 
         if options['term']:
-            if options['year']: # Show an error if they provided both term and year
-                print(("ERROR: You can't use both --term and --year as arguments,"
-                       " only use one or the other."))
+            if options['year'] or options['recent']:
+                print("ERROR: Too many arguments!")
                 sys.exit(1)
 
             term = options['term']
             depts_terms = get_department_names([term])
 
-        else: # scrape all
-            # Use the specific term if it's given, otherwise get all terms like normal
-            terms = get_all_terms(options['year']) if options['year'] else get_all_terms()
+        else:
+            if options['year'] and options['recent']:
+                print("ERROR: Too many arguments!")
+                sys.exit(1)
+
+            terms = get_all_terms()
+            if options['year']:
+                terms = get_all_terms(options['year'])
+            elif options['recent']:
+                terms = determine_terms_to_scrape()
 
             depts_terms = get_department_names(terms)
 
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -234,6 +234,10 @@ def save_models(instructors: List[Instructor], sections: List[Section], # pylint
 
     start = time.time()
     with transaction.atomic():
+        # Our "bulk-update" method below will cascdate-delete the Grades, so collect them
+        # here to resave later. Note we're force-evaluating the QuerySet
+        # (by calling list()), as it would be empty if we evaluated it after due to the
+        # cascade-deletion.
         grades_to_resave = list(Grades.objects.filter(section__in=sections))
         print(f"Retrieved the grades models in {(time.time()-start):.2f}")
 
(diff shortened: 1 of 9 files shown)


gannonprudhomme:
Oh yeah I also need to make it so that we pass in the db as a environment variable so we don't track it

docker-compose.yml
```diff
@@ -29,6 +29,28 @@ services:
...
+      - SECRET_POSTGRES_USER
+      - SECRET_POSTGRES_PASS
+
+  cloud-sql-proxy:
+    image: gcr.io/cloudsql-docker/gce-proxy:1.11
+    # The 0.0.0.0 is key to having this work on Compute Engine
+    command: /cloud_sql_proxy -instances=$GCP_DB_NAME=tcp:0.0.0.0:3306
+    ports:
+      - 3306:3006
+
```
ryan-conn, line 53: Pretty minor but you should put this container before `scrape-courses` since it depends on `cloud-sql-proxy`

autoscheduler/scraper/tests/commands/scraper_utils_tests.py
```diff
@@ -0,0 +1,58 @@
...
+
+    def setUp(self):
+        self.strp_format = "%m/%d/%Y"
+        self.spring_2021 = ['20211']
+        self.summer_fall_2020 = ['20202', '20203']
+        self.summer_fall_2021 = ['20212', '20213']
+
+    def test_dec30_2020_is_spring_2021(self):
+        """ Tests that 11/30/2020 results in spring 2021 (20211) """
+        date = datetime.strptime('12/30/2020', self.strp_format)
```
ryan-conn, line 18: Up to you if you want to change it but I think using `date = datetime(year, month, day)` for these dates looks cleaner, and avoids having to specify the format

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -232,9 +234,12 @@ def save_models(instructors: List[Instructor], sections: List[Section], # pylint
 
     start = time.time()
     with transaction.atomic():
+        grades_to_resave = list(Grades.objects.filter(section__in=sections))
```
ryan-conn, line 237: idk why it took me so long to figure out what this was doing considering it's literally called `grades_to_resave`, but can you move/copy the comment about why we do this here instead of where they're resaved

ryan-conn (changes requested):
Some small things for you to address, for the most part this looks good

autoscheduler/scraper/tests/commands/scraper_utils_tests.py
gannonprudhomme: Yeah you're totally right I like this a lot better, will change it to this

autoscheduler/scraper/management/commands/scrape_courses.py
ryan-conn: Oh yeah also you should note that it calls `list()` so it's forced to evaluate the queryset here instead of later (when it'll be empty), since this is a little unintuitive

ryan-conn (approved):
Apologies I took so long to get to this, but I looked over all your changes again and it all looks good!

PR #342, opened 2020-09-09
Add Google Auth .env filling for CD

### LINKED ISSUE #314: Add .env loading for Google OAuth on deployment
Currently you can't do Google Oauth on the deployment b/c we don't track the `.env` files. We should create it in the deployment steps so it'll work.

gannonprudhomme:
## Description

Pretty straightforward - this adds the filling of the `.env` file for Google Auth.

## How to test

Per 6c7d749, I added a temp commit that makes the deployment run on PR so we can double check that this is actually working. As such, just go to http://revregistration.com and check that you can log in with Google!

Note that frankly this is a terrible way of testing this (cause I could break the deployment), but lol yolo

## Related tasks

Closes #314

### DIFF
--- a/.github/workflows/deploy-workflow.yml
+++ b/.github/workflows/deploy-workflow.yml
@@ -20,6 +20,9 @@ jobs:
       - name: Fill in Postgres credentials
         run: bash fill_credentials.sh ${{ secrets.POSTGRES_USER }} ${{ secrets.POSTGRES_PASS }}
 
+      - name: Fill .env Google Auth credentials
+        run: bash fill_google_auth_credentials.sh ${{ secrets.GOOGLE_CLIENT_ID }} ${{ secrets.GOOGLE_CLIENT_SECRET }}
+
       - name: Initialize Google Cloud SDK
         uses: zxyle/publish-gae-action@master
         with:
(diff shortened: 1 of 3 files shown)


gannonprudhomme:
@ThatJuanGuy Google OAuth is currently giving a 500 Server Error when trying to log in, any idea why?

gannonprudhomme:
Oh just kidding I got it - we haven't applied the migrations on the server yet, so it doesn't know to create all the tables need for OAuth and such. Fixing it now by connecting to it through cloud_sql_proxy and running `python manage.py migrate`.

gannonprudhomme:
I force pushed so you can check this on GCP right now

.github/workflows/deploy-workflow.yml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+  pull_request:
```
ryan-conn, line 6: Make sure to remove this

ryan-conn (changes requested):
Looks good, I think at some point in the future we'll want a `deployment` folder or something along those lines for docker files, shell scripts, etc. but that can be in a different PR after the coding club meeting so we can make sure nothing breaks

ryan-conn (approved):

PR #349, opened 2020-09-13
Add term select to navbar for schedule page

### LINKED ISSUE #289: Add current term select dropdown in schedule page nav bar
Should add a dropdown to the right of the nav bar title so the user can switch terms whenever they like (as well as not be confused as to which term they're on)

gannonprudhomme:
## Description

Pretty straightforward, this adds the SelectTerm to the navbar so it'll appear in the schedule page.

There's a few clean-up things I'd like to add (and also need to add some tests), but I wanted to get it out so we could discuss the look of it during the meeting.

## How to test

Select the term and make sure it navigates to the schedule page from the landing page. Also make sure it switches correctly when you're already on the schedule page.

## Screenshots

|Landing Page|Schedule Page|
|--|--|
|chrome_3l5DEqoVju|chrome_wMEyUwKNfF|

## Related tasks

Closes #289


gannonprudhomme:
I "fixed" the tests, but it's really a patch fix and I would like to replace my change with `fetchMock.mockIf('api/terms', async () => JSON.stringify({}));` but for some reason whenever I added that to the tests, the fetches that occur fail either with:

```
TypeError: Only absolute URLs are supported

      34 |   // Fetch all terms to use as ListItem options
      35 |   function getTerms(): void {
    > 36 |     fetch('api/terms').then((res) => res.json()).then(
```

or, when you add `fetchMock.mockIf('session/get_full_name', ....)` above the `api/terms` one, you'll get this:

```
autoscheduler\frontend\src\node_modules\react-dom\cjs\react-dom.development.js:302
      var evt = document.createEvent('Event'); // Keeps track of whether the user-provided callback threw an error. We
                         ^

TypeError: Cannot read property 'createEvent' of null
```

@firejake308 @ryan-conn if either of y'all could help me out with this I would greatly appreciate it - I've been dealing with this one error for the past hour and I'm miserable 😢. Also think once we nail it here we should be able to replace it in other places, which would be very nice.

gannonprudhomme:
Added tests for the new additions to `SelectTerm` and made it left-aligned. Should be ready for review now (still can't figure out `fetchMock.mockIf` though)

autoscheduler/frontend/src/tests/ui/SelectTerm.test.tsx
```diff
@@ -128,3 +129,52 @@ describe('SelectTerm', () => {
...
+    // component itself
+    expect(termThing).toHaveLength(2);
+  });
+
+  test('Shows "Select Term" when term is null', async () => {
+    mockTermsAPI(); // We don't care what this is, so use the function since we have it
+
+    // arrange
+    const store = createStore(autoSchedulerReducer);
+    // store.dispatch(setTerm());
```
ryan-conn, line 166: Remove this since it's not needed

autoscheduler/frontend/src/tests/ui/App.test.tsx
```diff
@@ -18,6 +18,10 @@ test('renders without errors', async () => {
   fetchMock.mockResponseOnce(JSON.stringify({}));
   // Mock response for SelectTerm component so that App renders correctly
   fetchMock.mockResponseOnce(JSON.stringify({}));
+  fetchMock.mockResponseOnce(JSON.stringify({})); // Mock api/terms, again
+
+  //  WHY DOES THIS BREAK THINGS
+  // fetchMock.mockIf('api/terms', async () => JSON.stringify({}));
```
ryan-conn, line 24: I tried making `mockIf` and `doMockIf` work for a while and whenever I used either it would crash the test suite. It would be really nice if we could get this to work, but it seems out of our control

autoscheduler/frontend/src/components/NavBar.tsx
```diff
@@ -39,29 +51,34 @@ const NavBar: React.SFC = () => {
         color="primary"
       >
         <Toolbar>
-          <div className={classes.title}>
```
ryan-conn, line 42: GitHub really mangled these changes, I compared it locally and you just added 5 lines but it looks like you completely reorganized the file here lmao

ryan-conn (changes requested):
I don't like how adding the second term select makes it so we have to fetch `api/terms` twice, the most efficient way to do this imo would be to move the term list to redux and have `App.tsx` dispatch an event to update the terms (I suppose the term select component in the navbar could do it too, since it's never unloaded). I'm also open to other solutions but if we did it that way then it would only have to fetch terms once per refresh

firejake308:
So, just so that I know where to pick up from tomorrow, and also to let y'all know how the progress is going, I'm going to leave a note of what I've figured out today:

The second error, where `document` is undefined occurs because some work was done after Jest had already torn down the DOM. This occurred because the fetch mock failed and a real `fetch` was called. Real `fetch` takes longer to resolve than most of our tests, so the `document` was already destroyed by the time the real `fetch` returned. This happened because `mockIf` will mock only the API call specified, and allow all other API calls to fall through to the real `fetch`. To fix this, we have 2 options:

1. Go to another library. We are currently using the more popular `jest-fetch-mock`, but there is another option called `fetch-mock-jest`. The API for this library looks a little cleaner imo, but it would be confusing to have 2 separate libraries for mocking fetch and annoying to convert our existing code. Would it be worth having cleaner code? See for yourself:
```
fetchMock
  .get('api/the-url-we-want', 200)
  .post('api/another-url', 400)
  .get(/bad\.com/, 500)
```
2. Stick with the current library. The recommended solution is something roughly like this:
```
fetchMock.mockIf(/^https?:\/\/example.com.*$/, req => {
      if (req.url.endsWith("/the-url-we-want")) {
        return "some response body"
      } else {
        return {} // who cares
      }
  })
```

ryan-conn:
Can you rebase this with master so we can see how this interacts with the new method for saving course cards? Other than that looks good, we talked about the things I commented on and your cleanup commit addressed the style issues I would have pointed out

gannonprudhomme:
Done

ryan-conn (changes requested):
Functionality looks good except for a bug I found:

When quickly changing terms, the redux calls to load saved cards can interleave with the calls that are loading section data for another term. The easiest way to fix this imo would be for the `setTerm` action to check if any course cards are loading before updating the term.

I also can't comment on it in the PR since you didn't change the line, but we should also change `selectedTerm` to a `useSelector` instead of being in the internal state of `SelectTerm`. Especially with my proposed change, they could become out of sync, and I don't see a reason not to use redux as the source of truth for the selected term.

gannonprudhomme:
> When quickly changing terms, the redux calls to load saved cards can interleave with the calls that are loading section data for another term. The easiest way to fix this imo would be for the setTerm action to check if any course cards are loading before updating the term.

What would it do if they're loading? Clear them?

> I also can't comment on it in the PR since you didn't change the line, but we should also change `selectedTerm` to a `useSelector` instead of being in the internal state of `SelectTerm`. Especially with my proposed change, they could become out of sync, and I don't see a reason not to use redux as the source of truth for the selected term.

Mind elaborating on how they could become out of sync? Since `selectedTerm` really is just the long-form of the term (e.g. "fall 2020 college station" vs "202031") I don't really see a need to also track it - it could really just be dependent on the term that's stored in redux no?

Honestly on that note, wtf is it even used for? It seem like it's only referenced for displaying what the "selected" dropdown option is

gannonprudhomme:
ok _now_ it should be good

gannonprudhomme:
ok _now_ it should be good (part 2)

ryan-conn:
I found a way that you can still break it, if you select another term while cards for the initial term is still loading it will overwrite the new term:
termselect

gannonprudhomme:
> I found a way that you can still break it, if you select another term while cards for the initial term is still loading it will overwrite the new term:

Can you describe what's happening here? I honestly can't tell

ryan-conn:
> Can you describe what's happening here? I honestly can't tell

Easiest way to replicate it:
- Refresh the page
- Select another term after the initial term has loaded but before cards have been fully loaded
- The cards for the original term will load in place of the ones for the new term

autoscheduler/scraper/management/commands/fill_terms.py
```diff
@@ -0,0 +1,27 @@
+from django.utils import timezone
+from django.core.management import base
+from scraper.models import Term
+from scraper.management.commands.utils.scraper_utils import get_all_terms
+
+class Command(base.BaseCommand):
```
ryan-conn, line 6: Not a change but I just wanted to comment here as a reminder to run this once this branch gets deployed

autoscheduler/frontend/src/components/LandingPage/LandingPage.tsx
```diff
@@ -1,15 +1,22 @@
...
+import { useDispatch } from 'react-redux';
 import HelpText from './HelpText/HelpText';
 import SelectTerm from './SelectTerm/SelectTerm';
 import * as styles from './LandingPage.css';
+import setTerm from '../../redux/actions/term';
+
+const LandingPage: React.FC<RouteComponentProps> = () => {
+  // Set the term to null so we don't display a term in the navbar when we're on this page
+  const dispatch = useDispatch();
+  dispatch(setTerm(null));
```
ryan-conn, line 12: It's not a big deal since it's functionally the same, but this would be better off as a `useEffect` with no dependencies since the intended behavior is to set the term to null whenever the user goes to the landing page (not every time it rerenders)

autoscheduler/frontend/src/components/LandingPage/SelectTerm/NavBarSelectTerm.css
```diff
@@ -0,0 +1,15 @@
+.button-container {
+    display: flex;
+    align-items: center;
+    padding-left: 1%;
```
ryan-conn, line 4: We don't use percentage-based padding anywhere else on the site, so why is it being used here? Also tbh I think this looks fine with no extra padding, but using a fixed pixel value would be fine:
1% padding | no padding
--|--
image | image

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
```diff
@@ -23,19 +29,38 @@ const getTermsJson = ((): () => Promise<any> => {
...
 
+  // Used for retrieving the user-friendly term phrase given the term code (e.g. "202031")
+  const [inverseTermMap, setInverseTermMap] = React.useState<Map<string, string>>(new Map());
+  const globalTerm = useSelector<RootState, string>((state) => state.termData.term);
+
+  let styles = defaultStyles;
+
+  if (navBar) {
+    styles = navBarStyles;
+  }
```
ryan-conn, line 48:
```suggestion
  const styles = navBar ? navBarStyles : defaultStyles;
```

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
```diff
@@ -23,19 +29,38 @@ const getTermsJson = ((): () => Promise<any> => {
...
     getTermsJson().then(
       (res) => {
-        const termsMap = new Map<string, string>(Object.entries(res));
+        const termsMap: Map<string, string> = new Map(Object.entries(res));
+        // Inverse the terms map so that it's in the format of [termNumber: termSentence],
+        // Ex: terms map is like ["Fall 2020 - College Station": 202031],
+        // and inverseTermsMap is [202031, "Fall 2020 - College Station"]
+        const inverseTermsMap: Map<string, string> = new Map(
+          //  a[1] (ex. "202031") is given as an "unknown", so we need to cast it to a string
+          Object.entries(res).map((a) => [String(a[1]), a[0]]),
```
ryan-conn, line 60:
```suggestion
          Array.from(termsMap.entries(), ([name, code]) => [code, name]),
```
A little more verbose, but it's more clear what this is going for

ryan-conn (changes requested):
Some nitpicks, I'm so glad this was done in parts instead of a single 2000 line PR

autoscheduler/frontend/src/components/LandingPage/SelectTerm/NavBarSelectTerm.css
gannonprudhomme: yeah percent padding is dumb idk why I did it, but I agree without padding is fine so I'll remove it

autoscheduler/frontend/src/components/LandingPage/LandingPage.tsx
```diff
@@ -1,15 +1,24 @@
...
 import SelectTerm from './SelectTerm/SelectTerm';
 import * as styles from './LandingPage.css';
+import setTerm from '../../redux/actions/term';
+
+const LandingPage: React.FC<RouteComponentProps> = () => {
+  // Set the term to null so we don't display a term in the navbar when we're on this page
+  const dispatch = useDispatch();
+  React.useEffect(() => {
+    dispatch(setTerm(null));
+  });
```
ryan-conn, line 14: `useEffect` will run on every rerender rather than only on mount unless you provide the second argument
```suggestion
  React.useEffect(() => {
    dispatch(setTerm(null));
  }, [dispatch]);
```

ryan-conn (changes requested):
One comment on the changes from my previous review but other than that this is fine

ryan-conn (approved):

PR #357, opened 2020-09-18
Optimize course card loading

### LINKED ISSUE #323: Improve course session performance
With the shared CPU for the database it can take up to 1 second for a cached CSCE 121 card to appear due to having to wait for the sections to load

As such, sessions shouldn’t have to wait for the sections to load in order for the course title to appear in the course card. Instead, fill in the course title like you would if you were entering it manually (and switch to section customization level) and have it show the loading icon for the loading sections

ryan-conn:
## Description

This PR moves `loading` for course cards into redux, and makes a couple of really nice changes that are now possible because of this change:
- Course names are loaded as soon as saved courses are fetched, and the loading indicator is displayed while it fetches the sections for each course
- It's now impossible (at least as far as I can tell) to overwrite your saved courses, since it checks if any course cards are loading before trying to save, and there should always be a course card loading from the time the page loads until either the fetch for saved courses fails, or all of the cards have been loaded.
I also added error handling to several places in the courseCards redux and course card column, I think that now it's robust enough that it shouldn't throw errors regardless of what any fetches return.

## Rationale

I moved course card loading to redux so that I could abstract away loading from the component itself (and let redux deal with all the logic), and also use it outside of individual course cards. I had to reorganize a fair bit of the redux to do this, but I think it's just as clean/cleaner than before.

## How to test

Try closing the page or switching terms whenever you want, and I think that there's no way it will overwrite your course cards until they're fully loaded, also look at how you can see loading indicators while course cards are initially loading

## Screenshots

image

## Related tasks

Closes #323.

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
@@ -29,11 +29,10 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
 
   const term = useSelector<RootState, string>((state) => state.term);
   const dispatch = useDispatch();
-  const { course, customizationLevel, sections } = courseCardOptions;
+  const { course, customizationLevel, loading } = courseCardOptions;
 
   const [options, setOptions] = React.useState([]);
   const [inputValue, setInputValue] = React.useState('');
-  const [loading, setLoading] = React.useState(false);
 
   function getAutocomplete(text: string): void {
     fetch(`api/course/search?search=${text}&term=${term}`).then(
@@ -73,7 +72,6 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
       </div>
     );
   }
-  React.useEffect(() => setLoading(false), [sections]);
 
   return (
     <Card>
@@ -120,7 +118,6 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
             if (options.length === 0) setInputValue('');
           }}
           onChange={(_evt: object, val: string): void => {
-            if (val) setLoading(true);
             dispatch(updateCourseCard(id, {
               course: val,
             }, term));
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
@@ -107,6 +107,7 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
           size="small"
           autoHighlight
           autoSelect
+          disabled={loading}
           inputValue={inputValue}
           value={course}
           multiple={false}
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
@@ -44,34 +44,32 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
     });
   }
 
-  // determine customization content based on customization level
-  let customizationContent: JSX.Element = null;
-  switch (customizationLevel) {
-    case CustomizationLevel.BASIC:
-      customizationContent = <BasicSelect id={id} />;
-      break;
-    case CustomizationLevel.SECTION:
-      customizationContent = course
-        ? <SectionSelect id={id} />
-        : (
-          <Typography className={styles.grayText}>
-            Select a course to show available sections
-          </Typography>
-        );
-      break;
-    default:
-      customizationContent = null;
-  }
-
-  // show loading if we're not sure what sections are available
-  if (loading) {
-    customizationContent = (
-      <div id={styles.centerProgress}>
-        <SmallFastProgress />
-        Loading sections...
-      </div>
-    );
-  }
+  // determine customization content based on whether the card is loading and customization level
+  const customizationContent = React.useMemo(() => {
+    // show loading if we're not sure what sections are available
+    if (loading) {
+      return (
+        <div id={styles.centerProgress}>
+          <SmallFastProgress />
+          Loading sections...
+        </div>
+      );
+    }
+    switch (customizationLevel) {
+      case CustomizationLevel.BASIC:
+        return <BasicSelect id={id} />;
+      case CustomizationLevel.SECTION:
+        return course
+          ? <SectionSelect id={id} />
+          : (
+            <Typography className={styles.grayText}>
+              Select a course to show available sections
+            </Typography>
+          );
+      default:
+        return null;
+    }
+  }, [course, customizationLevel, id, loading]);
 
   return (
     <Card>
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
@@ -26,7 +26,7 @@ const CourseSelectColumn: React.FC = () => {
     if (term) {
       fetch(`sessions/get_saved_courses?term=${term}`).then((res) => (
         res.json()
-      )).then((courses: SerializedCourseCardOptions[]) => {
+      )).catch(() => []).then((courses: SerializedCourseCardOptions[]) => {
         dispatch(replaceCourseCards(courses, term));
       });
     }
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
@@ -7,7 +7,10 @@ import { RootState } from '../../../redux/reducer';
 import { CourseCardArray, SerializedCourseCardOptions } from '../../../types/CourseCardOptions';
 import CourseSelectCard from './CourseSelectCard/CourseSelectCard';
 import { addCourseCard, replaceCourseCards, clearCourseCards } from '../../../redux/actions/courseCards';
-import throttle from '../../../utils/throttle';
+import createThrottleFunction from '../../../utils/createThrottleFunction';
+
+// Creates a throttle function that shares state between calls
+const throttle = createThrottleFunction();
 
 /**
  * Renders a column of CourseSelectCards, as well as a button to add course cards
@@ -75,9 +78,8 @@ const CourseSelectColumn: React.FC = () => {
       });
     };
 
-    throttle(`saveCourseCards${term}`, saveCourses, 15000, true);
+    throttle(`${term}`, saveCourses, 15000, true);
   }, [courseCards, term]);
-
 
   const rows: JSX.Element[] = [];
 
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
@@ -42,6 +42,11 @@ const CourseSelectColumn: React.FC = () => {
   React.useEffect(() => {
     if (!term) return;
 
+    // if any course cards are loading, don't try to save
+    for (let i = 0; i < courseCards.numCardsCreated; i++) {
+      if (courseCards[i]?.loading) return;
+    }
+
     const saveCourses = (): void => {
       // Serialize courseCards and make API call
       const courses: SerializedCourseCardOptions[] = [];
--- a/autoscheduler/frontend/src/redux/actions/courseCards.ts
+++ b/autoscheduler/frontend/src/redux/actions/courseCards.ts
@@ -172,7 +172,8 @@ function sortSections(sections: SectionSelected[]): SectionSelected[] {
 }
 
 /**
- * Fetches sections for course in courseCard, then updates courseCard with new sections
+ * Fetches sections for course in courseCard, then updates courseCard with new sections.
+ * If the course card is valid or sections can't be fetched, returns undefined.
  * @param courseCard course to get sections for
  * @param term term to fetch sections for
  */
@@ -180,7 +181,7 @@ async function fetchCourseCardFrom(
   courseCard: CourseCardOptions | SerializedCourseCardOptions,
   term: string,
 ): Promise<CourseCardOptions> {
-  const [subject, courseNum] = courseCard.course.split(' ');
+  const [subject, courseNum] = courseCard.course?.split(' ') || ['', ''];
   return fetch(`/api/sections?dept=${subject}&course_num=${courseNum}&term=${term}`)
     .then((res) => res.json())
     .then(parseSectionSelected)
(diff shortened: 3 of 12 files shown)


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -26,7 +26,7 @@ const CourseSelectColumn: React.FC = () => {
     if (term) {
       fetch(`sessions/get_saved_courses?term=${term}`).then((res) => (
         res.json()
-      )).then((courses: SerializedCourseCardOptions[]) => {
+      )).catch(() => []).then((courses: SerializedCourseCardOptions[]) => {
```
gannonprudhomme, line 29: How come the catch is here? Should it not be after the second `.then`?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -42,6 +42,11 @@ const CourseSelectColumn: React.FC = () => {
   React.useEffect(() => {
     if (!term) return;
 
+    // if any course cards are loading, don't try to save
+    for (let i = 0; i < courseCards.numCardsCreated; i++) {
+      if (courseCards[i]?.loading) return;
+    }
```
gannonprudhomme, line 48: Do you think this would be better with a filter? I guess probably not, since a filter would have you go through all course cards, whereas this stops as soon as you get one that's loading, but thought I would mention it

autoscheduler/frontend/src/utils/throttle.ts
```diff
@@ -45,11 +45,12 @@ const throttle = ((): (id: string, callback: any, delay: number, runOtherIDs: bo
...
 
-    // Update function to run when interval completes
-    functions.set(id, callback);
-
     // Update beforeunload event listener so callback is fired before close/refresh
+    window.removeEventListener('beforeunload', functions.get(id));
     window.addEventListener('beforeunload', callback);
+
+    // Update function to run when interval completes
+    functions.set(id, callback);
```
gannonprudhomme, line 53: Mind explaining how this "fixes the incorrect removal of event listener in throttle?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -193,8 +194,10 @@ async function fetchCourseCardFrom(
         sections,
         hasHonors,
         hasWeb,
+        loading: false,
       };
-    });
+    })
+    .catch(() => undefined);
```
gannonprudhomme, line 200: What's this `.catch(() => undefined);`?

gannonprudhomme (changes requested):
Haven't done an entire pass of the code yet, but wanted to mention some bugs I ran into that are pretty important to be covered I think

First, I believe our course card (or at least SectionSelect) performance has regressed. If you spam click a section in the SectionSelect, it doesn't respond quickly like it does on master/GCP. Note that while I have the mouse over the section, I am constantly clicking.

|Currently|On GCP|
|--|--|
|spam-right-click-bug|spam-right-click-working|

Also, it seems like the course cards (that are finished loading) behave weird when the other cards are also loading. Not sure how related this is to the above issue

reload-section-select

Overall great job with this! I'm super excited to get this out, as I really think this has a huge impact in how responsive sessions feel (thus making our website feel better as a whole). In terms of improving the course card session performance, this is exactly what I had in mind so you really nailed it in that regard.

gannonprudhomme:
Also, related to that second gif I sent above, if you spam click on a section select, I think it costs a bunch of performance on the website as a whole, as it also makes other parts of the page (generate schedules in this case) perform worse. 

In this gif, while I have my mouse over SectionSelect or "Generate Schedules" I am spamming click. Note how long it takes for Generate Schedules to respond

spam-performance

autoscheduler/frontend/src/utils/throttle.ts
ryan-conn: I was setting the new function before the `removeEventListener` call before (which needs you to use the same callback as the event listener was added with), so it wasn't actually removing it

ryan-conn (commented):
> First, I believe our course card (or at least SectionSelect) performance has regressed. If you spam click a section in the SectionSelect, it doesn't respond quickly like it does on master/GCP. Note that while I have the mouse over the section, I am constantly clicking.

Yeah I agree, I was too lazy to restart my computer yesterday even though it was stuck at 100% ram usage so I assumed that was why it wasn't performing well. I guess now I have to dive more into javascript async hell to find out what exactly is taking so long

> Also, it seems like the course cards (that are finished loading) behave weird when the other cards are also loading. Here I refresh the page and try to click a section for Math 151. Not sure how related this is to the above issue

It's more than likely that this is the case, I'll be looking into it more over the next couple days

also idk why it's putting all my comments as review but I'm too lazy to retype them

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
gannonprudhomme: Oh yeah great point, I forgot the weirdness of our courseCard type setup

ryan-conn:
I made some changes to improve performance (and also made the loading/not loading logic more airtight). One thing I'd like to point out is that MATH 151 in particular has very bad performance even on GCP since it has so many sections (you were comparing CSCE 121 on GCP to MATH 151 on my branch), and when I went on GCP to test performance, it was very similar to this branch:
| GCP | This branch |
|-|-|
|gcp | branch |

and CSCE 121 runs fine on this branch:
csce
So I propose that we handle any more performance optimizations in a different PR, since it hasn't really changed all that much. When I was looking at performance, I noticed that rerendering `<SectionSelect>` was by far the most costly operation both when loading saved courses and toggling selection of a section (since it has to reprocess every single section), so we should think of a way to avoid rerendering every single section when one is toggled.

firejake308 (commented):
Just reading the code, I have a couple of questions about how some of it works. I'll try to run it some time later

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
```diff
@@ -45,35 +44,32 @@ const ExpandedCourseCard: React.FC<ExpandedCourseCardProps> = ({
...
-    customizationContent = (
-      <div id={styles.centerProgress}>
-        <SmallFastProgress />
-        Loading sections...
-      </div>
-    );
-  }
-  React.useEffect(() => setLoading(false), [sections]);
+  // determine customization content based on whether the card is loading and customization level
+  const customizationContent = React.useMemo(() => {
```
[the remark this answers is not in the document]
ryan-conn: I definitely noticed that it's faster now, but honestly I think that's mostly due to me checking if the card is loading, rather than from the `useMemo`, since now that I look at this again it will have to recalculate `customizationContent` most of the time when a variable changes. I'll try to think of a more scientific method of whether the memoization is helping and post my results here when I do.

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -205,12 +207,15 @@ async function fetchCourseCardFrom(
...
-): ThunkAction<void, RootState, undefined, UpdateCourseAction> {
-  return (dispatch): void => {
+): ThunkAction<Promise<void>, RootState, undefined, UpdateCourseAction> {
+  return (dispatch): Promise<void> => new Promise((resolve) => {
     fetchCourseCardFrom(courseCard, term).then((updatedCourseCard) => {
-      dispatch(updateCourseCardSync(index, updatedCourseCard));
+      if (courseCard) {
+        dispatch(updateCourseCardSync(index, updatedCourseCard));
+        resolve();
+      }
```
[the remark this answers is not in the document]
ryan-conn: It just keeps the card how it is without updating section data, since there's nothing to base it off of

autoscheduler/frontend/src/redux/reducers/courseCards.ts
```diff
@@ -66,6 +67,7 @@ export default function courseCards(
       return {
         ...state,
         [action.index]: { ...state[action.index], ...action.courseCard },
+        numCardsCreated: Math.max(state.numCardsCreated, action.index + 1),
```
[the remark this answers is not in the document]
ryan-conn: I used `updateCourseCardSync` when creating cards from serialized save cards since `addCourseCard` seems a bit unintuitive to use when updating the first card rather than adding it, but the same could be said for calling `updateCourseCard` for cards that don't exist. We can use `addCourseCard` instead here without changing any code, so I could change it to that if you'd prefer (and remove this line since it would no longer be needed)

autoscheduler/frontend/src/tests/ui/CourseSelectCard.test.tsx
```diff
@@ -561,6 +562,8 @@ describe('Course Select Card UI', () => {
         const { getByText } = render(
           <Provider store={store}><CourseSelectCard id={0} /></Provider>,
         );
+        // Stop course card from loading
+        store.dispatch<any>(updateCourseCard(0, { loading: false }));
```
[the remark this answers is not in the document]
ryan-conn: It would attempt to fetch saved sections in `<CourseSelectColumn>`, find an empty array, and dispatch `replaceCourseCards` with an empty array which would set the initial course card to stop loading. Since these tests don't actually have the `<CourseSelectColumn>` component to tell the card to stop loading, this is the only way I could think of to do this other than overwriting the default redux store.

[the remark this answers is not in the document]
firejake308: Okay, makes sense

autoscheduler/frontend/src/redux/actions/courseCards.ts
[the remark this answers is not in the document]
ryan-conn: Oh yeah you're totally right, my bad

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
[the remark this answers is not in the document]
ryan-conn: So I've done some testing for this now, as a "benchmark" I looked at the react profiler with/without the `useMemo` and did 2 things:
1. Made a bunch of cards with good ol' MATH 151, reloaded, and looked at render times whenever each card loaded. With the `useMemo` they averaged around 200ms, and without it was around 400ms.
2. Typed in MATH 151 and clicked on the result after it loaded. With `useMemo` it took around 400ms to render, without it divided the render into 3 events that were around 600, 1200, and 600 ms
So this is definitely a good use case for `useMemo`

ryan-conn:
Comments addressed (except for the `updateCourseCard` change, I left it as-is since Adel hasn't responded yet)

As I said in discord, I think it would also be a good idea to disable the `<Autocorrect>` component for a course card while it's loading, but I didn't make that change yet. Let me know what you guys think about it

gannonprudhomme:
This looks good to me, but I'm gonna let frontend wizard @firejake308 have the final approval

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.tsx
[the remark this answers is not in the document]
firejake308: 美しい

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -70,10 +78,9 @@ const CourseSelectColumn: React.FC = () => {
       });
     };
 
-    throttle(`saveCourseCards${term}`, saveCourses, 15000, true);
+    throttle(`${term}`, saveCourses, 15000, true);
```
firejake308, line 81: What's your reasoning for removing the `saveCourseCards` in the key? Are there no other functions that use `throttle`?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -172,15 +172,16 @@ function sortSections(sections: SectionSelected[]): SectionSelected[] {
 }
 
 /**
- * Fetches sections for course in courseCard, then updates courseCard with new sections
+ * Fetches sections for course in courseCard, then updates courseCard with new sections.
+ * If the course card is valid or sections can't be fetched, returns undefined.
```
firejake308, line 176: Did you mean `invalid` here?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -274,17 +286,32 @@ export function replaceCourseCards(
...
+    const deserializedCards: CourseCardOptions[] = [];
     courseCards.forEach((courseCard, idx) => {
-      updateSectionsForCourseCard(courseCard, term).then((updatedCard) => {
-        dispatch(addCourseCard(updatedCard, idx));
+      const deserializedCard = deserializeCourseCard(courseCard);
+      deserializedCards.push(deserializedCard);
+      dispatch(updateCourseCardSync(idx, deserializedCard));
+    });
+
+    // all course cards are now loading (preventing courses from being overwritten
```
firejake308, line 304:
```suggestion
    // all course cards are now marked as loading (preventing courses from being overwritten
```

slight clarification that would help me understand it if I ever came back to this code

autoscheduler/frontend/src/redux/reducers/courseCards.ts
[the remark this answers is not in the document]
firejake308: No, updateSync is fine

autoscheduler/frontend/src/utils/createThrottleFunction.ts
```diff
@@ -27,7 +31,7 @@ const throttle = ((): (id: string, callback: any, delay: number, runOtherIDs: bo
         const toRun = functions.get(id);
         if (typeof toRun === 'function') toRun();
         functions.delete(id);
-        window.removeEventListener('beforeunload', functions.get(id));
+        window.removeEventListener('beforeunload', toRun);
```
firejake308, line 34: Nice, so before you fixed this, it used to attempt to get the function that you deleted in the line before?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
firejake308: Wait, so I read the rest of the code, and I think what happened is that your nre `createThrottleFunction` gives a separate state to each file that imports and calls it. Therefore, if you used this function in a new file, it would have a different state, so there would be no possibility for conflict with this usage here. Did I understand that correctly?

firejake308 (changes requested):
Feels good in browser. I had like a couple of suggestions for better comments and some questions for my own understanding, but this is pretty close to finished imo

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
ryan-conn: Yeah that's correct

autoscheduler/frontend/src/utils/createThrottleFunction.ts
ryan-conn: Correct

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: Yes, fixed

ryan-conn:
I made the changes you suggested to comments, and also updated the formatting of the courseCards reducer redux since I just noticed it was really weird before

firejake308 (approved):

firejake308:
Yeah, sometimes the formatting is weird when VSCode auto-imports a lot of things from one module

PR #361, opened 2020-09-21
Changed web default to no preference

### LINKED ISSUE #345: Change basic options default value from "exclude" to "no preference"
We should make it start with the one that's the least restrictive imo. It's also sort of weird to by-default exclude web-only classes, when that's like almost every class now

gannonprudhomme:
## Related tasks

Closes #345

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicOptionRow.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicOptionRow.tsx
@@ -16,7 +16,7 @@ interface BasicOptionRowProps {
  * option selected by this row, formatted as it is found in the Redux course cards
  */
 const BasicOptionRow: React.FC<BasicOptionRowProps> = ({ id, value }) => {
-  const option = useSelector<RootState, string>((state) => state.courseCards[id][value] || 'exclude');
+  const option = useSelector<RootState, string>((state) => state.courseCards[id][value] || 'no_preference');
   const dispatch = useDispatch();
 
   return (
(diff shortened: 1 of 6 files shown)


firejake308:
Can you also change the defaults in `autoscheduler/scheduler/utils.py`?

gannonprudhomme:
Done!

gannonprudhomme:
Ok it should be good to go now

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicOptionRow.tsx
```diff
@@ -16,7 +16,7 @@ interface BasicOptionRowProps {
  * option selected by this row, formatted as it is found in the Redux course cards
  */
 const BasicOptionRow: React.FC<BasicOptionRowProps> = ({ id, value }) => {
-  const option = useSelector<RootState, string>((state) => state.courseCards[id][value] || 'exclude');
+  const option = useSelector<RootState, string>((state) => state.courseCards[id][value] || 'no_preference');
```
ryan-conn, line 19: This changes the default for honors as well, can you instead make it so the default is `no_preference` when a card is created with `addCourseCard`?

ryan-conn (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicOptionRow.tsx
gannonprudhomme: Oh yeah duh, this was the suggestion Adel gave. I was wondering why I hadn't done it previously when he suggested and this was the reason why.

ryan-conn (approved):
Third time's the charm lol, looks good

PR #369, opened 2020-09-24
Fix Z index for dialog and availability card

firejake308:
## Description

Fix Z index for dialog and availability card

## Rationale

I also had to update the CSS `.d.ts` files built by our CSS loader, most likely due to an updating the package back in #351 

## How to test

To test dialog, clear local storage. Reload. Add JAPN 201 for term Fall 2020. Include Full sections. Generate. Availability instructions dialog should cover classes.

To test availability cards, generate any schedule. Hover over meeting time in schedule. Move cursor to the start time label. Click and begin dragging. Stop dragging at end time label. Availability should be covering classes the entire time.

### DIFF
--- a/autoscheduler/frontend/src/components/App/App.css.d.ts
+++ b/autoscheduler/frontend/src/components/App/App.css.d.ts
@@ -1,13 +1,13 @@
-declare namespace AppCssModule {
+declare namespace AppCssNamespace {
   export interface IAppCss {
     "app-container": string;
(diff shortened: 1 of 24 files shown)


ryan-conn (approved):
Not much to comment on, I made sure the fixes worked and confirmed that the `.d.ts` files changing is due to upgrading our packages (they were being compiled the old way until I ran `npm i`). Good job

PR #380, opened 2020-09-30
Memoized SectionInfo component

### LINKED ISSUE #360: Improve SectionSelect performance
As discovered in #357, when courses have a lot of sections (i.e. MATH-151) it can cause the selecting of the sections perform pretty slowly

Gif of the problem. Note that I'm spamming click anytime I'm hovering over the sections

firejake308:
## Description

Spam-clicking a section in `SectionSelect` currently results in a really uncomfortable lag period where the UI is frozen. I originally attempted to fix this by virtualizing the list, but virtualization results ina an uncomfortable lag period where the `SectionSelect` is blank. I have found a better solution in memoization, which is basically the functional component equivalent of `shouldComponentUpdate`. Specifically, I moved the display for each section out of `SectionSelect` and into a new component called `SectionSelect`. Next, I memo-ized this `SectionInfo`, instructing it not to re-render unless it has been selected/de-selected. Significantly, this makes the assumption that none of the other fields or sub-fields in `SectionSelected` will ever change without also completely re-rendering the course card.

## Rationale

I had to make a new `ThunkAction` to toggle the selected state of each section due to some really weird behavior with JavaScript closures. If anyone thinks they understand closures, please explain how they work because apparently I don't.

## How to test

Spam-click a section. Should update quickly. Now try selecting multiple sections. Selecting a new section shouldn't un-select your previous choices.

## Screenshots

memo

## Related tasks

Closes #360

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { List, Typography } from '@material-ui/core';
-import { useSelector, useDispatch } from 'react-redux';
+import { useSelector, useDispatch, shallowEqual } from 'react-redux';
 import { SectionSelected } from '../../../../../../types/CourseCardOptions';
 import { RootState } from '../../../../../../redux/reducer';
 import * as styles from './SectionSelect.css';
@@ -13,19 +13,8 @@ interface SectionSelectProps {
 
 const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
   const sections = useSelector<RootState, SectionSelected[]>(
-    (state) => state.courseCards[id].sections,
+    (state) => state.courseCards[id].sections, (left, right) => left === right,
   );
-  const dispatch = useDispatch();
-
-  const toggleSelected = React.useCallback((i: number): void => {
-    dispatch(updateCourseCard(id, {
-      sections: sections.map((sec, idx) => (idx !== i ? sec : {
-        section: sec.section,
-        selected: !sec.selected,
-        meetings: sec.meetings,
-      })),
-    }));
-  }, [dispatch, id, sections]);
 
   // show placeholder text if there are no sections
   if (sections.length === 0) {
@@ -49,9 +38,9 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
       return (
         <SectionInfo
           secIdx={secIdx}
+          courseCardId={id}
           sectionData={sectionData}
           addInstructorLabel={makeNewGroup}
-          toggleSelected={toggleSelected}
           key={sectionData.section.id}
         />
       );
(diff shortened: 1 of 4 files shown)


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -33,174 +24,24 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
...
-        <td>{formatMeetingDays(mtg)}</td>
-        <td>{getMeetingTimeText(mtg)}</td>
-      </Typography>
-    </React.Fragment>
-  );
-
   const makeList = (): JSX.Element[] => {
     let lastProf: string = null;
     let lastHonors = false;
-    return sections.map(({ section, selected, meetings }, secIdx) => {
```
ryan-conn, line 132: I'm so glad this is a separate component now, no offense but this was an absolute mess before lol

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -1,28 +1,19 @@
...
+import SectionInfo from './SectionInfo';
 
 interface SectionSelectProps {
   id: number;
 }
 
 const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
   const sections = useSelector<RootState, SectionSelected[]>(
-    (state) => state.courseCards[id].sections,
+    (state) => state.courseCards[id].sections, (left, right) => left === right,
```
ryan-conn, line 15: I thought the default equality function checked reference equality, why did you explicitly do it here?

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -239,6 +239,22 @@ export function updateCourseCard(index: number, courseCard: CourseCardOptions, t
   };
 }
 
+export function toggleSelected(courseCardId: number, secIdx: number):
```
ryan-conn, line 242: I'm fine with you moving this here since we don't need a selector in the `SectionInfo` component, and it abstracts out the logic a bit from the component itself

ryan-conn (changes requested):
Couple of small things, great job making this a lot more responsive (and also making the sectionselect a lot more readable)! 

One thing I noticed is that it's still pretty slow when you switch between basic/section select since it has to rerender everything, it would be really nice if we could memoize it as well somehow (although I'm not sure we would do so right now). Definitely don't need to do that in this PR, but we should keep that in mind so that in the future we could potentially make it even faster.

Also I think we should wait for #384 to be merged before merging this so Shaham doesn't have to deal with rebasing his branch

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
firejake308: Nice catch. I think I put that in while debugging to see if it would change anything, and understandably, it didn't. Removed in latest commit.

gannonprudhomme (approved):
looks good to me!

PR #382, opened 2020-09-30
Database access optimization for /api/sections

ryan-conn:
## Description
Prefetches instructors and meetings in `/api/sections`, which solves the [N+1 selects problem](https://stackoverflow.com/questions/97197/what-is-the-n1-selects-problem-in-orm-object-relational-mapping) that made us use far more queries than necessary to fetch section data.

## Rationale
I also reorganized some parts of the section serializer to be easier to read (rename `obj` to `section`, ternary for instructor name), and added `related_name='meetings'` to meetings so that we can get meetings for a section with `section.meetings`.

Note that getting grades still takes a query for each section, I couldn't figure out a way to optimize it since `instructor_performance` performs a query and with the way that serializers work in DRF, the only way to stop this is with caching (which should be done in a different PR). Fixing this should dramatically improve performance.

## How to test
Just fetch sections and you'll see that it's ~10-20% faster now, here are some more concrete examples:
KINE 199 (course with the most sections)
| Before | After |
|-|-|
| image | image |

CSCE 121 (decent number of sections)
| Before | After |
|-|-|
| image | image |

## Related tasks

No issue for this, I just noticed that we could be performing queries more efficiently when I was working on APIs for a couple of my classes

### DIFF
--- a/autoscheduler/scraper/models/section.py
+++ b/autoscheduler/scraper/models/section.py
@@ -41,7 +41,11 @@ class Meeting(models.Model):
     start_time = models.TimeField(null=True)
     end_time = models.TimeField(null=True)
     meeting_type = models.CharField(max_length=4) # Meeting types: LEC, LAB, REC, INS, etc
-    section = models.ForeignKey(Section, on_delete=models.CASCADE)
+    section = models.ForeignKey(
+        Section,
+        on_delete=models.CASCADE,
+        related_name='meetings'
+    )
 
     class Meta:
         db_table = "meetings"
--- a/autoscheduler/scraper/serializers.py
+++ b/autoscheduler/scraper/serializers.py
@@ -1,6 +1,6 @@
 from datetime import time
 from rest_framework import serializers
-from scraper.models import Course, Section, Meeting, Department, Grades
+from scraper.models import Course, Section, Department, Grades
 
 def format_time(time_obj: time) -> str:
     """ Formats a time object to a string HH:MM, for use with section serializer """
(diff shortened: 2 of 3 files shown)


gannonprudhomme:
Mind showing how many queries this decreased? Given that access to the database is consistently slower with GCP (namely due to the shared CPU) I expect there to be a larger improvement in the regard. In fact, would you also mind using `cloud-sql-proxy` and running `loadtest` when connected to GCP (if you do `export SETTINGS_MODE=proxy` before you run `runsever` when connected to `cloud-sql-proxy`, any requests to `localhost:8000` will use the GCP DB). I originally would've have expected this to be a larger improvement based on our convo's, so I'm hoping it'll be more noticeable when connected to GCP.

firejake308:
I'll let Ryan give the definitive answer, but based on my understanding, the old method runs the following queries:
```
SELECT * FROM sections;
SELECT * FROM meetings WHERE section_id = 0;
SELECT * FROM meetings WHERE section_id = 1;
SELECT * FROM meetings WHERE section_id = 2;
...
```
Whereas the new `select_related` runs the following query:
```
SELECT * FROM sections JOIN instructors ON sections.instructor_name = instructors.name;
```
And the new `prefetch_related` runs the following queries:
```
SELECT * FROM sections;
SELECT * FROM meetings WHERE meetings.section_id IN (0, 1, 2, ...);
```

As I said, though, I'll let Ryan tell you whether or not I understood it correctly.

ryan-conn:
> I'll let Ryan give the definitive answer, but based on my understanding, the old method runs the following queries:
> ```
> SELECT * FROM sections;
> SELECT * FROM meetings WHERE section_id = 0;
> SELECT * FROM meetings WHERE section_id = 1;
> SELECT * FROM meetings WHERE section_id = 2;
> ...
> ```
> Whereas the new `select_related` runs the following query:
> ```
> SELECT * FROM sections JOIN instructors ON sections.instructor_name = instructors.name;
> ```
> And the new `prefetch_related` runs the following queries:
> ```
> SELECT * FROM sections;
> SELECT * FROM meetings WHERE meetings.section_id IN (0, 1, 2, ...);
> ```
> 
> As I said, though, I'll let Ryan tell you whether or not I understood it correctly.
> 

Yeah, it used to do 1 query for all sections, and then 3 extra queries per section for grades, instructors, and meetings. Now it does 1 query that gets all sections and grades, another 1 query for all meetings, and a query per section for grades.

I'll test the performance on GCP either later today or tomorrow, I'm not at home right now

gannonprudhomme:
You have those GCP numbers?

gannonprudhomme:
|Section|This PR 50%|This PR 100%|Default 50%| Default 100%|
|--|--|--|--|--|
|KINE 199|7939 ms|8563 ms|21753ms|29685 ms|
|CSCE 121|1467 ms|1611 ms|3503 ms|3839 ms|

It really surprised me how much of an improvement this was (and how slow it is normally). We typically don't get worse than 8000 ms on revregistration.com for KINE 199 so I think a factor of why it's not this slow on GCP to begin with is that App Engine connecting to Cloud SQL is significantly faster than us connecting to Cloud SQL. Regardless I def think this is a big improvement.

gannonprudhomme (approved):

PR #383, opened 2020-10-03
Add SSL redirection

### LINKED ISSUE #335: Add SSL Certificate to domain so we can use https
- It currently says "not secure", presumably b/c it doesn't use https - need to make it not say this

ryan-conn:
## Description

Makes app engine requests to the `http` version of the website redirection to `https`

## Rationale

I tried finding a way to do this with google domains, but you can't use DNS records to both provide redirection and use app engine to serve pages, so this seems like the only way to do it

## How to test
I made it deploy a version of the site that redirects (will make sure the workflow change gets reverted) so we can test it out before merging into master

## Screenshots
sexy
redirecto!

## Related tasks

Closes #335.

### DIFF
--- a/.github/workflows/deploy-workflow.yml
+++ b/.github/workflows/deploy-workflow.yml
@@ -1,8 +1,5 @@
 name: GCP Deploy
-on:
-  push:
-    branches:
-      - master
+on: push
 
 jobs:
   deploy-app-engine:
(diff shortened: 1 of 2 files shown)


gannonprudhomme (approved):
Very nice, thanks for this!

PR #385, opened 2020-10-05
Ordering of meetings

### LINKED ISSUE #359: Ordering of meetings
As you can see in the image below, sometimes, lectures are shown before labs, and other times, labs are shown before lectures. I feel like it would be better to have a consistent order for meetings (probably lecture before lab). What do y'all think?

image

ajayramsunder:
## Description
This pull request holds the sort meeting function that was put in the utils folder which allows this function to be used in different places.  
## Rationale
This encompasses a need and is a simple function that just orders the meetings based on a hierarchy that was given.
## How to test
If a course is added to the left bar, the Lecture meeting is first consistently and everything else is laid out directly underneath it in a certain order. The meetings are not all over the place now.
## Screenshots
Before: 
image

After: 
image

## Related tasks
Closes #359


.idea/.gitignore
```diff
@@ -0,0 +1,3 @@
+
+# Default ignored files
+/workspace.xml
```
firejake308, line 3: Can you add `.idea` to your `.git/info/exclude`?

firejake308 (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -13,6 +13,7 @@ import { RootState } from '../../../../../../redux/reducer';
 import * as styles from './SectionSelect.css';
 import { updateCourseCard } from '../../../../../../redux/actions/courseCards';
 import GradeDist from './GradeDist/GradeDist';
+import sortMeeting from '../../../../../../utils/sortMeetingFunction';
```
firejake308, line 16: I think this would make more sense in the `Meeting.ts` folder. Thoughts @ryan-conn ?

ryan-conn: You mean the `Meeting.ts` file? Yeah I agree, I actually told him to put it in `utils` but since it relates specifically to meetings it makes sense to put them together

config/postgres_password
```diff
@@ -1 +1 @@
-postgres
```
gannonprudhomme, line 1: Hopefully you got it figured out, but this actually isn't where you're supposed to change the postgres information - it should be in `autoscheduler/autoscheduler/config/postgres-info.json`. As such, could you remove these changes? This is also the reason that the backend tests are failing - these files are used to initialize the postgres DB in our docker container that runs the backend tests.

ajayramsunder: Yeah I made the change sorry about that

autoscheduler/frontend/src/utils/sortMeetingFunction.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+// sorts meeetings based on meeting type
+const sortMeeting = (a: Meeting, b: Meeting): number => {
+  if (a.meetingType > b.meetingType) {
+    return 1;
+  }
+  if (a.meetingType < b.meetingType) {
+    return -1;
+  }
+  return 0;
```
gannonprudhomme, line 11: I _think_ this can be done with [string.localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare), so you should be able to change this to `return a.meetingType.localeCompare(b.meetingType)`. Might have to flip a & b though. Other than this I think it'll be good to go

ryan-conn: Actually I think the way he has it is the way we should be doing this. This is comparing the enum values rather than the string values, which means we can reorder the enum values in order to define the order they show up in, and let important meetings like LEC/LAB show up before EXAM meetings. It's also a lot more flexible because if we decide to change it later we can just reorder the enum

firejake308: Why not just return `a.meetingType - b.meetingType`?

ryan-conn:
Can you resolve merge conflicts and remove the `.idea` folder like Adel suggested? Everything else looks good

gannonprudhomme:
You added the folder to the `.gitignore` correctly but you still need to delete the files so they're untracked in git.

autoscheduler/frontend/src/utils/sortMeetingFunction.tsx
```diff
@@ -0,0 +1,8 @@
+import Meeting from '../types/Meeting';
+
+// sorts meeetings based on meeting type
+const sortMeeting = (a: Meeting, b: Meeting): number => {
+  const value = a.meetingType - b.meetingType;
+  return value;
+};
```
ryan-conn, line 7:
```suggestion
const sortMeeting = (a: Meeting, b: Meeting): number => a.meetingType - b.meetingType;
```
Since this immediately returns a value, we can use this shorthand instead

autoscheduler/frontend/src/utils/sortMeetingFunction.tsx
```diff
@@ -0,0 +1,8 @@
+import Meeting from '../types/Meeting';
+
+// sorts meeetings based on meeting type
```
ryan-conn, line 3:
```suggestion
/** Sorts meetings based on meeting type */
```
A couple things:
- `meetings` is spelled wrong
- If you wrap a comment with `/** */`, it will be a docstring (meaning that when it autocompletes, you'll be able to see this text as a reminder of what the function does

ryan-conn (changes requested):
2 really minor comments that make this a bit cleaner

gannonprudhomme:
Pretty sure you removed the code that called the sort meeting function

gannonprudhomme (changes requested):
We'll ask you do it in the future, but can you insert this test in `tests/redux/CourseCard.test.ts` under line 315? Just so we can be sure this functionality works going forward. It's under that block b/c it's testing the `parseSectionSelected` function.

```typescript
    describe('sorts meetings', () => {
      test('when the meetings are out of order', () => {
        // arrange
        const genericMeetingInput = {
          days: [true, false, false, false, false, false, false],
          start_time: '8:00',
          end_time: '8:50',
        };

        const input = [{
          id: 1,
          crn: 1,
          subject: 'CSCE',
          course_num: '121',
          section_num: '500',
          min_credits: 3,
          max_credits: 3,
          current_enrollment: 0,
          max_enrollment: 1,
          instructor_name: 'Instructor Name',
          honors: false,
          web: false,
          asynchronous: false,
          meetings: [
            {
              ...genericMeetingInput,
              id: 13,
              type: 'EXAM',
            },
            {
              ...genericMeetingInput,
              id: 12,
              type: 'LAB',
            },
            {
              ...genericMeetingInput,
              id: 11,
              type: 'LEC',
            },
          ],
          grades: null as any,
        }];

        const section = new Section({
          id: 1,
          crn: 1,
          subject: 'CSCE',
          courseNum: '121',
          sectionNum: '500',
          minCredits: 3,
          maxCredits: 3,
          currentEnrollment: 0,
          maxEnrollment: 1,
          instructor: new Instructor({ name: 'Instructor Name' }),
          grades: null as any,
          honors: false,
          web: false,
          asynchronous: false,
        });

        const genericMeetingOutput = {
          building: '',
          meetingDays: [true, false, false, false, false, false, false],
          startTimeHours: 8,
          startTimeMinutes: 0,
          endTimeHours: 8,
          endTimeMinutes: 50,
          section,
        };

        const meetings = [
          new Meeting({
            ...genericMeetingOutput,
            id: 11,
            meetingType: MeetingType.LEC,
          }),
          new Meeting({
            ...genericMeetingOutput,
            id: 12,
            meetingType: MeetingType.LAB,
          }),
          new Meeting({
            ...genericMeetingOutput,
            id: 13,
            meetingType: MeetingType.EXAM,
          }),
        ];

        const expected = [{ section, meetings, selected: false }];

        // act
        const output = parseSectionSelected(input);

        // assert
        expect(output).toEqual(expected);
      });
    });
```
It looks like a lot, but it's mostly just the creation of the input/outputs

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -591,4 +591,101 @@ describe('Course Cards Redux', () => {
       expect(store.getState().courseCards[1].collapsed).toBe(true);
     });
   });
+  describe('sorts meetings', () => {
```
gannonprudhomme, line 594: You added it on line 594 - can you add it on line 315 instead?

gannonprudhomme (changes requested):

PR #394, opened 2020-10-06
Redirect on null term

### LINKED ISSUE #301: Redirect to landing page or alert user when scheduling page is open with no term selected
Since using the scheduling page with no term selected is completely useless, I propose that we just redirect back to the landing page if this happens. This is possible if you don't have cookies enabled, if you've never selected a term, and probably a couple other situations, so it would be nice to just redirect users to prevent confusion.

eelyort:
## Description

This makes the SchedulingPage redirect itself to the homepage if term is null (when someone enters ./schedule url).

## How to test

Visit ./schedule in incognito mode and it should open the homepage. Tested in automatic tests.

## Screenshots

Normal operation:
normal

Redirect (go straight to schedules without cookies):
redirect

Refresh:
refresh

## Update:

Changes reverted, there is now slight flashing on redirect but no flashing on refresh. See above screenshots.

Closes #301

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
@@ -1,14 +1,24 @@
 import * as React from 'react';
-import { useDispatch } from 'react-redux';
-import { RouteComponentProps } from '@reach/router';
+import { useDispatch, useSelector } from 'react-redux';
+import { RouteComponentProps, navigate } from '@reach/router';
 import Schedule from './Schedule/Schedule';
 import * as styles from './SchedulingPage.css';
 import ConfigureCard from './ConfigureCard/ConfigureCard';
 import SchedulePreview from './SchedulePreview/SchedulePreview';
 import CourseSelectColumn from './CourseSelectColumn/CourseSelectColumn';
 import setTerm from '../../redux/actions/term';
+import { RootState } from '../../redux/reducer';
 
 const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
+  // get term from redux state
+  const term = useSelector<RootState, string>((state) => state.term);
+  // when no term is selected, redirect back to the home page
+  React.useEffect(() => {
+    if(!term){
+      navigate('/');
+    }
+  });
+
   const dispatch = useDispatch();
 
   // Set redux state on page load based on term from user session
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useDispatch, useSelector } from 'react-redux';
+import { useDispatch } from 'react-redux';
 import { RouteComponentProps, navigate } from '@reach/router';
 import Schedule from './Schedule/Schedule';
 import * as styles from './SchedulingPage.css';
@@ -7,7 +7,6 @@ import ConfigureCard from './ConfigureCard/ConfigureCard';
 import SchedulePreview from './SchedulePreview/SchedulePreview';
 import CourseSelectColumn from './CourseSelectColumn/CourseSelectColumn';
 import setTerm from '../../redux/actions/term';
-import { RootState } from '../../redux/reducer';
 
 const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
   const dispatch = useDispatch();
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
@@ -14,7 +14,7 @@ const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
   const term = useSelector<RootState, string>((state) => state.term);
   // when no term is selected, redirect back to the home page
   React.useEffect(() => {
-    if(!term){
+    if (!term) {
       navigate('/');
     }
   });
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
@@ -10,23 +10,16 @@ import setTerm from '../../redux/actions/term';
 import { RootState } from '../../redux/reducer';
 
 const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
-  // get term from redux state
-  const termForRedir = useSelector<RootState, string>((state) => state.term);
-  // when no term is selected, redirect back to the home page
-  React.useEffect(() => {
-    if (!termForRedir) {
-      navigate('/');
-    }
-  });
-
   const dispatch = useDispatch();
 
   // Set redux state on page load based on term from user session
   React.useEffect(() => {
     fetch('sessions/get_last_term').then((res) => res.json()).then(({ term }) => {
-      // If unable to get a term, do nothing (term is set by SelectTerm on landing page,
-      // but session functionality will be unavailable)
+      console.log(term);
+      // If unable to get a term, redirect to hompage since term is set by
+      // SelectTerm on landing page and session functionality will be unavailable)
       if (term) dispatch(setTerm(term));
+      else navigate('/');
     });
   }, [dispatch]);
 
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { useDispatch } from 'react-redux';
+import { useDispatch, useSelector } from 'react-redux';
 import { RouteComponentProps, navigate } from '@reach/router';
 import Schedule from './Schedule/Schedule';
 import * as styles from './SchedulingPage.css';
@@ -7,6 +7,7 @@ import ConfigureCard from './ConfigureCard/ConfigureCard';
 import SchedulePreview from './SchedulePreview/SchedulePreview';
 import CourseSelectColumn from './CourseSelectColumn/CourseSelectColumn';
 import setTerm from '../../redux/actions/term';
+import { RootState } from '../../redux/reducer';
 
 const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
   const dispatch = useDispatch();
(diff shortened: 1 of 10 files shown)


autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
```diff
@@ -10,11 +10,33 @@ import { Provider } from 'react-redux';
...
 import { mockFetchSchedulerGenerate } from '../testData';
 
+// Mocks navigate, so we can assert that it redirected to the correct url for Redirects to /schedule
+// This must be outside of all describes in order to function correctly
+jest.mock('@reach/router', () => ({
+  navigate: jest.fn(),
+}));
+
 describe('Scheduling Page UI', () => {
+  describe('redirects to the homepage', () => {
```
firejake308, line 25: Can you also add a test asserting that it doesn't redirect if a term is selected?

firejake308 (dismissed):
Works in browser. Just one testing thing and this should be good to go

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -1,14 +1,24 @@
...
 import * as styles from './SchedulingPage.css';
 import ConfigureCard from './ConfigureCard/ConfigureCard';
 import SchedulePreview from './SchedulePreview/SchedulePreview';
 import CourseSelectColumn from './CourseSelectColumn/CourseSelectColumn';
 import setTerm from '../../redux/actions/term';
+import { RootState } from '../../redux/reducer';
 
 const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
+  // get term from redux state
+  const termForRedir = useSelector<RootState, string>((state) => state.term);
```
gannonprudhomme, line 14: So honestly I think we simplified this a bit too much when we explained this. While this does technically fix the issue we had, it also ignores the fact that we have `sessions/get_last_term`, which restores what term you had picked last from your saved session. Moreover, because the Redux gets completely wiped on refreshing, term will _always_ be null on refresh.

What this should actually do is wait for the `sessions/get_last_term` response (or more over go in the response handling of it), and if there is no saved term, then should it redirect you to the landing page. You can instead remove this `useSelector` and such, and just add the check + `navigate` around line 29 in this file.

gannonprudhomme (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
eelyort: image
Is this more what you wanted?
This does reintroduce the flashing problem but idt we can stop that since its on a short delay now?
- If you want I can try to move that code to a pre-render hook (getDerivedStateFromProps()?)

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
eelyort: image
How's this?

firejake308: wait holdup. I just realized that this test is 100% empty, as in, nothing is actually rendered. How does that work?

eelyort: I think that when the fetchMock is reset it calls render which then leads to the redirect on the effectHook. If there's anything I assumed wrong let me know. Thanks!

eelyort: image
image
Do these look good?

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
eelyort: image
image
Do these look good?

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -1,22 +1,24 @@
...
-import { RouteComponentProps } from '@reach/router';
+import { useDispatch, useSelector } from 'react-redux';
+import { RouteComponentProps, navigate } from '@reach/router';
 import Schedule from './Schedule/Schedule';
 import * as styles from './SchedulingPage.css';
 import ConfigureCard from './ConfigureCard/ConfigureCard';
 import SchedulePreview from './SchedulePreview/SchedulePreview';
 import CourseSelectColumn from './CourseSelectColumn/CourseSelectColumn';
 import setTerm from '../../redux/actions/term';
+import { RootState } from '../../redux/reducer';
```
gannonprudhomme, line 10: Make sure to fix these linting errors

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
```diff
@@ -10,11 +10,67 @@ import { Provider } from 'react-redux';
...
 import SchedulingPage from '../../components/SchedulingPage/SchedulingPage';
 import { mockFetchSchedulerGenerate } from '../testData';
 
 describe('Scheduling Page UI', () => {
+  // setup and teardown spy function on navigate
+  const navSpy = jest.spyOn(router, 'navigate');
+  // don't actually call navigate
+  beforeAll(() => navSpy.mockImplementation(() => null));
+  // restore navigate to original
+  afterAll(navSpy.mockRestore);
```
gannonprudhomme, line 24: What's your thought-process for this being `beforeAll`/`afterAll` instead of `beforeEach`/`afterEach`?

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
```diff
@@ -10,11 +10,67 @@ import { Provider } from 'react-redux';
...
+      fetchMock.mockResponseOnce(JSON.stringify({}));
+
+      // render
+      render(
+        <Provider store={store}>
+          <SchedulingPage />
+        </Provider>,
+      );
+
+      // do nothing - redirects automatically
```
gannonprudhomme, line 41: typically in these scenarios instead of saying "do nothing", we'd instead consider the "act" part of the test rendering, so you can remove the `do nothing` line and change `render` -> `act.

Seeing the
```
// render
render(...)
```
made me think of a pic I saw on reddit today (frontpage of /r/programmerhumor):
image

gannonprudhomme (changes requested):
Great job with getting these tests working! Tried to mess with them (e.g. intentionally making them fail) and it seemed to work well. For some reason the `redirects to the home page, when no term is selected` test ends up calling `navigate('/')` 3 times, but testing it in the browser it seems to only call it once so I think it's probably fine.

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
gannonprudhomme: Yeah that looks good!

Just btw when you post code, it's easier to copy/paste it in and surround it with 3 grave (`)'s to make a [codeblock](https://guides.github.com/features/mastering-markdown//#GitHub-flavored-markdown). Just makes it easier to view/read on different devices.

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
eelyort: Wait would the 3 time call also apply to the fetch last term? Cause the navigate call is just directly inside that promise.

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
eelyort: So basically the `beforeAll` changes the navigate function to not do anything. The other tests in the file fail if navigate is actually called since navigating when app is null causes undefined behavior (I think). So you don't actually need to re-do that multiple times. The counter is reset by `beforeAll(navSpy.mockClear)` once in the beginning of the doesn't redirect test, otherwise the counter doesn't matter anyways so I tried to clear/reset as little as possible for speed.

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
gannonprudhomme: You can do `waitFor(() => expect(navSpy).toHaveBeenCalledTimes(50)` and it'll say how many were called (if you do less than the number of times it's been called it will have at some point been that value so it'll pass and not show you how many there actually were)

eelyort: Cool thanks!

autoscheduler/frontend/src/tests/ui/SchedulingPage.test.tsx
eelyort: Cool thanks! I changed the comments a bit.

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -21,6 +22,14 @@ const SchedulingPage: React.FC<RouteComponentProps> = (): JSX.Element => {
     });
   }, [dispatch]);
 
+  // get current term
+  const termCurr = useSelector<RootState, string>((state) => state.term);
```
gannonprudhomme, line 26: Could you move this `termCurr` definition to right below the `const dispatch = `? We typically like to have the "state" declaring before like taking actions/calling things (e.g. the `React.useEffect` or your `return null` when the term is fasly)

gannonprudhomme (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
eelyort: Fixed with the revert to redirect flashing.

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
gannonprudhomme, line 10: Can you fix these two linting errors? Afterwards we'll be good to go

gannonprudhomme (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
eelyort: Fixed. Thanks for finding that for me!

gannonprudhomme (approved):
Looks good! Make sure when you merge this to choose "Squash and merge" (as opposed to "make a merge commit" or "rebase and merge")

Issue #389, opened 2020-10-05
Add asynchronous-only filter to the frontend


gannonprudhomme:
This entails adding a field to the `BasicSelect` that's identical to the current `Honors` and `Web` ones. You'll also need to add it to the Redux store, so when we press the `Generate schedules` button it knows to pass it to the backend (note you'll also need to make it pass it to the backend). Note that this won't actually function until #388 is done.

### Add asynchronous to Section type on the frontend

1) Add it to the `Section` type
    - Located in `autoscheduler/frontend/src/types/Section.ts`
    - It should be identical to `honors` and `web`
    - Make sure you add the error checking in the constructor like `honors` and `web` have it.
2) You'll also have to update the tests that initialize their own `Section` types - just run `npm run test` and go around updating the respective objects.
    - You'll know you're done when all of the tests start passing again.

### Create the `BasicOptionRow` in `BasicSelect`

1) You'll want to create an identical `BasicOptionRow` like how `honors` and `web` have it.
       - Located in `autoscheduler/frontend/src/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicSelect.tsx`

The Web/Honors filters are the things at the bottom of this:

 ![chrome_sdeYlr8DZW.png](https://images.zenhubusercontent.com/5da358de33f74500019c3b7f/b011f9f8-136c-4267-b7b2-e555c8cd3be0)

PR #397, opened 2020-10-08
Backend - async only filter

### LINKED ISSUE #388: Add asynchronous-only functionality to backend
So this is really two parts: adding the `asynchronous` field to the model and assign it inside of `scrape_courses`, then the accepting of this field from the `scheduler/generate` route which you'll then pass to the `create_schedules` function and have it correctly exclude the sections accordingly.

### `Asynchronous-only` field to `scrape_courses`

1. You'll need to add a model to the `Section` model called `asynchronous` that is identical to the `honors` and `web` fields. The file is located in `autoscheduler/scraper/models/Section.py`
    - After you do this, you'll then run `python manage.py makemigrations` to create the migration file (which basically records this column addition to the database table, so we can apply it to our local databases + Google Cloud SQL). 
    - You'll then run `python manage.py migrate` to actually apply the migration to your database
    -  Make sure you commit the migration file (will be located in `autoscheduler/scraper/migrations/`)
2. You will then need to correctly assign this when we're scraping (`scrape_courses`, i.e.`python manage.py scrape_courses`) so we can quickly query for sections that are/aren't `asynchronous`.
    - Scrape courses is located in `autoscheduler/scraper/management/commands/scrape_courses.py`
    - For each Section we scrape, you'll need to determine if is `asynchronous` and assigning it accordingly by iterating over the `meetings` object (line 76) and seeing if all of their meeting times are null/`None` - if they are, then that section is asynchronous.    
        - Inside of that file is the function `parse_section`, which is where you'll need to make these changes

### Passing the filter from `scheduler/generate` route to `create_schedules` function

On the frontend they'll now be sending you an `asynchronous` field which will look + act exactly like the current honors/web filter on the `scheduler/generate` route. You'll then take this field when processing the route, pass it to `create_schedules`, which will correctly exclude/include `asynchronous` sections before generating the schedules.

1. Processing it in `create_schedules`

Since I'll probably end up doing this, I'm not going to finish this lol

gannonprudhomme:
## Description

Adds the `asynchronous` field to the `Section` model (and serializer), sets the field during scraping.

## How to test

Run scrape_courses, then run the server and look at like `COMM 203` (or any other course with a bunch of async sections) and double check the route is returning `asynchronous: true` for the async sections.

Since this adds a column to the Section table, you'll need to run `python manage.py migrate`

Note that `/scheduler/generate` won't work since the `asynchronous` field isn't being sent by the frontend - this will be fixed in #389, and I'll be sure to wait on merging this until that part is done.

## Related tasks

Closes #388


autoscheduler/scraper/tests/commands/scrape_courses_tests.py
```diff
@@ -351,6 +351,41 @@ def test_parse_section_gets_web(self):
...
+
+        # Arrange
+        fake_instructor = Instructor(id="Fake", email_address="a@b.c")
+        fake_instructor.save()
+
+        # Act
+        section, _ = parse_section(self.csce_web_section_json, fake_instructor)
+        section.save()
+
+        print(Section.objects.all().values())
```
ryan-conn, line 367: Remove this print

ryan-conn (changes requested):
Looks good except a print you forgot to remove, the actual changes are pretty clean

ryan-conn (dismissed):

ryan-conn:
Meant to comment that you should fix the merge conflict first but it's kinda obvious since you literally can't merge it right now lol

ryan-conn (approved):

PR #398, opened 2020-10-08
Frontend/full height course cards

### LINKED ISSUE #377: (idea) Add option to "focus" on CourseCard, making it take 100% of height
I think there should be an option to make a given CourseCard take up the entire height of the CourseSelectColumn (first column), as it might be useful when you're doing the Section customization level. We can discuss if this is actually a feature we want or not (hence the `(idea)` in title) - I don't think it's something we need before we launch

ryan-conn:
## Description

Moves expand/collapse for course cards to redux, only one can be expanded at a time, and the expanded card will fill as much height as possible (with a minimum height roughly equal to what it was previously). No tests right now because I wanted some input on if we like how this works, specific behavior, etc.

## Rationale
The only thing in this branch that isn't how I imagined it is the extra height when the currently selected card is in section view and only has a couple sections. I couldn't think of any way to keep the fit-to-height behavior as well as having a minimum height without making this compromise (despite spending 4 hours in CSS hell), but I still think that in the vast majority of use cases this is a lot nicer.

## How to test
Make some cards, the expanded one will fit to your screen size (unless you're Adel and have a screen made for ants, in which case I've made it so this works basically the same way as before)

## Screenshots
I think that 7 course cards is about as high as most users will go, so here are some example sizes:

| Old appearance (always 300px) | 1080, 5 cards (417px) | 1080p, 7 cards (340px) |
|--|--|--|
| image | image | image |

| 1440p, 7 cards (697px) | 720p (314px, any more than 3 cards makes it scroll) |
|--|--|
| image | image |

Finally, here's the one case I don't really like the appearance of. If you're on the section select and there are only 1 or 2 sections, there will be some extra empty space that I couldn't find any way to avoid even using refs:
image
It looks normal on basic select though since I check if you're on section select with a course selected before giving a minimum height:
image

## Related tasks

Closes #377.

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
@@ -1,5 +1,6 @@
 .container {
-    width: calc(100% - 8px);
+    box-sizing: border-box;
+    width: 100%;
     height: 100%;
     display: flex;
     flex-direction: column;
@@ -13,21 +14,30 @@
 .column-wrapper {
     position: relative;
     width: 100%;
+    height: 100%;
     flex-grow: 1;
     margin-bottom: 8px;
 }
 
 .course-select-column {
     position: absolute;
+    display: flex;
+    flex-direction: column;
     top: 0px;
     bottom: 0px;
     overflow-y: auto;
     width: 100%;
+    height: 100%;
 }
 
 .row {
-    padding-bottom: 8px;
+    padding-top: 8px;
     padding-right: 8px;
+    box-sizing: border-box;
+}
+
+.expanded-row {
+    min-height: 500px;
 }
 
 #add-course-button {
@@ -37,7 +47,6 @@
 #button-container {
     position: sticky;
     bottom: 0;
-    padding-bottom: 8px;
     padding-right: 8px;
     background-color: #c2c2c2;
     z-index: 2;
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
@@ -30,10 +30,6 @@
     padding-right: 8px;
 }
 
-.row:nth-last-of-type(2) {
-    padding-bottom: 0px;
-}
-
 #add-course-button {
     width: 100%;
 }
@@ -41,7 +37,7 @@
 #button-container {
     position: sticky;
     bottom: 0;
-    padding-top: 8px;
+    padding-bottom: 8px;
     padding-right: 8px;
     background-color: #c2c2c2;
     z-index: 2;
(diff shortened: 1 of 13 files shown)


ryan-conn:
Should be ready for review now that the edge case with low height section select has been fixed and I added tests.

firejake308:
I want this merged because I am so tired of manually collapsing course cards

gannonprudhomme:
> I want this merged because I am so tired of manually collapsing course cards

Then review it 😉

ryan-conn:
Changes I just pushed make it so if you delete the expanded card, it will expand one next to it (below if possible, or above if none are below it)

gannonprudhomme:
Bug I ran into:
full-course-cards-bug

ryan-conn:
> Bug I ran into:

Fixed, this was due to destructuring the existing state instead of the new card attributes when `addCourseCard` is called with an existing card.

I also noticed making this change that it will *always* expand the new card (even when loading your cards, when you don't want to expand them), so I fixed that as well

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -19,10 +20,20 @@ const CourseSelectColumn: React.FC = () => {
...
-
+  const term = useSelector<RootState, string>((state) => state.term);
   const dispatch = useDispatch();
 
-  const term = useSelector<RootState, string>((state) => state.term);
+  const expandedRowRef = React.useRef<HTMLDivElement>(null);
+  React.useLayoutEffect(() => {
+    if (expandedRowRef.current) {
+      const expandedRowHeight = expandedRowRef.current.children[0].clientHeight;
+      if (expandedRowHeight < 500 - 8) {
```
firejake308, line 30: We should probably mention that 500 needs to be the minimum height from the `.expanded-row` rule, and that the 8 comes from the `padding-top` on `.row`

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -13,25 +14,35 @@
...
+}
+
+.expanded-row {
+    overflow-y: hidden;
+    min-height: 500px;
 }
 
-.row:nth-last-of-type(2) {
-    padding-bottom: 0px;
+.expanded-row-overflow {
```
firejake308, line 44: I would rename this to `expanded-row-small` since it's applied to small/short cards that fit within the 500px limit

firejake308 (changes requested):
Tried to break the code in browser by mindlessly adding, deleting, expanding, and collapsing different course cards, no bugs found. Love `autoFocus` btw. 2 suggestions for comments/code style and this should be good to go. We should probably have Gannon test it too, though, since he's better at teasing out bugs than I am usually.

gannonprudhomme (commented):
minor things I wrote a bit ago, will re-review later tn

ryan-conn:
Comments should be addressed, updated with master

autoscheduler/frontend/src/redux/reducers/courseCards.ts
```diff
@@ -43,34 +43,97 @@ const initialCourseCardArray: CourseCardArray = {
...
+        return getStateAfterExpanding(state, action.index, action.courseCard);
+      }
+
+      // otherwise just update this card
       return {
         ...state,
         [action.index]: { ...state[action.index], ...action.courseCard },
         numCardsCreated: Math.max(state.numCardsCreated, action.index + 1),
       };
+    }
```
gannonprudhomme, line 136: Don't think you need the braces around this case block, as you don't declare anything

gannonprudhomme (changes requested):
Few minor things, otherwise this looks good to me; also couldn't find any bug when I tried messing with it

PR #402, opened 2020-10-11
Mini dividers + make instructors sticky for Course Cards

### LINKED ISSUE #348: Pin instructors + gpa row top of SectionSelect
We should pin the instructors/gpa row to the top of the section select, so that when a professor has a long list of sections and we're like half way down, we can still see what professor it is.

### LINKED ISSUE #362: Add mini dividers between sections
As we discussed in this week's meetings, we should add dividers for sections with less padding and more indentation to distinguish them from the instructor dividers. Here's the mockup Adel made to show what it'll look like:
image
Note that the last section for each instructor doesn't have a divider, since it would look a bit cluttered.

firejake308:
## Description

Makes instructors sticky to make long lists more comprehensible and adds inset dividers below each section to make the space feel more organized.

## Rationale

In order to make all instructor subheaders the same height, I set the minimum height to 20px. This is because the instructor subheaders stack on top of each other when they are sticky, so if one is bigger than the others, it will stick out. There is a possibility that 20px will not work on other screens, but I doubt it since the honors icon (which is the determining factor) should always be the same size.

## How to test
Look at a course with multiple instructors and sections, like CSCE 121. No automated tests included, mainly because it's impossible.

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #362 
Closes #348

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
@@ -27,9 +27,11 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
   const makeList = (): JSX.Element[] => {
     let lastProf: string = null;
     let lastHonors = false;
+    let currProfGroupStart = 0;
     return sections.map((sectionData, secIdx) => {
       const firstInProfGroup = lastProf !== sectionData.section.instructor.name
         || lastHonors !== sectionData.section.honors;
+      if (firstInProfGroup) currProfGroupStart = secIdx;
 
       lastProf = sectionData.section.instructor.name;
       lastHonors = sectionData.section.honors;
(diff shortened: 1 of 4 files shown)


firejake308:
Applied changes from 10/11 meeting, ready for review

ryan-conn (approved):
Looks good, one *really* small nitpick I found is that when you stop right when 2 professors are at the top, it looks a bit weird:
image
Compared to the documentation for list headers, which "slides out" the old header when a new one starts:
image
Also see the [documentation](https://material-ui.com/components/lists/#pinned-subheader-list) for a live example

Honestly I'm fine with it as is if you don't want to make that change because it's a pretty minor thing and it already looks really good, so I'll approve this. You can decide if you want to implement that change in this PR or not and if you decide not to work on it we can make it a future issue as something to look into, I think it has something to do with the `subheader` prop but I've looked at the documentation for a little bit and nothing obvious jumps out at me as the reason for the current behavior

gannonprudhomme:
> Looks good, one _really_ small nitpick I found is that when you stop right when 2 professors are at the top, it looks a bit weird:
> Compared to the documentation for list headers, which "slides out" the old header when a new one starts:

Looking at that documentation demo I would personally really like if we did it as it feels really clean on the demo, but like Ryan said if it's not like super easy I'm fine leaving it for now

firejake308:
Basically, I had to wrap each professor group in its own `<ul>` in order to get it to work. This required making the section iterating logic a little more complicated, but it looks super clean in the UI. Basically,

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -27,21 +27,33 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
   const makeList = (): JSX.Element[] => {
```
ryan-conn, line 27: One last thing I would say is to add some comments to this function (especially about how the `ul` wrappers make it so the instructor groups slide in/out), this component is getting a little complicated so it would be nice to have some documentation in case we make changes in the future

ryan-conn (changes requested):
Functionally this is great, I'd really appreciate some comments on `makeList` for the ease of maintainability

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
firejake308: Done

ryan-conn (approved):
Somehow didn't notice you added the comments I requested until now, but this looks *noice*

Issue #399, opened 2020-10-10
Add calculating+displaying of the reason /scheduler/generate failed


gannonprudhomme:
One piece of feedback we've gotten is that it's not easy to tell the reason generating a schedule failed. As such, on the backend we figure out what is the reason it failed and send back some message explaining why it failed. Ideally we'd also give suggestions on how to fix it, but we'll see how easy that ends up being. We'll split this up into frontend/backend later on, but leaving this how it is for now (especially since it's mainly a backend task).

The reasons for failing that we could account for that pop up in my head are:

- All sections are full
    - Only for basic customization level, as we'll eventually make it so that you can manually select a full section
- All available sections for a course intersect with availablities
- Different courses' sections intersect with each other?
- Failed b/c basic filters are too restrictive
    - e.g. looked for a class with both honors and online but there weren't any

Feel free to edit this @ryan-conn as this is really your domain

PR #404, opened 2020-10-12
Thunkified generateSchedules

firejake308:
## Description

Multi day availabilities were very laggy, so I did some profiling and realized that the ConfigureCard was being re-rendered very frequently. This occurred because the ConfigureCard needed access to the most up-to-date availabilities list, which was constantly being updated. By extracting the `generateSchedules` functionality to a Thunk Action, I can simply request the most up-to-date availabilities list once, at the time of schedule fetching, rather than asking for it constantly via `useSelector`.

## Rationale

I got rid of the "no schedules generated snackbar" because it would be a hassle to pass information back from Redux to the UI, and we can deal with that problem later, since we have to replace the snakcbar with a dialog anyway.

## How to test

Automated tests are note possible for this bug, as it deals with lag. To test, create multi-day availabilities as you normally would when using the site. There should be noticeably less lag than there is on master, which you can test by opening [revregistration.com](https://revregistration.com) in another tab. To stress test, start a click in the middle of the schedule, then drag around in circles. The availabilities should follow your mouse quickly. Again, you can compare to master to verify that these changes have actually made a difference.

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
@@ -1,8 +1,7 @@
 import * as React from 'react';
 import {
-  Button, Checkbox, ListItem, ListItemIcon, ListItemText, Snackbar, IconButton,
+  Button, Checkbox, ListItem, ListItemIcon, ListItemText,
 } from '@material-ui/core';
-import CloseIcon from '@material-ui/icons/Close';
 import { useDispatch } from 'react-redux';
 import GenericCard from '../../GenericCard/GenericCard';
 import SmallFastProgress from '../../SmallFastProgress';
@@ -16,7 +15,6 @@ import { generateSchedules } from '../../../redux/actions/schedules';
 const ConfigureCard: React.FC = () => {
   const [includeFull, setIncludeFull] = React.useState(false);
   const [loading, setLoading] = React.useState(false);
-  const [showSnackbar, setShowSnackbar] = React.useState(false);
 
   const dispatch = useDispatch();
 
@@ -25,12 +23,6 @@ const ConfigureCard: React.FC = () => {
   React.useEffect((): VoidFunction => (): void => {
     isMounted.current = false;
   }, []);
-
-  // closes the snackbar if the user presses the close icon, but not if they click away
-  const handleSnackbarClose = (_event: any, reason: string): void => {
-    if (reason === 'clickaway') return;
-    setShowSnackbar(false);
-  };
 
   const fetchSchedules = React.useCallback(() => {
     // show loading indicator
@@ -69,17 +61,6 @@ const ConfigureCard: React.FC = () => {
             : 'Generate Schedules'}
         </Button>
       </div>
-      <Snackbar
-        open={showSnackbar}
-        autoHideDuration={5000}
-        message="No schedules found. Try widening your criteria."
-        onClose={handleSnackbarClose}
-        action={(
-          <IconButton aria-label="close" onClick={(): void => setShowSnackbar(false)}>
-            <CloseIcon fontSize="small" style={{ color: 'white' }} />
-          </IconButton>
-        )}
-      />
     </GenericCard>
   );
 };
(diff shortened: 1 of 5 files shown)


gannonprudhomme:
Just btw this is failing a few frontend tests

gannonprudhomme:
> I got rid of the "no schedules generated snackbar" because it would be a hassle to pass information back from Redux to the UI, and we can deal with that problem later, since we have to replace the snakcbar with a dialog anyway.

While you're right in that we will be replacing the snackbar with a dialog (in #399), I don't want to launch the website without at least _something_ notifying them that schedules weren't generated. As such, I say either figure out at way of getting the snackbar added back, or we postpone this until we get #399 working.

firejake308:
Ready for review

gannonprudhomme:
Also can you re-add the tests that you removed in 567144a

autoscheduler/frontend/src/hooks/useThunkDispatch.ts
```diff
@@ -0,0 +1,8 @@
+import { useDispatch } from 'react-redux';
+import { AnyAction } from 'redux';
+import { ThunkDispatch } from 'redux-thunk';
+import { RootState } from '../redux/reducer';
+
+export default function useThunkDispatch(): ThunkDispatch<RootState, undefined, AnyAction> {
```
ryan-conn, line 6: This literally just typecasts `useDispatch` so that typescript doesn't get upset at it, right? Would appreciate a comment as to why/when this is necessary

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -25,95 +16,29 @@ import { formatTime } from '../../../utils/timeUtil';
...
-  };
-
   // closes the snackbar if the user presses the close icon, but not if they click away
   const handleSnackbarClose = (_event: any, reason: string): void => {
     if (reason === 'clickaway') return;
-    setShowSnackbar(false);
+    setSnackbarMessage('');
   };
 
   const fetchSchedules = React.useCallback(() => {
```
ryan-conn, line 34: Is it still worth using `useCallback` here? Since all of the logic got moved to the action it seems like this wouldn't increase performance any more

autoscheduler/frontend/src/redux/actions/schedules.ts
```diff
@@ -47,3 +55,74 @@ export function renameSchedule(index: number, name: string): RenameScheduleActio
...
+    ).then(
+      (generatedSchedules: any[][]) => generatedSchedules.map(parseAllMeetings),
+    ).then(
+      checkIfEmpty,
+    )
+      .then((schedules: Meeting[][]) => {
+        dispatch(replaceSchedules(schedules));
+        dispatch(selectSchedule(0));
+      });
+  };
```
ryan-conn, line 127: In case something else goes wrong, it would be nice to catch an error at the end here with something like "There was an error generating schedules, please try again"

ryan-conn (changes requested):
A couple of small comments and this should be good

autoscheduler/frontend/src/redux/actions/schedules.ts
ryan-conn: Actually nevermind I did this in my branch for generate schedules error messages

autoscheduler/frontend/src/components/SchedulingPage/ConfigureCard/ConfigureCard.tsx
```diff
@@ -144,18 +69,18 @@ const ConfigureCard: React.FC = () => {
...
-        open={showSnackbar}
-        autoHideDuration={5000}
-        message="No schedules found. Try widening your criteria."
-        onClose={handleSnackbarClose}
-        action={(
-          <IconButton aria-label="close" onClick={(): void => setShowSnackbar(false)}>
-            <CloseIcon fontSize="small" style={{ color: 'white' }} />
-          </IconButton>
+        <Snackbar
+          open={!!snackbarMessage}
```
ryan-conn, line 73: Really minor but just for future reference it's more "typescripty" (that's totally a word) to use `Boolean(x)` instead of `!!x` since it's more clear that you're typecasting, I'll just make this change in the generate schedules error message PR though since it's really small, I'll be changing this part of the code anyways, and I don't want to hold up this PR any more

ryan-conn (approved):
Looks good

PR #406, opened 2020-10-12
Separate honors grades

### LINKED ISSUE #261: Separate honors/non-honors sections for Grades.instructor_performance
Since honors and non-honors section for the same course & professor's grade distributions typically vary, we should add an option to `GradeManager.instructor_performance` that only includes honors / non-honors sections. Should just be able to add a `section__honors=honors_param` to the `self.prefetch_related("section").filter(...)` call. We'll also have to change the call in the `SectionSerializer` to make sure that it uses the correct calls.

firejake308:
## Description

Sepearated instructor performance for honors and non-honors sections to give users a better idea of what they can expect from a course

## Rationale

Deleted unused api/grades route because it was failing tests and also unused

## How to test

Look at screenshot below of CSCE 222, or any other course where the same instructor has honors and non-honors courses. FINC 341 is another good example

## Screenshots

|Before|After|
|--|--|
|Before Screenshot|image|

## Related tasks

Closes #261


gannonprudhomme:
You have some backend linting errors you need to fix in this just btw

autoscheduler/scraper/tests/models_tests.py
```diff
@@ -129,7 +129,53 @@ def test_instructor_performance_calculates_from_all_sections(self):
...
         # Act
         result = Grades.objects.instructor_performance(
-            dept=subject, course_num=course_num, instructor=instructor
+            dept=subject, course_num=course_num, instructor=instructor, honors=False,
+        )
+
+        # Assert
+        self.assertEqual(expected, result)
+
+    def test_instructor_performance_ignores_honors_sections_for_non_honors_query(self):
```
gannonprudhomme, line 138: Sort of a nitpick, but can you duplicate this test and make it do an `honors=True` query and assert that it got the correct honors grade?

gannonprudhomme (changes requested):

autoscheduler/scraper/tests/models_tests.py
firejake308: No, that's valid. Will add.

gannonprudhomme (approved):

gannonprudhomme:
Oh lmao approved it without noticing you had merge conflicts, so just resolve those and I'll approve again.

gannonprudhomme (approved):

Issue #408, opened 2020-10-13
Add privacy policy page / popup


gannonprudhomme:
We don't necessarily need to have a popup, but with the addition of Google Analytics I'd like to be transparent as possible with our usage of their data. Basically, we'd outline that we don't share their data with Google / anyone else, what we use it for (hitcounts) and what data we actually look at. Also should probably tell them we're not connecting the analytics session to them logging in with Google (i.e. it's basically completely anonymized), as well as that 

We should probably double check all of these claims too, cause I wouldn't say I'm completely positive we're not inadvertently sharing their data with Google and etc

PR #409, opened 2020-10-13
Add Google Analytics script

gannonprudhomme:
## Description

As discussed in the last meeting, we'll be adding Google Analytics to the site so we can track hitcounts and the like.

Note that if you have an adblocker (uBlock origin, etc) it'll block this script.

I've invited everyone to the Google analytics account, which you can view at http://analytics.google.com, but here's what the homepage looks like:

chrome_Gx0HHeSM8O

## Related tasks

Per this addition, I made a task (#408) for adding a privacy policy page.

### DIFF
--- a/autoscheduler/frontend/templates/index.html
+++ b/autoscheduler/frontend/templates/index.html
@@ -21,7 +21,7 @@
             gtag('js', new Date());
 
             gtag('config', 'UA-179924187-1');
-        </script>        
+        </script>
     </head>
     <body>
         <div id="root"></div>
--- a/autoscheduler/frontend/templates/index.html
+++ b/autoscheduler/frontend/templates/index.html
@@ -13,6 +13,15 @@
                 background-color:#c2c2c2;
             }
         </style>
+        <!-- Global site tag (gtag.js) - Google Analytics -->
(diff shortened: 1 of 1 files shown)


autoscheduler/frontend/templates/index.html
```diff
@@ -13,6 +13,15 @@
...
         </style>
+        <!-- Global site tag (gtag.js) - Google Analytics -->
+        <script async src="https://www.googletagmanager.com/gtag/js?id=UA-179924187-1"></script>
+        <script>
+            window.dataLayer = window.dataLayer || [];
+            function gtag(){dataLayer.push(arguments);}
+            gtag('js', new Date());
+
+            gtag('config', 'UA-179924187-1');
+        </script>
```
ryan-conn, line 24: There's some extra whitespace here
```suggestion
        </script>
```

ryan-conn (changes requested):
One nitpick and this is good, gotta save those 8 bytes

autoscheduler/frontend/templates/index.html
gannonprudhomme: I am amazed you caught this

ryan-conn (approved):

PR #413, opened 2020-10-15
Fixed bug where cards deleted when loading are recreated

### LINKED ISSUE #412: Mid-loading deleted course cards re-appear when loading is finished

ryan-conn:
## Description

Fixes the bug described in #412.

## Rationale
This can still technically happen if you delete the auto-generated card before `get_last_term` completes:
oops
since the term changes, causing the card you deleted to be recreated after the term changes. The only way around this would be to not recreate the initial card when `clearCourseCards` is dispatched, but I think this way is probably better since that's a bit weird and it's hard to replicate this unless you're trying to do it on purpose.

## How to test

Try to replicate this bug on master/production and see how easy it is, then try it on this branch. No tests made because there's no determinate way to replicate it (as it has to do with how async actions are ordered)

## Screenshots

wowamazin

## Related tasks
Closes #412.

### DIFF
--- a/autoscheduler/frontend/src/redux/actions/courseCards.ts
+++ b/autoscheduler/frontend/src/redux/actions/courseCards.ts
@@ -282,7 +282,7 @@ function getSelectedSections(
 ): SectionSelected[] {
   const selectedSections = new Set(serialized.sections);
 
-  return courseCard.sections?.map((section): SectionSelected => ({
+  return courseCard?.sections?.map((section): SectionSelected => ({
     ...section,
     selected: selectedSections.has(section.section.id),
   })) || [];
@@ -311,7 +311,7 @@ export function replaceCourseCards(
     courseCards.forEach((courseCard, idx) => {
       const deserializedCard = deserializeCourseCard(courseCard);
       deserializedCards.push(deserializedCard);
-      dispatch(updateCourseCardSync(idx, deserializedCard));
+      dispatch(addCourseCard(deserializedCard, idx));
     });
 
     // all course cards are now marked as loading (preventing courses from being overwritten
--- a/autoscheduler/frontend/src/redux/reducers/courseCards.ts
+++ b/autoscheduler/frontend/src/redux/reducers/courseCards.ts
@@ -63,14 +63,13 @@ export default function courseCards(
         ...state,
         [action.index]: undefined,
       };
-    case UPDATE_COURSE_CARD: {
+    case UPDATE_COURSE_CARD:
       if (!state[action.index]) return state;
       return {
         ...state,
         [action.index]: { ...state[action.index], ...action.courseCard },
         numCardsCreated: Math.max(state.numCardsCreated, action.index + 1),
       };
-    }
     case CLEAR_COURSE_CARDS:
       return initialCourseCardArray;
     default:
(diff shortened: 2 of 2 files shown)


autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -42,12 +42,16 @@ export function removeCourseCard(index: number): RemoveCourseAction {
...
    */
-function updateCourseCardSync(index: number, courseCard: CourseCardOptions): UpdateCourseAction {
+function updateCourseCardSync(
+  index: number, courseCard: CourseCardOptions, canCreate = false,
+): UpdateCourseAction {
   return {
     type: UPDATE_COURSE_CARD,
     index,
     courseCard,
+    canCreate,
```
gannonprudhomme, line 54: What's the use of `canCreate`? I don't see it used anywhere

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -311,7 +315,7 @@ export function replaceCourseCards(
     courseCards.forEach((courseCard, idx) => {
       const deserializedCard = deserializeCourseCard(courseCard);
       deserializedCards.push(deserializedCard);
-      dispatch(updateCourseCardSync(idx, deserializedCard));
+      dispatch(addCourseCard(deserializedCard, idx));
```
gannonprudhomme, line 318: What's the thought process of the change from `updateCourseCardSync` to `addCourseCard`? Is there a reason it was `updateCourseCardSync` in the first place?

autoscheduler/frontend/src/redux/reducers/courseCards.ts
```diff
@@ -63,12 +64,14 @@ export default function courseCards(
         ...state,
         [action.index]: undefined,
       };
-    case UPDATE_COURSE_CARD:
+    case UPDATE_COURSE_CARD: {
+      if (!state[action.index]) return state;
```
gannonprudhomme, line 68: I assume this change is the actual fix itself? What's your thought process on doing this here vs in the `fetch` like I did for changing mismatched terms?

gannonprudhomme: Also why'd you add the brackets around it? I thought you only have to do that if you declare variables

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: Oops I rebased and forgot to push the change, that was my original solution but did something else (see my comment on addCourseCard)

autoscheduler/frontend/src/redux/actions/courseCards.ts
ryan-conn: Since I made `updateCourseCard` not change state unless the card already exists (which is how it's no longer possible for deleted cards to pop into existence), the call that originally creates the cards can't use it anymore

autoscheduler/frontend/src/redux/reducers/courseCards.ts
ryan-conn: > I assume this change is the actual fix itself? What's your thought process on doing this here vs in the `fetch` like I did for checking for mismatched terms?

I think this is better since it makes `addCourseCard` and `updateCourseCard` more distinct: use `addCourseCard` if you're adding a new one, `updateCourseCard` if you're trying to update info of an existing one. I also figured that making the change here would prevent any kinds of similar shenanigans from happening in the future, since we now have to be clear if we really want to add a new course card rather than inadvertently creating one as a side-effect of updating

gannonprudhomme (approved):

PR #414, opened 2020-10-16
Added favicon + navbar logo

### LINKED ISSUE #333: Create a logo & make it our website's favicon
Need to make a logo so we can make it our favicon (tab icon) for the website.

gannonprudhomme:
## Description

This adds a navbar logo and favicon to the site.

### Favicon

So first, most of the files were generated using http://realfavicongenerator.net , and I used the "favicon checker" on the home page to check that the favicons for different locations were working correctly.

You can use the favicon checker [here](https://realfavicongenerator.net/favicon_checker?protocol=http&site=www.revregistration.com) to see what the other things they test for are, but here's what we failed:

|iOS Safari Touch Icon|Classic, desktop browsers|
|--|--|
|chrome_Dz4gcZ8muR|chrome_qzslTii6ig|

I don't really care about the iOS one, since first we're not supporting mobile atm, and second I can't name a single time I've ever done the Safari website -> app shortcut thing in my 8 years of having an iPhone (pretty sure it's an OG feature). The second one also probably doesn't matter, since we're using the SVG favicon.

Second, the "main" favicon, `favicon.svg`, uses the CSS media property thing to detect the OS theme and switches from maroon to white when it detects dark mode. Note that you have to refresh the page to see this change (from what I can tell at least).

### NavBar logo

So I had to do a few things to get this to work both locally and on GCP. Because our `index.html` is served by django, it can dynamically determine what the static URL is to changed between `/static/` (for localhost) and `http://storage.googleapis.com/revregistration1.appspot.com/` (For GCP). To bring this functionality to the frontend, I used two things:

1. webpack's `DefinePlugin`, which functions in a really weird way - at compile time, it **replaces** the given word (`STATIC_URL` in this case) to whatever you have it defined as (excluding the declaration, but that isn't compiled to JS anyways). 
  a. That means that when you compile the project, and search for `STATIC_URL` in `main.js`, you won't see it anywhere 
  b. I added the `global.ts` file just so that we can treat `STATIC_URL` as a string without having webpack yell at us for not defining it.
2. the npm package `cross-env` to add the env var `PRODUCTION` whenever we run `npm run build`. This way it automatically changes from /static/ to the GCP URL whenever we run `npm run build`. 

## How to test

Build + run the site, and check that the logo on the navbar + favicon appear correct. Then to check the dark-mode SVG favicon, change your OS's theme and refresh.

## Screenshots

|Light mode|dark mode|
|--|--|
|chrome_i2Gt6wy4QV|chrome_doKVWf6sDg|

Here's what the logo looks like on the scheduling page
chrome_rz3tBhJM9I

## Related tasks

Closes #333


gannonprudhomme:
Also for some reason https://storage.googleapis.com/revregistration1.appspot.com/site.webmanifest gets blocked by our CORS policy - I tried to fix it with bd71c91 but I guess that didn't work. Will look into it.

.github/workflows/deploy-workflow.yml
```diff
@@ -1,8 +1,5 @@
 name: GCP Deploy
-on:
-  push:
-    branches:
-      - master
+on: push
```
firejake308, line 2: Reminder to reset before merge

autoscheduler/autoscheduler/settings/base.py
```diff
@@ -181,6 +181,10 @@
 
 USE_TZ = True
 
+# print(os.path.join(BASE_DIR, "assets"))
```
firejake308, line 184: Thank you for moving this file. Also, I guess we could erase the commented line now

autoscheduler/frontend/assets/browserconfig.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="utf-8"?>
```
firejake308, line 1: Is this for PWA-ification of the website?

autoscheduler/frontend/assets/site.webmanifest
```diff
@@ -0,0 +1,19 @@
+{
+    "name": "",
```
firejake308, line 2: I think we should go ahead and add a name, since it isn't too hard, and it'll help us achieve full PWA if we ever want to

firejake308 (changes requested):
Works for me in browser on both Chrome and Firefox, light mode and dark mode. Couple of tiny changes and this should be good to go

autoscheduler/frontend/assets/browserconfig.xml
gannonprudhomme: Yeah I think so - was just one of the things generated from realfavicongenerator.net. Believe this one is for like if you saved it on windows somehow, it'd be the icon that shows as the tile.

.github/workflows/deploy-workflow.yml
gannonprudhomme: Dropped the commit

gannonprudhomme:
Addressed changes and also resized the logo to 64x64, which reduced it from 148 KB to 1.5 KB. Could've probably done 32x32 since that's the size it gets rendered, but figured having it a higher resolution wouldn't hurt.

firejake308:
>Also for some reason https://storage.googleapis.com/revregistration1.appspot.com/site.webmanifest gets blocked by our CORS policy and logs as an error in the console - I tried to fix it with bd71c91 but I guess that didn't work. Will look into it.

Is that something we need to fix immediately? Favicon is showing, so I think it doesn't matter atm but just wanted to confirm

ryan-conn (approved):

PR #417, opened 2020-10-20
Save schedules

### LINKED ISSUE #267: Add saved schedule saving to sessions
This should add some route that the save button on the SchedulePreview (#228) can click to save that schedule. Should also add the sending of the saved schedules with the current retrieve-sessions functionality. 

Dependent on #260 (probably)

gannonprudhomme:
## Description

Adds saving schedules to the session. Note that this is based off of #410 due to its addition of the abstract get/set state route functionality.

## Rationale

I set the throttle time 5 seconds since I don't expect schedules to be modified much, but we can increase it if we really want to.

## How to test

Generate schedules then save (and rename) them, refresh the page, and ensure the schedules were saved.

## Screenshots

save-schedules-demo

## Related tasks

Closes #267

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
@@ -11,6 +11,7 @@ import { setSchedules } from '../../../redux/actions/schedules';
 import createThrottleFunction from '../../../utils/createThrottleFunction';
 import { parseAllMeetings } from '../../../redux/actions/courseCards';
 import SmallFastProgress from '../../SmallFastProgress';
+import { SaveSchedulesRequest, SerializedSchedule } from '../../../types/APIRequests';
 
 const throttle = createThrottleFunction();
 
@@ -62,7 +63,7 @@ const SchedulePreview: React.FC<SchedulePreviewProps> = ({
 
     // Serialize schedules and make API call
     const saveSchedules = (): void => {
-      const serializedSchedules = schedules.filter(
+      const serializedSchedules: SerializedSchedule[] = schedules.filter(
         // Filter out unsaved schedules
         (schedule) => schedule.saved,
       ).map((schedule) => ({
@@ -71,13 +72,15 @@ const SchedulePreview: React.FC<SchedulePreviewProps> = ({
         sections: [...new Set(schedule.meetings.map((m) => m.section.id))],
       }));
 
+      const request: SaveSchedulesRequest = { term, schedules: serializedSchedules }
+
       fetch('sessions/save_schedules', {
         method: 'PUT',
         headers: {
           'Content-Type': 'application/json',
           'X-CSRFToken': Cookies.get('csrftoken'),
         },
-        body: JSON.stringify({ term, schedules: serializedSchedules }),
+        body: JSON.stringify(request),
       });
     };
 
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
@@ -62,22 +62,14 @@ const SchedulePreview: React.FC<SchedulePreviewProps> = ({
 
     // Serialize schedules and make API call
     const saveSchedules = (): void => {
-      let serializedSchedules: { name: string; sections: number[] }[] = [];
-
-      if (schedules) {
-        if (schedules.length === 0) {
-          return;
-        }
-
-        serializedSchedules = schedules.filter(
-          // Filter out unsaved schedules
-          (schedule) => schedule.saved,
-        ).map((schedule) => ({
-          name: schedule.name,
-          // Get a list of the section IDs that are in this schedule
-          sections: [...new Set(schedule.meetings.map((m) => m.section.id))],
-        }));
-      }
+      const serializedSchedules = schedules.filter(
+        // Filter out unsaved schedules
+        (schedule) => schedule.saved,
+      ).map((schedule) => ({
+        name: schedule.name,
+        // Get a list of the section IDs that are in this schedule
+        sections: [...new Set(schedule.meetings.map((m) => m.section.id))],
+      }));
 
       fetch('sessions/save_schedules', {
         method: 'PUT',
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
@@ -10,6 +10,7 @@ import Schedule from '../../../types/Schedule';
 import { setSchedules } from '../../../redux/actions/schedules';
 import createThrottleFunction from '../../../utils/createThrottleFunction';
 import { parseAllMeetings } from '../../../redux/actions/courseCards';
+import SmallFastProgress from '../../SmallFastProgress';
 
 const throttle = createThrottleFunction();
 
@@ -24,6 +25,7 @@ const SchedulePreview: React.FC<SchedulePreviewProps> = ({ throttleTime = 10000 
   const dispatch = useDispatch();
   const schedules = useSelector<RootState, Schedule[]>((state) => state.schedules);
   const term = useSelector<RootState, string>((state) => state.term);
+  const [isLoadingSchedules, setIsLoadingSchedules] = React.useState(true);
 
   const scheduleListItems = schedules.length === 0
     ? <p className={styles.noSchedules}>{noSchedulesText}</p>
@@ -35,12 +37,12 @@ const SchedulePreview: React.FC<SchedulePreviewProps> = ({ throttleTime = 10000 
       fetch(`sessions/get_saved_schedules?term=${term}`).then(
         (res) => res.json(),
       ).then((arr: any[]): Schedule[] => arr.map((val) => ({
-        // TODO: Should we rename the schedule names if they're in the form of "Schedule #"?
         name: val.name,
         meetings: parseAllMeetings(val.sections),
         saved: true,
       }))).then((savedSchedules: Schedule[]) => {
         dispatch(setSchedules(savedSchedules));
+        setIsLoadingSchedules(false);
       });
     }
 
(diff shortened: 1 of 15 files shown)


autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,19 +1,88 @@
...
+        meetings: parseAllMeetings(val.sections),
+        saved: true,
+      }))).then((savedSchedules: Schedule[]) => {
+        dispatch(setSchedules(savedSchedules));
+      });
+    }
+
+    // on unmount, clear schedules
+    return (): void => {
+      dispatch(setSchedules([]));
```
gannonprudhomme, line 47: Instead of adding a `clearSchedules` action, I figured calling `setSchedules` with an empty array was a better solution

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,19 +1,88 @@
...
     ? <p className={styles.noSchedules}>No schedules available.</p>
     : schedules.map((schedule, idx) => <ScheduleListItem index={idx} key={schedule.name} />);
 
+  // Whenever the term changes, fetch and load saved schedules
+  React.useEffect(() => {
+    if (term) {
+      fetch(`sessions/get_saved_schedules?term=${term}`).then(
+        (res) => res.json(),
+      ).then((arr: any[]): Schedule[] => arr.map((val) => ({
+        // TODO: Should we rename the schedule names if they're in the form of "Schedule #"?
```
firejake308, line 36: What would you rename them to?

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
firejake308: Using an empty array is fine, but why do we have to clear schedules on unmount anyway? Is it like in case the user goes back to the landing page and then returns later? If os, what's the problem with that exactly?

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,19 +1,88 @@
...
+
+  // Call throttle to serialize and save the schedules anytime we make a change to the schedules
+  React.useEffect(() => {
+    if (!term) return;
+
+    // Serialize schedules and make API call
+    const saveSchedules = (): void => {
+      let serializedSchedules: { name: string; sections: number[] }[] = [];
+
+      if (schedules) {
```
firejake308, line 59: What is the difference between the case where `schedules` is `null` and the case where `schedules` is an empty array? The first would lead to making a fetch call with an empty array of schedules, and the first would just return. Why don't both have the same behavior?

firejake308 (commented):
Couple of questions skimming through the code

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
gannonprudhomme: It would basically just reorder the names, so if it was like `Schedule 3, Schedule 5, Schedule 1`, they'd now be `Schedule 1, Schedule 2, Schedule 3`. Honestly think it's super minor, just something I ran across that I figured I'd write down.

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
gannonprudhomme: I really did it b/c I saw that Ryan did it for course card sessions, he responded about it here: https://github.com/aggie-coding-club/Rev-Registration/pull/410#discussion_r513046517 (which I agree with)

gannonprudhomme:
Reminder for me to add a loading indicator when schedules are loading (from `sessions/get_saved_schedules`). Need to do history for now :'(

autoscheduler/user_sessions/views.py
```diff
@@ -97,6 +98,49 @@ def save_availabilities(request):
...
+@api_view(['GET'])
+def get_saved_schedules(request):
+    """ Returns the saved schedules from the session for the requested term """
+    term = request.query_params.get('term')
+
+    if not term:
+        return Response(status=400)
+
+    with retrieve_data_session(request) as data_session:
+        schedules = data_session.get(term, {}).get('schedules')
```
ryan-conn, line 110: Why doesn't this use `_get_state_from_session`

autoscheduler/user_sessions/views.py
```diff
@@ -97,6 +98,49 @@ def save_availabilities(request):
...
+            return Response([])
+
+        # Gather all of the section IDs
+        section_ids = set(sec for schedule in schedules for sec in schedule['sections'])
+
+        models = Section.objects.filter(
+            id__in=section_ids
+        ).select_related('instructor').prefetch_related('meetings')
+
+        sections_dict = {section.id: section for section in models.iterator()}
```
ryan-conn, line 122:
```suggestion
        sections_dict = {section.id: section for section in models}
```
From [the documentation](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#iterator):
> Also, use of iterator() causes previous prefetch_related() calls to be ignored since these two optimizations do not make sense together.

Also make sure this doesn't perform a query per section

autoscheduler/user_sessions/views.py
```diff
@@ -97,6 +98,49 @@ def save_availabilities(request):
...
+            sections = set(sections_dict[section] for section in sections)
+
+            return SectionSerializer(
+                sections, many=True, context={'skip_grades': True}
+            ).data
+
+        ret = [{
+            'name': schedule['name'],
+            'sections': serialize_schedule(schedule['sections']),
+        } for schedule in schedules]
```
ryan-conn, line 134: Is this not exactly the same as `_serialize_schedules` in the scheduler? If not, why does it have to be different and if so, why is it repeated here instead of imported

ryan-conn (changes requested):
A few things I noticed looking over this real fast, I'll do a more exhaustive review once you have the loading indicator added

autoscheduler/user_sessions/views.py
gannonprudhomme: Oh yeah duh - I knew it was going to be different from the normal `_get_state_from_session` but forgot I could just use it in the first part

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,21 +1,98 @@
...
```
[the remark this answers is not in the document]
gannonprudhomme: I had a reason at some point but I don't remember what it was, seems to be fine without it

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
gannonprudhomme: Had a reason at some point for having this check but I don't remember what it was so removing it

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,21 +1,98 @@
...
+    };
+  }, [term, dispatch]);
+
+  // Call throttle to serialize and save the schedules anytime we make a change to the schedules
+  React.useEffect(() => {
+    if (!term) return;
+
+    // Serialize schedules and make API call
+    const saveSchedules = (): void => {
+      let serializedSchedules: { name: string; sections: number[] }[] = [];
```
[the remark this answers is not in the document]
gannonprudhomme: Would you still like it to be an interface now that I removed this declaration?

autoscheduler/user_sessions/views.py
gannonprudhomme: Oh wait it's b/c _get_state_from_session returns a `Response` and we need it as a list - I'll instead change `_get_state_from_session` to just return the object and each individual route will form it as a `Response`

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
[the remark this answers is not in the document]
ryan-conn: It'll probably be fine but I'll take a look at it when you push your changes so leave this unresolved

autoscheduler/user_sessions/views.py
gannonprudhomme: Their input and output is slightly different which is why I didn't do it initially, but I'll change it to use it + apply all of the optimizations to it (removing the `models.iterator()`, prefetch, etc)

autoscheduler/frontend/src/redux/reducers/schedules.ts
```diff
@@ -39,8 +40,12 @@ export interface RenameScheduleAction {
   index: number;
   name: string;
 }
+export interface SetScheduleAction {
+  type: 'SET_SCHEDULE';
+  schedules: Schedule[];
+}
```
ryan-conn, line 46: Current name makes it seem like it replaces a single schedule rather than all of them (also change the SET_SCHEDULES declaration obviously)
```suggestion
export interface SetSchedulesAction {
  type: 'SET_SCHEDULES';
  schedules: Schedule[];
}
```

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -1,31 +1,112 @@
...
+    };
+  }, [term, dispatch]);
+
+  // Call throttle to serialize and save the schedules anytime we make a change to the schedules
+  React.useEffect(() => {
+    if (!term) return;
+
+    // Serialize schedules and make API call
+    const saveSchedules = (): void => {
+      const serializedSchedules = schedules.filter(
```
ryan-conn, line 65: I'd still create an interface for this since it's being used for the tests and an interface would prevent headaches if we refactor this or add more properties to schedules in the future for whatever reason, it also makes the expected format more explicit

ryan-conn (changes requested):
A couple small comments (except it saving courses as availabilties, that's pretty major) and this looks good

autoscheduler/user_sessions/views.py
```diff
@@ -89,14 +96,43 @@ def get_full_name(request):
 @api_view(['GET'])
 def get_saved_availabilities(request):
     """ Returns the saved availabities from the session for the requested term"""
-    return _get_state_from_session(request, 'availabilities')
+    availabilities = _get_state_from_session(request, 'courses')
```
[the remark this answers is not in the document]
gannonprudhomme: shit good catch

autoscheduler/frontend/src/types/APIRequests.ts
```diff
@@ -0,0 +1,12 @@
+// Module containing type definitions for API Reuqests
```
ryan-conn, line 1:
```suggestion
// Module containing type definitions for API Requests
```
tyop

ryan-conn (changes requested):
Just a typo fix and we can merge this

ryan-conn (approved):

PR #420, opened 2020-10-25
Added dynamic CSS rules to expanded row

### LINKED ISSUE #1: Setup Django basics
Setup the basics of the django repo

firejake308:
## Description

TODO

## Rationale

If necessary, defend any changes that you made.

## How to test

TODO

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|

## Related tasks

Closes #1

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
@@ -1,4 +1,3 @@
-/* eslint-disable no-labels */
 import * as React from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import * as Cookies from 'js-cookie';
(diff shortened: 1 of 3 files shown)


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -20,10 +21,53 @@ const CourseSelectColumn: React.FC = () => {
...
+    if (expandedRowRef.current?.children[0].clientHeight < 500 - 8) {
+      const styleSheet = document.styleSheets[styleSheetNum] as CSSStyleSheet;
+      if (styleSheet) {
+        styleSheet.deleteRule(ruleNum);
+        styleSheet.insertRule(
+          `.${styles.expandedRow} {
+            overflow-y: visible;
+          }`, ruleNum,
+        );
+      } else throw Error('styleSheet is undefined');
```
ryan-conn, line 57: A couple things about this error:
- How is this better than just silently ignoring the fact that the ref isn't current?
- If it is better, why doesn't the `else` part of this expression throw an error?

ryan-conn: Also you can just ignore this if we end up using the changes from my branch

ryan-conn (changes requested):
Thanks a lot for this, I was trying to do something similar before but the refs didn't like to stay updated, probably because I was attaching the ref to the card itself rather than a wrapper.

The biggest thing I have to say is that I don't think it's necessary to mess with the CSS rules directly, see PR #422 for the changes I made to clean this up. Then if you think the changes I made are sufficient we can merge my branch into your branch into my branch into master. Hooray!

gannonprudhomme:
Shouldn't this have been merged in `frontend/full-height-course-cards`, not `frontend/full-height-course-cards-test`?

ryan-conn:
> Shouldn't this have been merged in `frontend/full-height-course-cards`, not `frontend/full-height-course-cards-test`?

Yeah but it was based on the test branch which is slightly different and I didn't want to make a merge conflict. I've rebased the actual branch now so it's fine

PR #421, opened 2020-10-25
Section select overrides include full sections

ThatJuanGuy:
## Description

Previously, when selecting a full section in section select and generating a schedule when "include full sections" was not selected, the section manually selected in section select would not be included. This PR makes it so that all sections manually selected in section select can be included in schedules regardless of whether include full sections is selected or not

## How to test

Select some courses and go to section select. Select full sections for that course. After you have selected the sections and courses, make sure "include full sections" is not selected on the configure card. Generate schedules. If the schedule generated includes the full sections you manually selected, it should be working. I also added one backend test.

## Related tasks

Doesn't close anything but a temp quality of life improvement meanwhile the "include full sections" checkbox is moved into the basic select of course cards and the configure card is removed.

### DIFF
--- a/autoscheduler/scheduler/create_schedules.py
+++ b/autoscheduler/scheduler/create_schedules.py
@@ -36,7 +36,7 @@ def _get_meetings(course: CourseFilter, term: str, include_full: bool,
     # Get id for each valid section to filter and order meeting data
     # Also removes full sections if include_full is False
     sections = sections.values('id', 'current_enrollment', 'max_enrollment')
-    #if manually selected don't check if section is full before adding
+    # if manually selected don't check if section is full before adding
     if course.section_nums or include_full:
         section_ids = set(section['id'] for section in sections)
     else:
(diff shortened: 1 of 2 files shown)


ryan-conn (dismissed):
Small optimization and this looks good

autoscheduler/scheduler/create_schedules.py
```diff
@@ -36,9 +36,12 @@ def _get_meetings(course: CourseFilter, term: str, include_full: bool,
     # Get id for each valid section to filter and order meeting data
     # Also removes full sections if include_full is False
     sections = sections.values('id', 'current_enrollment', 'max_enrollment')
-    section_ids = set(section['id'] for section in sections
-                      if include_full or
-                      section['current_enrollment'] < section['max_enrollment'])
+    #if manually selected don't check if section is full before adding
```
gannonprudhomme, line 39: god this is such a small nit pick (and really should be a linter rule) 
```suggestion
    # if manually selected don't check if section is full before adding
```

gannonprudhomme (changes requested):

gannonprudhomme (approved):

PR #427, opened 2020-10-26
Select all section select

### LINKED ISSUE #363: Add select all to SectionSelect
Figma_YTfeeGDlYj

eelyort:
## Description

It adds a select/deselect all button to section select.

## How to test

For any term and class, click the "section" button and you should see the button.

## Screenshots

New Version:
myGif

## Related tasks

Closes #363


gannonprudhomme:
Change you merge master into this? As you have some conflicts atm

Also recap from a meeting, but I believe we want to decrease the font text size and make it lightly less bold and more gray.

ryan-conn:
To fix the button-in-button issue with the checkbox inside the toggle button, try using a checkbox with a `<FormControlLabel />` instead: [MUI documentation page](https://material-ui.com/components/checkboxes/)

To style the text, the `label` component of the form control label can take a typography component instead of just a string, that you can provide with `variant="button"` (should be able to replace most of the css) to style in a similar way to it is right now

eelyort:
I made the change to remove the double ripple, replaced the gif in the pr with the new version, you can also see it below:
myGif

autoscheduler/frontend/src/tests/ui/SectionSelect.test.tsx
```diff
@@ -67,6 +69,124 @@ function makeCourseCard(...args: any): CourseCardOptions {
...
+        fireEvent.click(getByText('201'));
+        fireEvent.click(getByText('202'));
+        fireEvent.click(getByText('201'));
+
+        // assert
+        expect(getAllByDisplayValue('allOff')).toHaveLength(1);
+      });
+    });
+
+    test('selects all sections', async () => {
```
gannonprudhomme, line 146:
```suggestion
    test('when we select all sections', async () => {
```
the `describe` and their children should read like a sentence - so for this it'd be `select all is unchecked when we select all sections`

On that note - I believe this one needs test should be under the `is checked` block

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -24,6 +49,7 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
     );
   }
 
+  let countSelected = 0;
```
gannonprudhomme, line 52: nitpick: maybe rename this to `selectedCount` or `numSelected`? `countSelected` sounds like it would be a function name to me

gannonprudhomme (changes requested):
I think really the main problem (which is minor) I have with this is that it feels like there's too much space between the Customization Level buttons to the select all button. I think the easiest way of fixing this is to decrease the height of the Select all button (which I also think is slightly too large). Functionality wise this is great tho! Fantastic job with this, especially the tests

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -261,6 +261,22 @@ ThunkAction<void, RootState, undefined, UpdateCourseAction> {
   };
 }
 
+// select all
```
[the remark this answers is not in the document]
eelyort: image
How's this?

autoscheduler/frontend/src/tests/ui/SectionSelect.test.tsx
eelyort: Sorry, this test is actually under the broad describe, so it would read `select all selects all sections`.
I changed it to `select all selects all sections when clicked` since writing that out felt weird.

The reason I didn't put it in the `is checked` block is because the `is checked` and `is unchecked` blocks were meant to be a reactive thing, while this test is more an active thing.

autoscheduler/frontend/src/tests/ui/SectionSelect.test.tsx
```diff
@@ -359,7 +479,7 @@ describe('SectionSelect', () => {
       }, '201931'));
 
       // assert
-      expect(getAllByText(/(EXAM)|(LEC)/)).toHaveLength(2);
+      expect(getAllByText(/(EXAM)|(LEC$)/)).toHaveLength(2);
```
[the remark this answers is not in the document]
eelyort: Its a regex thing that denotes end of line. It was required because that selector would return the `LEC` in `SELECT`.

autoscheduler/frontend/src/redux/actions/courseCards.ts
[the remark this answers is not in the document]
gannonprudhomme: Yep that works!

autoscheduler/frontend/src/tests/ui/SectionSelect.test.tsx
gannonprudhomme: Sounds good, that makes sense

gannonprudhomme (approved):

PR #429, opened 2020-10-26
Added untracked migration

gannonprudhomme:
Ryan forgot to track a migration in #382. This tracks that migration.


ryan-conn (approved):
Oops haha I'm a dumb dummy who is dumb

PR #434, opened 2020-10-27
Removed asyncio from requirements

gannonprudhomme:
## Description

For some reason, GCP keeps failing b/c asyncio is listed as a requirement, but it's built into Python since 3.4 (I think) so there's no need for it to be a requirement.


ryan-conn (approved):
Hopefully this will fix it, I've heard that app engine doesn't support asyncio at all but since it's mostly been working we should just keep an eye out and make sure we don't run into the same issue on future deploys

PR #437, opened 2020-10-28
Frontend/am pm

### LINKED ISSUE #371: Change time formatting to use AM/PM on frontend
Due to the increase in late classes (or perceived increase), I think we can no longer be confident that people will be able to tell if a time is AM/PM. For instance, this exam is 8:30 PM - 10:30 PM, but just by seeing that I would not guess that.

If only we Americans used 24 hour time :cry: (haha jk screw 24 hour time AM/PM RULES)

ajayramsunder:
## Description

This shows the AM - PM within the meetings.

## Rationale

This is necessary because people may get confused if the class is in the morning or in the evening. 

## How to test

The best way to test this is to see the AM and PM that was added when adding the classes that are being displayed. 

## Screenshots

Before
image

After
image

## Related tasks

Closes #371

### DIFF
--- a/autoscheduler/frontend/src/utils/timeUtil.ts
+++ b/autoscheduler/frontend/src/utils/timeUtil.ts
@@ -11,10 +11,7 @@ export const LAST_HOUR = 22;
 export function formatTime(h: number, m: number, use24Hour = false, padZeroes = false): string {
   let amOrpm = '';
   if (!use24Hour) {
-    amOrpm = 'AM';
-    if (h >= 12) {
-      amOrpm = 'PM';
-    }
+    amOrpm = h < 12 ? 'AM' : 'PM';
   }
 
   const formattedHours = use24Hour ? h : (h - 1) % 12 + 1;
--- a/autoscheduler/frontend/src/utils/timeUtil.ts
+++ b/autoscheduler/frontend/src/utils/timeUtil.ts
@@ -9,9 +9,17 @@ export const LAST_HOUR = 22;
  * @param use24Hour if true, applies 24-hour format instead of the default 12-hour format
  */
 export function formatTime(h: number, m: number, use24Hour = false, padZeroes = false): string {
+  let amOrpm = '';
+  if (!use24Hour) {
+    amOrpm = 'AM';
+    if (h >= 12) {
+      amOrpm = 'PM';
+    }
+  }
+
   const formattedHours = use24Hour ? h : (h - 1) % 12 + 1;
   const padZero = new Intl.NumberFormat('en-US', { minimumIntegerDigits: 2 });
-  return `${padZeroes ? padZero.format(formattedHours) : formattedHours}:${padZero.format(m)}`;
+  return `${padZeroes ? padZero.format(formattedHours) : formattedHours}:${padZero.format(m)}${amOrpm}`;
 }
 
 /**


autoscheduler/frontend/src/utils/sortMeetingFunction.tsx
```diff
@@ -0,0 +1,8 @@
+import Meeting from '../types/Meeting';
+
+// sorts meeetings based on meeting type
+const sortMeeting = (a: Meeting, b: Meeting): number => {
```
firejake308, line 4: Do these changes need to be in here?

autoscheduler/frontend/src/utils/timeUtil.ts
```diff
@@ -9,11 +9,13 @@ export const LAST_HOUR = 22;
  * @param use24Hour if true, applies 24-hour format instead of the default 12-hour format
  */
 export function formatTime(h: number, m: number, use24Hour = false, padZeroes = false): string {
+  let amOrpm = 'AM';
```
firejake308, line 12:
```suggestion
  let amOrPm = 'AM';
```

autoscheduler/frontend/src/utils/timeUtil.ts
```diff
@@ -9,11 +9,13 @@ export const LAST_HOUR = 22;
...
+  if (h >= 13) {
```
firejake308, line 13:
```suggestion
  if (h >= 12) {
```
right? Because 12 noon is 12 PM?

autoscheduler/frontend/src/utils/timeUtil.ts
```diff
@@ -9,11 +9,13 @@ export const LAST_HOUR = 22;
...
+    amOrpm = 'PM';
+  }
   const formattedHours = use24Hour ? h : (h - 1) % 12 + 1;
-
   const padZero = new Intl.NumberFormat('en-US', { minimumIntegerDigits: 2 });
-
-  return `${padZeroes ? padZero.format(formattedHours) : formattedHours}:${padZero.format(m)}`;
+  return `${padZeroes ? padZero.format(formattedHours) : formattedHours}:${padZero.format(m)}${amOrpm}`;
```
firejake308, line 18: Does this need a space before the AM/PM text?

firejake308 (changes requested):

autoscheduler/frontend/src/utils/sortMeetingFunction.tsx
gannonprudhomme: nope, he needs to reset to master and remake the changes - this includes the changes from #385

gannonprudhomme (changes requested):
You'll need to run `npm run test` locally and fix all of the failing tests from this. Probably easiest to go one file at a time (you can run `npm run test FileName.test.tsx` to run a specific file)

gannonprudhomme:
Cleaned this up (should look identical to what it was previously) just cause I hated the 26 commits that really should've just been one.

autoscheduler/frontend/src/utils/timeUtil.ts
```diff
@@ -9,9 +9,17 @@ export const LAST_HOUR = 22;
  * @param use24Hour if true, applies 24-hour format instead of the default 12-hour format
  */
 export function formatTime(h: number, m: number, use24Hour = false, padZeroes = false): string {
+  let amOrpm = '';
+  if (!use24Hour) {
+    amOrpm = 'AM';
+    if (h >= 12) {
+      amOrpm = 'PM';
+    }
```
ryan-conn, line 17: Might as well switch this to an if/else or ternary instead of always making it AM first

ryan-conn (changes requested):

ryan-conn (approved):

PR #442, opened 2020-10-28
Made text when no schedules have been created more helpful

ryan-conn:
## Description

Makes the no schedules text seem less like an error and instead direct users to generate schedules.

## Rationale

I thought this wording was good because it directs users to the button and tells them that schedules will be generated using their course cards. Could change classes to course though, idk

## How to test

use ur eyes and look at the screen

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|
| image | image |

## Related tasks
Nothing, I thought this would be helpful

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
@@ -7,7 +7,7 @@ import * as styles from './SchedulePreview.css';
 import ScheduleListItem from './ScheduleListItem/ScheduleListItem';
 import Schedule from '../../../types/Schedule';
 
-export const noSchedulesText = "Click Generate Schedules to find schedules with the classes you've added.";
+export const noSchedulesText = "Click Generate Schedules to find schedules with the courses you've added.";
 
 const SchedulePreview: React.FC = () => {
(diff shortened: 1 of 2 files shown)


autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
```diff
@@ -7,11 +7,13 @@ import * as styles from './SchedulePreview.css';
 import ScheduleListItem from './ScheduleListItem/ScheduleListItem';
 import Schedule from '../../../types/Schedule';
 
+export const noSchedulesText = "Click Generate Schedules to find schedules with the classes you've added.";
```
gannonprudhomme, line 10:
```suggestion
export const noSchedulesText = "Click Generate Schedules to find schedules with the courses you've added.";
```

B/c we say `Add course` for the add course card button - per you

gannonprudhomme (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/SchedulePreview.tsx
ryan-conn: yeah fair enough, I was looking for feedback on this but I like it more too

gannonprudhomme (approved):

PR #445, opened 2020-10-30
Added open graph/twitter meta tags + banner image

### LINKED ISSUE #332: Add Twitter metadata
Having rich previews will make out website more marketable to new users. To do so, we could use [Twitter's summary metadata](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary) to achieve something that looks like this:
image

gannonprudhomme:
## Description

First, this adds the OpenGraph protocol tags so that rich preview works on various sites. It also adds the description tag, which is what search engines use to display information below it. The banner included is 1.97:1 aspect ratio (1200x597), which is what various sites suggest (although there doesn't really seem to be a concrete standard).

## How to test

Honestly you can't really test this without it being deployed, so I'm just going to post screenshots of it.

## Screenshots

These previews may be slightly off, but they give a gist pretty well. If we haven't deployed to GCP again by the time you're viewing this, you can check out what it looks like by going to https://metatags.io

Also note that these screenshots are old, and the actual banner we included has a different font (and slightly different size). It's close enough though.

|Google (metatags.io)|Facebook (official validator)|Twitter (official validator)|LinkedIn (official validator)|Slack (actual)|Discord (actual)|
|--|--|--|--|--|--|
|chrome_Dz3Crlw3vs|chrome_YUtpmjj0rQ|chrome_ZQeeD4jugU|chrome_M9m4Bqwk75|slack_0VnCx0axWc|Discord_Pn2METg1pP|

## Related tasks

Closes #332


ryan-conn (approved):
Looks good to me, everything works and I don't have problems with any of the wording. If we decide to change some of it in the meeting or something we can make a really quick PR in the future but I don't think there's anything important enough to prevent us from merging this now

PR #450, opened 2020-10-30
Keep route on login

ryan-conn:
## Description

Redirects you back to your current page when logging in

## Rationale

We had said we wanted this functionality but didn't know how to do it, so I decided to implement it after seeing the solution looking up an unrelated problem

## How to test

Try logging in from the homepage, scheduling page, some 404 page, etc.

## Screenshots
owen wilson

## Related tasks
n/a


ThatJuanGuy (approved):
Really nice job on this. You added in the feature in a much simpler way than I thought was possible. I just tested this and it seems to be working so I'm going to approve it.

PR #455, opened 2020-11-02
Fixed mini schedule sizing problems

firejake308:
## Description

There was a weird 1-px margin at the top of the mini schedules, and when I went to fix that, I realized that the times didn't quite match up for classes later in the day. Consequently, I also included the fix to that bug as well.

## Rationale

Removing the random 1px margin should help make the border look better. Also, tbh, I no longer remember what its original purpose was. And fixing the bug is good because it makes the mini schedules more accurate, although tbf, users probably wouldn't look hard enough to notice.

## How to test

Generate a schedule. Zoom in on the mini schedule. Look if there is any space between the maroon headera and the 8:00 bar. Then, count out the hours to a class in the middle of the day. Make sure it matches what you see in the big schedule.

## Screenshots

The dark green class is KINE 199-242, which should go from 4:15 to 5:05. On the old mini schedule, it looks like it starts at 5.

|Before|After|
|--|--|
|image|image|

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/MiniSchedule/MiniSchedule.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/MiniSchedule/MiniSchedule.css
@@ -11,7 +11,7 @@
 
 .hour-marker {
     flex-grow: 1;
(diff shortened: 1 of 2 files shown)


gannonprudhomme:
Can you merge master into this? Just so we can see this with #417

gannonprudhomme (approved):

PR #457, opened 2020-11-04
Added GNU GPL v3 License

### LINKED ISSUE #428: Pick a license
Related to the About popup, we need to pick a license and set it on the repo. Once I get a few options I'll send them in the discord and we can discuss

gannonprudhomme:
## Description

Adds the license. I don't think we need to edit anything in this, but we might have to add a copyright notice in each file? Which I really don't want to do lol. Followed [these steps](https://docs.github.com/en/free-pro-team@latest/github/building-a-strong-community/adding-a-license-to-a-repository) from GitHub docs.

Closes #428

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,674 @@
+                    GNU GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
+
+ Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
+ Everyone is permitted to copy and distribute verbatim copies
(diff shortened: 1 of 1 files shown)


ryan-conn (approved):
According to [this stackoverflow post with 5 whole votes](https://stackoverflow.com/questions/1632066/how-to-minimize-the-amount-of-place-used-by-gpl-copyright-notice), we should be fine without adding the notice to each file, and I'd prefer we don't do it because it'll be a huge pain for not a lot of benefit.
As such I'll approve this, just the license file should be fine

PR #463, opened 2020-11-06
New course colors

### LINKED ISSUE #374: New color scheme for sections on schedule
The current color scheme for the classes is a little dificult for users to quickly interpret, since all classes are some shade of maroon. We could replace them with the colors of the rainbow, which would make it much faster to differentiate between classes, but it risks losing visual appeal. Therefore, my plan is ot use [colormind.io](https://colormind.io) to look for colors that are distributed along the rainbow but "fit" well into our overall maroon theme.

firejake308:
## Description

The current system renders all meetings on the schedule using variants of maroon. This is aesthetically consistent, but makes it difficult to quickly differentiate between different classes. Therefore, we are considering a new color scheme with more diverse colors. By keeping all of the colors on the dark side, we should be able to maintain aesthetic consistency.

## Rationale

So glad I made that `useMeetingColors` hook now

## How to test

Add like 10 courses to your schedule

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #374


gannonprudhomme:
So I used this website: https://www.color-blindness.com/coblis-color-blindness-simulator/ to simulate color blindness (you can read about the different types [here](https://www.nei.nih.gov/learn-about-eye-health/eye-conditions-and-diseases/color-blindness/types-color-blindness)) and it seems like these two pairs are gonna be the hardest to differentiate:
|A|B|
|--|--|
|chrome_lypMJWF0N8|chrome_ho4wjOGchv|

gannonprudhomme:
Also while I do like having the color order be like a gradient, it'll probably be best to have the colors that are alike to be farther apart from each other (basically mix them up so it's not a gradient).

firejake308:
So the red and purple honestly looked kind of similar if I just focused away from the screen/zoomed out, and my solution was to pump up the amount of red in the red and decrease the amount of green and blue channels. I didn't want to decrease the blue channel alone, because that made it too similar to orange.

As for the orange and the light brown, they look pretty different to me with full-color vision, but it seems that under red-blind and green-blind conditions, they look essentially the same. I guess that without red or green cones, you'd be relying disproprtionately on the blue channel, which is similar for both. I tried increasing the amount of blue in the light brown, which made it turn sort of pink-ish, but it was still differentiable from all other colors and also differentiable under color-blind conditions.

Screenshot of all colors:
image

gannonprudhomme (changes requested):
From the meeting, we said to replace the tan / "light brown/pink" to something else

firejake308:
Changed tan to pink:
|Old|New|
|--|--|
|image|image|

gannonprudhomme:
Just need to reorganize them and we'll be good

firejake308:
Renamed tan to yellow, moved dark brown to top and red to bottom

image

autoscheduler/frontend/src/components/SchedulingPage/Schedule/meetingColors.ts
```diff
@@ -2,8 +2,15 @@ import { useSelector } from 'react-redux';
...
-  '#871b1e', '#9a1d26', '#c2777d', '#9f494b', '#b76778',
+  '#f2c05b', // yellow
+  '#612b0d', // dark brown
+  '#2d683e', // green
+  '#ec4631', // orange
+  '#227d9c', // blue
+  '#83205c', // purple
+  '#db728e', // light brown/pink
+  '#249889', // turquoise
+  '#c82041', // red
```
gannonprudhomme, line 13:
```suggestion
   '#f2c05b', // yellow
  '#2d683e', // green
  '#ec4631', // orange
  '#227d9c', // blue
  '#83205c', // purple
  '#db728e', // light brown/pink
  '#249889', // turquoise
  '#c82041', // red
  '#612b0d', // dark brown
```
Moves brown to be the last one

firejake308:
Final screenshot:
image

gannonprudhomme (approved):

PR #464, opened 2020-11-10
Added a last updated indicator

### LINKED ISSUE #336: Add a "last updated at" indicator somewhere on the website for course information
We should add some an indicator somewhere on the site (maybe on the navbar?) that says when the course information was last updated.

An easy way of doing this would probably be to just have a table with a single row / column that says when `scrape_courses` was last ran. The route would then just retrieve this value and the frontend would display it accordingly

### Add the model to the backend

So I'm not really sure how we want to handle this, but my thought process would be that the easiest way to do it would be to have a model that just stores a term and date, and it's ID / primary key is the term. So it'll basically just show when the data was last updated for each term.

To do this, you'll need to:

1) Create the model
     - You'll want to create the file in `autoscheduler/scraper/models/` since this is related to scraping.
     - Will probably want to call this `LastUpdated` or something like that.
     - You'll want to have two fields, a `term` and a `date`, where the `term` is a string that is also it's ID / primary key.
     - You should be able to get a gist of how to make the model by looking at the other models in that folder, but lmk if you have any questions.
2) Run `python manage.py makemigrations` to create a migration file.
    - This is how we record changes to the database (new tables, changing columns, etc) so after this change is merged we can apply the same changes to our database.
    - Make sure you commit the file! It'll be located in `autoscheduler/scraper/migrations/`
3) Run `python manage.py migrate` to apply the migration to your database

### Save/update these models in `scrape_courses`

Now that the model is created, you'll need to actually save them whenever `scrape_courses` is ran.

- The file is located in `autoscheduler/scraper/management/commands/scrape_courses.py`
- You'll simply need to scrape over all of the terms that are being scraped, then create a `LastUpdated` model(s) that have the term as the respective term and the current datetime as the date model.
     - Note that there can be a singular term stored in `term`, or multiple terms in `terms`, so make sure you handle both situations
     - This is b/c you can scrape a singular term with `scrape_courses -t 202031`, but can also scrape a group of terms, such as `scrape_courses -y 2020` to scrape 2020, `scrape_courses -r` to do the most recent semester, or even just `scrape_courses` without args, which scrapes all terms possible.
     - You'll want to do something called an "upsert" - that is, insert if it doesn't exist, or update the model if it is.
         - The easiest way to do this is do just call `LastUpdated.objects.delete()` then `LastUpdated.objects.bulk_create(objects)`. Sounds counter-intuitive, but it's what we're doing elsewhere and it seems to work for us (Django's `bulk_update` is pretty slow from what we've seen)
     - You'll want to do this around line 331, after we call `save_models`

### Add a route that returns the respective model on the backend

We'll probably want to set this on a per-term basis? Or maybe a per-semester basis (since all locations are scraped each semester)

You should create a route like `api/get_last_updated` which takes a term query parameter. A potential query would look something like `api/get_lasted_updated?term=202031`

1) Create the API view in `views.py`
     - Since this is a `scraper` route, you'll do this in `autoscheduler/scraper/views.py`. Note that `scraper`, `scheduler`, and `user_sessions` all have a `views.py` file
     - You'll create a `RetrieveLastUpdated` view at the bottom of the file. It will take be a `get` request, so as such you can base this off of the `RetrieveCourseView` at the top of the file.
     - All you need to do is get the `term` 
2) Add the URL to `urls.py`
    - Again since this is a `scraper` route, you'll do this in `autoscheduler/scraper/urls.py`
    - You'll add a line to the bottom of the `urlpatterns` list that'll look something like `path('get_last_updated`, RetrieveLastUpdated.as_view())`
    - Note that the `api/` isn't in here - that's b/c all of the routes in `scraper` automatically have the `api/` appended to it. You can see this if you go to `autoscheduler/autoscheduler/urls.py`.

### Add the component on the frontend + make it fetch the data

You'll then want to add a component on the frontend (and make it display in the navbar) that fetches the last updated data and displays it

We'll also probably want to have some mechanism for updating this, since in peak times we hope to scrape the data every 15 minutes. Maybe we can try to update this value every minute? Whatever you think is best really

### Add frontend tests

You'll need to create some basic tests on the frontend. Since these are UI tests, you'll want to put them in `autoscheduler/frontend/src/tests/ui/LastUpdated.test.tsx` or something like that. When you get around to this message in Discord and we'll find a good example test.

### Add backend tests

The tests should be relatively straightforward. There are a lot of examples in `autoscheduler/scraper/tests/api_tests.py` (which is also in the folder where you'll be making these tests)

- Since `api_tests.py` is already huge, make a separate test file called like `api_last_updated_tests.py` or something like that.
- Note to run these tests, you can run `pytest` in the `autoscheduler/` folder (where the `pytest.ini` is located)

gannonprudhomme:
## Description

Adds a last updated indicator to the site. Does so on the backend by adding a `Term` model which stores when it was last scraped.

I also made a command called `fill_terms` that will just iterate through all of the terms and create models for them. As described in the docstring, this is just a command I'm going to run once when the migration is applied so that the `Term` table is filled out & I don't have to run `scrape_courses` to get every term to show up again (everyone will probably want to run it locally too). 

Also note that I added `lastTerm` state for `LastUpdatedAt`, so we could detect when the term changed and set the term to null accordingly.

## How to test

Run `fill_courses` or `scrape_courses` with whatever options and switch between terms and such

## Screenshots

|Before|After|
|--|--|
|chrome_VcGxJoqoVI|chrome_zkX2Qj5b6d|

## Related tasks

Closes #336


autoscheduler/frontend/src/components/LastUpdatedAt.tsx
```diff
@@ -0,0 +1,86 @@
...
+  const day = prev.getDate(); // gets the day of the month
+
+  return `on ${month}/${day}/${year}`;
+}
+
+/**
+ * Displays when the current term was last updated, e.g. "Last updated 3 hours ago."
+ */
+const LastUpdatedAt: React.FC = () => {
+  const [lastUpdated, setLastUpdated] = React.useState(null);
```
ryan-conn, line 49: If an initial state is "empty" you should give it an explicit type and the default should be undefined:
```suggestion
  const [lastUpdated, setLastUpdated] = React.useState<Date>();
```

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -369,5 +395,6 @@ def handle(self, *args, **options):
 
         instructors, sections, meetings, courses = get_course_data(depts_terms)
         save_models(instructors, sections, meetings, courses, term, terms, options)
+        save_terms(term, terms, options)
```
ryan-conn, line 398: Not really necessary in this PR but I would like if we could refactor this file to just use `terms`, I don't see why we couldn't set `terms = [options['term']]` instead of making a single term a special case

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -329,6 +330,31 @@ def save_models(instructors: List[Instructor], sections: List[Section], # pylint
...
+
+            if options['year'] or options['recent']:
+                queryset = Term.objects.filter(code__in=terms)
+            else: # If no options are provided, then we're scraping all terms
+                queryset = Term.objects.all()
+
+        queryset.delete()
+        Term.objects.bulk_create(term_models)
+
+        term_len = len(terms) if terms else 1
```
ryan-conn, line 354: This code is kinda ugly and would be a lot nicer by just providing `terms` like I describe below

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -35,15 +39,15 @@ def get_recent_semesters(now=datetime.now()) -> List[str]:
...
     if now < summer_fall_reg_start:
         return [f"{year}1"]
 
-    # Between [04/01/2020, 11/05/2020]
+    # Between [03/22/2020, 10/26/2020]
     if summer_fall_reg_start <= now < spring_reg_start:
         return [f"{year}2", f"{year}3"]
 
-    # Between [11/05/2020, 04/01/2021]
+    # Between [10/26/2020, 03/22/2021]
```
ryan-conn, line 50:
```suggestion
    # Between [10/26/2020, 12/31/2020]
```
Isn't this right?

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -19,8 +19,12 @@ def get_all_terms(year: int = -1) -> List[str]:
     semesters = range(1, 4)
     locations = range(1, 4)
 
-    return [f"{year}{semester}{location}"
-            for year, semester, location in product(years, semesters, locations)]
+    ret = [f"{year}{semester}{location}"
+           for year, semester, location in product(years, semesters, locations)]
+
+    ret.extend(get_recent_terms())
```
ryan-conn, line 25: This will always scrape recent terms even if you provide a year

autoscheduler/scraper/views.py
```diff
@@ -112,3 +114,20 @@ def get(self, request):
...
+    """
+    term = request.query_params.get('term')
+
+    if not term:
+        return Response(status=400)
+
+    try:
+        last_updated = Term.objects.get(code=term).last_updated
+    except Term.DoesNotExist:
+        last_updated = datetime(2020, 11, 1)
```
ryan-conn, line 131: This should return undefined so it doesn't actually show anything on the frontend since the current output is misleading

autoscheduler/scraper/views.py
```diff
@@ -40,14 +42,14 @@ class RetrieveTermView(generics.ListAPIView):
         This view returns all the terms
     """
```
ryan-conn, line 43: The changes you made broke the sort function somehow at least for me, here's what my term list looks like now:
image
Do you have the same problem? It seems to order terms for a semester correctly for around half of them which is why the test with terms for 2020 didn't break

ryan-conn (changes requested):
We should use an interval or something to periodically update the time displayed since especially if the time is in seconds it will get outdated pretty fast. Also maybe periodically refetch the API in case scraping has happened again? I feel like most users would expect they'd have to refresh to get updated courses though so idk how much this would help

Also make sure you fix linter/test errors

firejake308:
Do y'all think it would look better if it was 2 separate lines like:
```
Last updated:
6 minutes ago
```
I feel like it takes up a bit too much horizontal space right now

autoscheduler/scraper/views.py
gannonprudhomme: Yep the same is happening to me, will look into it

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Yeah I agree - it's super obvious in hindsight idk why I didn't do it before

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+  // If it was less than 24 hours ago, say it was x hours ago
+  const hours = minutes / 60;
+  if (hours < 24) {
+    return `${hours.toFixed(0)} hours ago`;
+  }
+
+  // Afterwards, say it was x days ago? Or on some date
+  const days = hours / 24;
+  if (days < 7) {
```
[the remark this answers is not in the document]
gannonprudhomme: Honestly it probably won't ever be shorter than a day (really it'll never be shorter than 3 hours) so I don't think showing weeks is necessary. I could extend it from 1 week to 2 though

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
```diff
@@ -0,0 +1,86 @@
+import * as React from 'react';
+import { Typography } from '@material-ui/core';
+import { useSelector } from 'react-redux';
+import { RootState } from '../redux/reducer';
+
+/** Given a date, formats it into a string for how long ago it was.
+ *  Note that now is only included for mocking the current time for tests.
+ */
+export function getLastUpdatedAtText(prev: Date, now = new Date()): string {
+  if (!prev) return '';
```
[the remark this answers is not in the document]
gannonprudhomme: Probably at some point during making this I was passing a falsy value and an error like `couldn't find function 'getTime' on undefined object` was raised so I threw it in there. Will remove.

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: I'm just gonna do this now - we probably wouldn't get around to it otherwise and it's really bugging me now that you pointed it now

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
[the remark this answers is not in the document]
gannonprudhomme: Yeah so as discussed in the meeting I'm an idiot and forgot about previous terms lmao

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
```diff
@@ -0,0 +1,114 @@
...
+
+        const date = new Date(dateStr);
+        setLastUpdated(date);
+        setLastUpdatedText(getLastUpdatedAtText(date));
+      });
+    };
+
+    fetchLastUpdated();
+
+    const interval = setInterval(() => fetchLastUpdated, 5 * 60 * 1000); // update every 5 minutes
```
ryan-conn, line 81: This doesn't actually periodically call the fetch again, it seems to do with the fact that you call `fetchLastUpdated` instead of doing the logic in the arrow function since it works when I remove the function definition. I'm not sure exactly why your current implementation doesn't work though, as far as I know it should be fine but there are probably some javascript shenanigans happening behind the scenes

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -329,6 +326,27 @@ def save_models(instructors: List[Instructor], sections: List[Section], # pylint
...
+    with transaction.atomic():
+        # Bulk upsert
+        term_models = [Term(code=term, last_updated=now) for term in terms]
+
+        if options['year'] or options['recent']:
+            queryset = Term.objects.filter(code__in=terms)
+        else: # If no options are provided, then we're scraping all terms
+            queryset = Term.objects.all()
+
+        queryset.delete()
```
ryan-conn, line 343: This deletes all term models other than the ones you're scraping if you use the `-t` flag

autoscheduler/scraper/management/commands/utils/scraper_utils.py
ryan-conn: Make sure you address this comment

autoscheduler/scraper/serializers.py
```diff
@@ -94,22 +94,29 @@ def campus_num_to_string(campus_num):
 class TermSerializer(serializers.ModelSerializer):
     """ Serializes a department into an object with information needed by /api/terms """
```
ryan-conn, line 95: Update based on the new functionality

ryan-conn (changes requested):
One small thing I'd like to see changed is the text when it's been less than a minute since the last update, the current way is a bit unnatural imo. I'd prefer "less than a minute ago" instead of what it currently looks like:
image

The current implementation also always pluralizes the time unit, for example:
image
It should say "1 minute/hour/day" ago instead

autoscheduler/scraper/management/commands/scrape_courses.py
gannonprudhomme: Great catch - man this is why we need `scrape_courses` tests

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
gannonprudhomme: Changed it from `fetchLastUpdated` to `fetchLastUpdated()` fixed it, although I don't really know why exactly. JS is weird

autoscheduler/scraper/management/commands/utils/scraper_utils.py
gannonprudhomme: No, look at the dates above them - I shifted them back a few days so that the scraper would have enough time to scrape them (I think the data was available ~9 days before registration opened)

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
ryan-conn: Oh I didn't notice that, without the parens at the end the arrow function just returns `fetchLastUpdated` instead of calling it. You could have done `setInterval(fetchLastUpdated, 5 * 60 * 1000)` to get the same effect

ryan-conn (approved):
Looks good

PR #465, opened 2020-11-11
Add migrations running on deploy

### LINKED ISSUE #343: Add migration running to GitHub Actions deploy-workflow
We should make it so our deploy-workflow runs `python manage.py migrate` for every single merge into master, so we can be confident that the database is up to date.

Stems from #342, when I discovered that the migrations for the Google OAuth / djangoauth hadn't been applied in GCP yet.

gannonprudhomme:
## Description

This adds docker containers and a workflow job for running migrations on deploy. Note that there really isn't that much new stuff here - I'm basically just combining the GCP authentication steps from App Engine deploy with the Docker containers I made for scraping courses on a schedule. The key difference is that I have to use the credential file and copy it into `docker/cloud-sql-proxy-cred-file/`, then make the `cloud-sql-proxy-cred-file` container correctly include it in its context and then use it for the actual `./cloud_sql_proxy` call.

## How to test

You can see it working [here](https://github.com/aggie-coding-club/Rev-Registration/runs/1383029073) (from commit b853cdd)

## Related tasks

Closes #343


.github/workflows/deploy-workflow.yml
```diff
@@ -95,3 +92,43 @@ jobs:
...
+        uses: actions/checkout@master
+
+      - name: Fill in Postgres credentials
+        run: bash fill_credentials.sh ${{ secrets.POSTGRES_USER }} ${{ secrets.POSTGRES_PASS }}
+
+      # Don't need this
+      - name: Fill .env Google Auth credentials
+        run: bash fill_google_auth_credentials.sh ${{ secrets.GOOGLE_CLIENT_ID }} ${{ secrets.GOOGLE_CLIENT_SECRET }}
+
+      # Don't need this
```
ryan-conn, line 112: What's with these comments? It seems like at least `fill_credentials` should be necessary to make migrations, and if you don't need the step what's preventing you from removing it

ryan-conn (changes requested):
I checked that it works, just a couple things to say on this

.github/workflows/deploy-workflow.yml
gannonprudhomme: lol yeah I totally forgot to remove them

gannonprudhomme:
Dropped the temp deploy commit, should be good to go

ryan-conn (approved):
Looks good now, sorry I took so long to review this 😢

PR #466, opened 2020-11-11
Made height apply correctly on macOS Safari

### LINKED ISSUE #459: Fix site formatting on macOS Safari
Need to give this div:

a height (I think `100%` should work).

gannonprudhomme:
## Description

Makes the full height of the page apply correctly on macOS Safari. Tbh only like 6% of our users use macOS Safari (36.51% use Safari, but only 15% of those are macOS) but still it's unusable without this.

## How to test

Just build it and make sure it doesn't regress anything on y'all's end

## Screenshots

Landing page
|Before|After|
|--|--|
|Screen Shot 2020-11-10 at 8 34 53 PM|Screen Shot 2020-11-10 at 8 35 00 PM|

Schedule page
|Before|After|
|--|--|
|Screen Shot 2020-11-10 at 8 35 17 PM|Screen Shot 2020-11-10 at 8 35 09 PM|

## Related tasks

Closes #459

### DIFF
--- a/autoscheduler/frontend/src/components/App/App.css
+++ b/autoscheduler/frontend/src/components/App/App.css
@@ -8,3 +8,9 @@
 .app-container > div:nth-child(2) {
     flex-grow: 1;
 }
+
+.router {
+    height: 100%;
+    display: flex;
+    flex-direction: column;
+}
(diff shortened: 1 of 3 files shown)


firejake308 (approved):
Still works in Chrome and Edge. Still hopelessly broken in IE, but we have no intent of supporting that anyway

PR #470, opened 2020-11-20
Meeting type is once again hidden if card is small

### LINKED ISSUE #458: Fix too small meeting bug
When meetings are too small the text gets cut off. Pictured is CSCE 482 - 933 for 202111.

firejake308:
## Description

Fixes the bug where the meeting type was being shown regardless of card size.

## Rationale

Turns out the bug was caused by my setting one of the `ScheduleCard` div's to use `display: contents;` in #438. I still think we need that CSS rule, so instead I just adapated the JS that we use to determine the height of all of the text in the card and the height of the card available.

## How to test

Add a small class to your schedule

## Screenshots

On a side note, the color scheme for this is literally awful after seeing how it looks on #463 

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #458 
Makes it easier to implement instructional method icons in the future

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { Typography } from '@material-ui/core';
-
+import * as styles from './MeetingCard.css';
 import Meeting, { MeetingType } from '../../../../types/Meeting';
 import ScheduleCard from '../ScheduleCard/ScheduleCard';
 
@@ -43,17 +43,18 @@ const MeetingCard: React.FC<MeetingCardProps> = ({
       backgroundColor={bgColor}
       borderColor={bgColor}
     >
-      <Typography variant="body2" data-testid="meeting-card-primary-content">
+      <Typography variant="body2" data-testid="meeting-card-primary-content" className={styles.meetingCardText}>
         {`${section.subject} ${section.courseNum}`}
         {isBig
           ? `-${section.sectionNum}`
           : (
             <Typography variant="subtitle2" component="span">
-              {` ${MeetingType[meetingType]}`}
+              &nbsp;
+              {`${MeetingType[meetingType]}`}
             </Typography>
           )}
       </Typography>
-      <Typography variant="subtitle2" hidden={!isBig}>
+      <Typography variant="subtitle2" style={{ display: isBig ? 'block' : 'none' }}>
         {MeetingType[meetingType]}
       </Typography>
     </ScheduleCard>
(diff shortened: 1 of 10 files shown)


firejake308:
Also, relevant:

firejake308:
Since it's more important to show meeting type than section number on the graphic schedule, their locations were swapped. For small sections, section number is now hidden

image

firejake308:
Restored original behavior when cards are big. This complicates the code somewhat, but it looks fine in browser.
image

Also had to fix some tests

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
```diff
@@ -44,7 +44,18 @@ const MeetingCard: React.FC<MeetingCardProps> = ({
...
     >
       <Typography variant="body2" data-testid="meeting-card-primary-content">
-        {`${section.subject} ${section.courseNum}-${section.sectionNum}`}
+        {`${section.subject} ${section.courseNum}`}
+        {isBig
+          ? (
+            <Typography variant="body2" component="span">
+              {`-${section.sectionNum}`}
+            </Typography>
+          )
```
ryan-conn, line 53:
```suggestion
          ? `-${section.sectionNum}`
```
Since this is the same variant as the outer typography, you don't need to make another component

ryan-conn (changes requested):
Mostly looks good, one thing I want to point out is that if the resolution is really low it will still overflow the bold text onto a second line like in the picture below:
image
Since as far as I know this only happens when the resolution is incredibly low (it's 1100x500 in the picture I sent), I don't think it's super important to fix but we might as well since it shouldn't be too difficult, what do you think @gannonprudhomme ?

firejake308:
Okay, so I hid the overflowing text using ellipses. Adding CSS was easier than adding more custom JS. It still looks the same on large screens:
image

But on very small ones, it looks more like this:
image

And if you go into the quantum realm (906 x 406), you might see this:
image

firejake308:
Fixed tests

gannonprudhomme (changes requested):
The functionality on larger screens has regressed slightly. I also think it's choosing the smaller meeting for the M/W sections when it really shouldn't shrink (on 1080p), as it seems it has enough space for 2 lines. 

|This branch|Expected (on deploy rn)|
|--|--|
|chrome_GgFYPudROP|chrome_101Dycvb39|

firejake308:
Fixed vertical centering by removing `height: 100;`, and `justify-content: center;` on the parent div takes care of the rest. Also noticed a bug where the first line of text was getting cut off.  (see blue arrow in pic below). This was fixed by adding `flex-shrink: 0;`

|Before|After|
|--|--|
|image|image|

gannonprudhomme (approved):
All previous issues are fixed, so this is good to go 🎉

PR #471, opened 2020-11-20
Transition and memoize course cards

### LINKED ISSUE #1: Setup Django basics

firejake308:
Just using this as a space to keep track of remaining bugs:
- CSCE 314 has extra whitespace. This is probably due to it having an intermediate size
- There are 2 transitions currently, one from like 30 to 50px and a second one from 50px to 500px.
  - could just eliminate the course description
  - but more importantly, I need to change the timing of when the CSS transition rule is added
- Initial section load still feels bad (seen on PHYS 202), future task imo to add loading indicator
- bug where card becomes tiny if opened at the same time that another one is closed
  - seems to affect some courses but not others
- display nine is being set early

# TODO
----
## Description

Describe what the pull request is achieving.

## Rationale

If necessary, defend any changes that you made.

## How to test

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|

## Related tasks

Closes #1


firejake308:
Replaced by #474

PR #472, opened 2020-11-20
Term-select: Re-arrange redux

gannonprudhomme:
## Description

This re arranges the Redux so that course cards, schedules, and availabilities have instant access to the current term.

I'll be following this up with a PR based off of this one that'll address the session desync that we were facing previously

## Rationale

To fix the dependency cycle I put all of the appropriate action type interfaces in `redux/actions/termData.ts`, which I really don't think is the right place/name for it.

Also there's a few linting warnings, but I'll fix them in the term-select PR at a later point just b/c I don't wanna update this with term-select.

## How to test

Tbh I think we just have to rely on our tests / linting for this one.

### DIFF
--- a/autoscheduler/frontend/src/redux/actions/termData.ts
+++ b/autoscheduler/frontend/src/redux/actions/termData.ts
@@ -1,0 +1,108 @@
+/* These all have to be in the same file due to dependency cycles
+*/
+
+import Availability, { AvailabilityArgs } from '../../types/Availability';
+import { CourseCardOptions } from '../../types/CourseCardOptions';
+import Meeting from '../../types/Meeting';
+import Schedule from '../../types/Schedule';
+import { RemoveSelectedAvailabilityAction } from '../reducers/selectedAvailability';
+
+/*
+  TERM
+*/
+// action type interface
+export interface SetTermAction {
+  type: 'SET_TERM';
+  term: string;
+}
+
+/*
+  AVAILABILITIES
+*/
+// action type interfaces
+export interface AddAvailabilityAction {
+    type: 'ADD_AVAILABILITY';
+    availability: AvailabilityArgs;
+}
+export interface DeleteAvailabilityAction {
+    type: 'DELETE_AVAILABILITY';
+    availability: AvailabilityArgs;
+}
+export interface UpdateAvailabilityAction {
+    type: 'UPDATE_AVAILABILITY';
+    availability: AvailabilityArgs;
+}
+export interface MergeAvailabilityAction {
+    type: 'MERGE_AVAILABILITY';
+    numNewAvs: number;
+}
+export interface SetAvailabilitiesAction {
+  type: 'SET_AVAILABILITIES';
+  availabilities: Availability[];
+}
+export type AvailabilityAction =
+    AddAvailabilityAction | DeleteAvailabilityAction | UpdateAvailabilityAction |
+    MergeAvailabilityAction | SetAvailabilitiesAction | RemoveSelectedAvailabilityAction;
+
+/*
+    SCHEDULES
+*/
+// action type interfaces
+export interface AddScheduleAction {
+    type: 'ADD_SCHEDULE';
+    meetings: Meeting[];
+}
+export interface RemoveScheduleAction {
+    type: 'REMOVE_SCHEDULE';
+    index: number;
+}
+export interface ReplaceSchedulesAction {
+    type: 'REPLACE_SCHEDULES';
+    schedules: Meeting[][];
+}
+export interface SaveScheduleAction {
+    type: 'SAVE_SCHEDULE';
+    index: number;
+}
+export interface UnsaveScheduleAction {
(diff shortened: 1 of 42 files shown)


autoscheduler/frontend/src/redux/actions/termData.ts
```diff
@@ -0,0 +1,108 @@
...
+import Availability, { AvailabilityArgs } from '../../types/Availability';
+import { CourseCardOptions } from '../../types/CourseCardOptions';
+import Meeting from '../../types/Meeting';
+import Schedule from '../../types/Schedule';
+import { RemoveSelectedAvailabilityAction } from '../reducers/selectedAvailability';
+
+/*
+  TERM
+*/
+// action type interface
```
gannonprudhomme, line 13: Going to remove these shortly

autoscheduler/frontend/src/redux/actions/termData.ts
```diff
@@ -0,0 +1,108 @@
+/* These all have to be in the same file due to dependency cycles
```
gannonprudhomme, line 1: Any suggestions for the location + name for this? I don't think `actions/termData` is really accurate since it doesn't actually contain actions. Maybe `reducers/termDataTypes`?

gannonprudhomme:
Updated #349 with master, so rebased this with #349 as a result

autoscheduler/frontend/src/redux/actions/termData.ts
firejake308: That makes more sense, since the other action type `interface`s are all in the `reducers` folder

autoscheduler/frontend/src/redux/reducer.ts
```diff
@@ -5,21 +5,16 @@ import { combineReducers } from 'redux';
 import availability from './reducers/availability';
 import availabilityMode from './reducers/availabilityMode';
 import meetings from './reducers/meetings';
-import schedules from './reducers/schedules';
-import courseCards from './reducers/courseCards';
 import selectedAvailabilities from './reducers/selectedAvailability';
 import selectedSchedule from './reducers/selectedSchedule';
-import term from './reducers/term';
+import termData from './reducers/termData';
```
firejake308, line 10: There's an unused import on line 5, since availbility is inside term data now

firejake308 (changes requested):
Pretty much all of these changes are just adding `.termData` after every `getState()` in our codebase. The only things that might need changing are the organization of all of the action type interfaces and a linter warning about an unused variable.

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -465,60 +442,167 @@ describe('Course Cards Redux', () => {
...
+            customizationLevel: CustomizationLevel.BASIC,
+            collapsed: false,
+          },
+        ];
+
+        // act
+        store.dispatch<any>(replaceCourseCards(wrongCourseCards, '201931'));
+        store.dispatch<any>(replaceCourseCards(rightCourseCards, '202031'));
+
+        // assert
```
ryan-conn, line 512: This test doesn't check the number of course cards like the others you added

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -465,60 +442,167 @@ describe('Course Cards Redux', () => {
...
+          {
+            course: 'POLS 207',
+            customizationLevel: CustomizationLevel.BASIC,
+            collapsed: false,
+          },
+        ];
+
+        // act
+        store.dispatch<any>(replaceCourseCards(wrongCourseCards, '201931'));
+        store.dispatch<any>(replaceCourseCards(rightCourseCards, '202031'));
```
ryan-conn, line 510: Since these tests are supposed to make sure the incorrect term isn't being applied, shouldn't it first replace with the cards for the correct term, then the incorrect one? I don't see how the current implementation tests whether the incorrect term is ignored

ryan-conn (changes requested):
A couple of nitpicks on the tests, I'd also appreciate if you could fix the remaining linter issues before this gets merged

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
gannonprudhomme: The order shouldn't matter - I set the term to `202031` at the start, so regardless of the order it should ignore the `201931` term course cards and use the `202031` cards. I can change it though

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
gannonprudhomme: Yeah that's true - but since both `201931` and `202031` have 2 course cards, the assertion that there are 2 cards would be pointless (since it would pass if the `201931` cards were used).

ryan-conn (approved):
Looks good

PR #473, opened 2020-11-22
Select by instructor

### LINKED ISSUE #259: Make instructors selectable in SectionSelect
You should be able to select a professor in the SectionSelect's `SECTION` customization level, which would select/deselect all of the sections for that professor. 

(insert design here)

firejake308:
## Description

A common request that we have received is the ability to select all sections taught by a particular professor, as this is a common way for students to pick classes. This PR adds such a checkbox next to the instructor name. From a design perspective, this makes the instructor header a little cramped on small screen sizes, but it can't really be helped imo 🤷. Maybe consider showing professor's last name only?

Anyway, in order to provide a sense of hierarchy, the `SectionInfo`s for the individual sections had to be indented to the right of the professor checkbox. Because of this indentation, on smaller screens, the meeting info may be wrapped onto 2 lines. If you don't like the current wrapping, we might need to look into CSS Grid rather than Flexbox (the current approach), which honestly scares me just a little bit.

Also, I elected to use alternating colors (light maroon & white) to distinguish between mulitple meetings in the same section. Because I didn't like how this looked when there was only one meeting, I added a CSS rule to override the alternating colors for single-meeting sections. ~Someone remind me to add~ Added a comment to that line, because I'm 90% sure we're going to forget what it does in the future.

We had previously discussed adding icons, but after trying them out, they felt like they cluttered the space to me. Since alternating colors was pretty effective at distinguishing between one meeting and the next, I honestly don't feel that icons are necessary.

Small design bug: If a section has an odd number of meetings (e.g. ENGR 102), the mini-divider will be covered by the background of the last meeting. First fix that I thought of was throwing in a 3px margin between the last section and the divider, and I think I can use CSS selectors to limit this to the cases when the last meeting is colored. Lmk if y'all have better ideas.

## How to test

Automated tests added for the selecting functionality. For visuals, check out KINE 199 for 1-meeting sections, BIOL 351 for 2-meeting sections, ENGR 102 for 3-meeting sections, PHYS 202 for 4-meeting sections, and any other classes you like.

## Screenshots

|1 meeting|2 meetings|3 meetings|
|--|--|--|
|image|image|image|

## Related tasks

Closes #259


eelyort (commented):
Looks good to me, its especially nice in that it seems like it selects only one group regardless of sorting order.

For addressing the instructor label being cramped, what do you think about dispensing with the checkbox and instead changing the color of the label or something?

ryan-conn (commented):
While we said to make this use grid in a prior meeting, looking at this again I think it looks fine with 2 columns since they're aligned. It looks like this hasn't been changed since then so we might want to discuss what exactly we want to do in the next meeting

While the 2 column layout likes fine now imo, it looks a bit odd with one especially as the width goes up, here's 1440p:
image
The two things I noticed are
- Widths aren't uniform since they depend on content widths
- Building and meeting days are really close together which looks odd imo. Off the top of my head I can't think of a simple solution to this since that's the behavior we want when there are two columns, but I still wanted to point it out

firejake308:
Here are the options for spacing out each of the cells if we stick with flexbox:
|Space between (current)|Space around|Space evenly|
|--|--|--|
|image|image|image|

Personally, I think space evenly is best when there is no wrapping, but I want to double-check with y'all since it changes the ones where it wraps too

firejake308:
Another screenshot that shows a class with a few more meetings (MATH 151):
image

gannonprudhomme:
Just btw you need to merge master into this as you have some conflicts

gannonprudhomme:
Reminder to update this with master!

firejake308:
The last merge was non-trivial, as it involved reconciling my "select all by instructor" with @eelyort 's "select all" changes in tests and also some conflicts with @ryan-conn 's instructional method icons. I did my best to make sure that both of these features still work, and tests are passing, which is a good sign, but please double-check to make sure that I haven't broken anything. T's and G'd

autoscheduler/frontend/src/redux/actions/courseCards.ts
```diff
@@ -272,6 +272,21 @@ ThunkAction<void, RootState, undefined, UpdateCourseAction> {
   };
 }
 
+export function setSelected(courseCardId: number, secIdx: number, value: boolean):
```
gannonprudhomme, line 275: Function docstring :P

gannonprudhomme (changes requested):
A couple things: 
- Instructor names should be selectable imo - can discuss this is in the meeting if you disagree tho (just kind of weird when you compare it to the select all)
- Sometimes columns aren't aligned when you'd expect them to be (ENGR 102-550 for 202111). If this can't be fixed it's really not a big deal given how much time you've seemingly put into the flex/grid for meetings
- Instructional methods are incorrectly on a new line instead of the same line as the section number / seats remaining

Screenshot containing the latter 2 issues:
chrome_OEzcItcUvi

Code wise this looks good though

gannonprudhomme:
Merge with master please :)

firejake308:
The text is *slightly* stretched out, but other than that, I think I undid all of the style changes that came with turning the professor's name and the checkbox into a `Button`
|Before|After|
|--|--|
|image|image|
|image|image|

firejake308:
I think I'll probably do a separate PR to get the transition between basic and section select

gannonprudhomme (dismissed):
- Instructional methods are incorrectly on a new line instead of the same line as the section number / seats remaining
- Meeting times don't touch the right side of the course card

|This PR|Remote|
|--|--|
|chrome_q99tZ2KWFb|chrome_asj4pHLJ3f|

firejake308:
As you can see below, I have fixed the issue with the instructional method icons. However, as for the alignment, it may or may not be possible to fix that. As discussed earlier in this thread, with the flex layout, pushing the meeting time to the right edge would require using `justify-content: space-between;`, which would also push the meeting day to the left. This could be uncomfortably close to the building name, but I could try it out if you want to see it.

That said, there is another issue that I noticed in the image below, specifically that if the card becomes so small that the meeting time is forced to wrap, then it aligns to the left, which is weird. I added another commit to center-align the text, but this only has an effect when the meeting time is wrapped. The reason for this is that the div with the text only takes up as much space as the text itself, so there is no alignment to speak of unless the div is stretched as wide as possible to fit the wrapped text.

|Before center align|After center align|
|--|--|
|image|image|

gannonprudhomme:
Btw backend linting will pass if you update with master

firejake308:
Applied changes discussed at 2/28 meeting, namely
- Right-align meeting times
- Move 1fr from meeting days to meeting time
- Increase cutoff to 1500px

Here are 2 screenshots of the 2 possible layouts
|2 rows|1 row|
|--|--|
|image|image|

gannonprudhomme:
This is failing frontend tests / build just btw

firejake308:
Merged

gannonprudhomme (changes requested):
Super close to being ready! Just two minor things.

Getting an error in the console upon render: `validateDOMNesting(...): <li> cannot appear as a descendant of <li>`

Also the meeting days are barely not aligned to how they were previously - I _might_ be fine leaving it like this, but really wouldn't prefer it tbh

|Before(1080p)|After(1080p)|
|--|--|
|chrome_h8Bt9jEYAl|chrome_X0jGdRP8hy|

Even worse on 1440p

chrome_BcONIg2nQE

firejake308:
Centered text and fixed the DOM nesting by changing the GPA text in GradeDist by telling `ListSubheader` to render as a `span` rather than an `li`. This shouldn't have affected any of the styles from what I can tell.

|Left aligned|Centered|
|--|--|
|image|image|

gannonprudhomme (dismissed):
From the meeting:
  - Meeting days spacing is off, isn't evenly spaced (see previous screenshots)

We need to revert the meeting days centering, as it not being left-aligned makes the low-width setting hard to parse. We could make it conditional (center aligned when high-width, left-aligned with low-width), but honestly I think left-aligned in both cases is fine personally
chrome_h6vvukKsYf

Also, I think the instructor text should be either center aligned or bottom aligned with the checkbox. It appears like its top-aligned currently. 
chrome_3gYt8KQOVe

firejake308:
To discuss at meeting: 
image

firejake308:
Prof looks centered to me:
image

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.css
```diff
@@ -71,10 +75,68 @@
...
+    }
+    .meeting-info-wrapper > div:nth-child(2) {
+        text-align: right; /* applies to building in 2-column view */
+    }
+    .meeting-info-wrapper > div:nth-child(3) {
+        text-align: left; /* applies to days in 2-column view */
+    }
+}
+
+.meeting-info-wrapper:nth-child(2n) div {
```
ryan-conn, line 106:
```suggestion
.meeting-info-wrapper:nth-child(even) div {
```
I don't actually care if you make this change since this is perfectly readable but just so you know, there are `nth-child(even)` and `nth-child(odd)` selectors

autoscheduler/frontend/src/.eslintrc.js
```diff
@@ -63,5 +63,6 @@ module.exports = {
     "@typescript-eslint/semi": ["error"],
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
+    "linebreak-style": 0,
```
gannonprudhomme, line 66:
```suggestion
```

gannonprudhomme (approved):

PR #474, opened 2020-11-22
Transition and memoize course cards

### LINKED ISSUE #387: useMemo for section select when expanding/collapsing, and switching between basic/section select
Currently if you open a course card with a lot of sections, it will have to rerender them all whenever you expand the card, or when you switch from basic to section select. We should be able to `useMemo` to memoize the contents of the expanded course card/section select, and avoid a rerender if the sections haven't changed.

### LINKED ISSUE #415: Add collapse/expand transitions to course cards
Especially with the additions of #377, we should add collapse/expand transitions to make our website feel better. Should be able to use the [Transitions](https://material-ui.com/components/transitions/#transitions) provide by material-ui. Bonus points if you animate the collapse/expand icons like Google does it:

We'll also probably need to make sure #387 is done before this just so it doesn't feel laggy.

firejake308:
## Description

Users have commented that it feels jerky for cards to suddenly collapse and expand. Using Material-UI's `Collapse` component, I have added a 300ms transition between these states. I think `Collapse` probably uses some secret wizardry where they keep a pre-rendered copy of the collapsed card on hand so that they can rapidly add it back to the DOM during expansion.

Although `Collapse` makes it fast to expand and collapse the cards, it has no effect on switching between Basic and Section modes. I tried applying `useMemo` to `sectionSelect` to speed up this second transition, but to no avail. It appears that the time-consuming portion is not actually running the React code to determine what the `SectionSelect` should be, but to take that result and render it to the screen. (This was fairly surprising because of how inefficiently the `SectionSelect` component is written.) To eliminate the expensive step of rendering HTML elements to the DOM, I switched to using `display: none;` to hide the `customizationContent` that wasn't currently being used.

## Rationale

There is one sort of sketchy area where I manually calculate the height of the card (the `fixHeight` function), and this is liable to be accidentally broken in the future. Actually, it'll break as soon as Troy merges his Select All button. I'm not sure how I could add tests for it, though, so lmk if you have ideas.

Also, the part with the `MutationObserver` seems pretty hacky, too. The problem was that when the initially expanded card is rendered, it only has the loading circle, which is small enough to apply `expandedRowSmall`. So if I try to set the height of `sectionRows` while the `overflow: visible;` from `expandedRowSmall` is still active, it'll tell me I have all the space in the world. Instead, I waited until `CourseSelectColumn` changed the class to `expandedRow`, and then applied `fixHeight`. Waiting for the class to change meantt using a mutation observer. unless y'all have any better ideas.

## How to test

I'd like to try to get some automated tests for this, but JS DOM doesn't mock height and CSS rules IIRC, so I can't think of a way to do it.

For manual tests, open small classes like JAPN 201, medium classes like CSCE 314, and enormous classes like KINE 199. They should all collapse and expand smoothly no matter what order you open them in and regardless of how many times you switch between Basic and Section in between.

## Screenshots

|Before|After|
|--|--|
|old|transitions|

## Related tasks

Closes #415 
Closes #387


gannonprudhomme:
This is failing 20 tests btw lol

firejake308:
Ready for review

ryan-conn (changes requested):
Most of my comments have to do with readability since this component is really important and it's getting pretty hard to read, but I also think I found a fix for the scroll bar issue

Also a bug I noticed is that if you increase the height, the expanded card will no longer take up the full height, instead staying at what it was at previously even after collapsing/expanding the card again:
expand bu

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -34,6 +34,10 @@
     padding-top: 8px;
     padding-right: 8px;
     box-sizing: border-box;
+    min-height: auto;
```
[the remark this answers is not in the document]
firejake308: I actually don't see any scroll bar on my computer, on either Chrome or Firefox. Can you confirm that the scroll bar still shows up on yours?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
```diff
@@ -1,32 +1,223 @@
...
-      onExpand={toggleCollapsed}
-      id={id}
-    />
-  ) : (
-    <ExpandedCourseCard
-      onCollapse={toggleCollapsed}
-      id={id}
-    />
+  const contentRef = React.useRef<HTMLDivElement>(null);
+  const fixHeight = React.useCallback((): void => {
```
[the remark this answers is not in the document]
firejake308: Added some comments. Does it make sense now?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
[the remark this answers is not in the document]
ryan-conn: This is fixed with your most recent changes

ryan-conn:
I looked through this PR again and didn't see anything that should prevent it from getting merged.

Once you get the changes from master merged in here I should be able to approve this unless something breaks

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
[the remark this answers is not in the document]
firejake308: Checked this on my own screen again, and using `38px` doesn't work on my screen if the course title is really long (see below). Do you think we could just remove the course title? I think that if we only showed the department and course number, it would always fit on 1 line no matter the screen size.

image

firejake308:
Also, bug:
image

firejake308:
I think the last 3 commits address the changes we discussed at the last couple of meetings

gannonprudhomme (changes requested):
use-memo-bug

The collapse/expand icon bug is still occurring - other bugs were fixed tho

gannonprudhomme:
Merge with master! Otherwise I think this is good

firejake308:
Merged with master. The transition between basic and section select is going to be trickier than I first thought because we're currently using Material UI's `Collapse` element to manage the transition for us, which only has 2 possible states: collapsed and expanded, and transitions only occur when that state changes. Since Basic Select and Section Select are both included in the expanded state, there is no transition between them. If it is doable, it will need to be a separate PR.

gannonprudhomme (dismissed):
Two bugs, although really the first one is the only one that's a bug:

On refresh/load, the last minimized course card in the column will collapse for some reason:
transition-memo-bug-1

Next, the expand animation is surprisingly choppy (recording of me going frame-by-frame in a recording of it expanding):

transition-memo-bug-2-abrupt

The latter one might not be fixable and is more of a nitpick, but just wanted to mention it

firejake308:
@gannonprudhomme I figured out what's causing the bug that you pointed out with the last card's unnecessary collapse animation:

When the page is first created, the CourseSelectColumn has a single, empty card that is expanded by default. On the user's first visit, this works nicely because now the card is big and they can see where to enter the course. However, if the user already has saved courses, then the collapsed state of the saved courses overwrites the expanded state of the formerly empty 0th card. Thus, the 0th card has gone from expanded to collpased, so the collapse animation plays.

The ideal behavior would be for the 0th card to start out collapsed if the user had saved courses and expanded if the user did not have any saved courses. However, it is not possible for us to know that until we make the `/sessions/get_saved_courses` API call, so we need to decide on the best initial state to use until then. If we change the 0th card to be collapsed by default, we run the risk of confusing first-time users if `/sessions/get_saved_courses` takes too long to return. However, if we change the 0th card to be expanded by default, we have to play an extra animation to collapse the 0th card and expand whatever card the user was looking at when they left.

What's y'all's opinion? I'm thinking the initial API call shouldn't take too long, so making the blank card collpased until it returns shouldn't be too much of an issue. However, I'd like y'all's input if y'all think leaving it as is would be better or if there's a better approach that I haven't thought of.

firejake308:
@gannonprudhomme I updated an outdated CSS rule that seems to improve transition smoothness on my computer. Can you confirm that it looks better for you too?

gannonprudhomme:
Yep that seems to have fixed it! Heck yeah

gannonprudhomme:
Need to merge master into here & fix the conflicts

gannonprudhomme (changes requested):
Can confirm that the previous loading-in bug was fixed, but now it does this weird animation when you add a card

use-memo-card-3-bug

firejake308:
Fixed the bug that you pointed out. However, there is still a bug when you have a large course card open and then add a new card. In that case, the card's contents are rendered like 8px above where they should be.

firejake308:
Ironed out *most* of the bugs. The opening animation does feel a little fast on my computer, and it does a noticeable adjsutment when it realizes that it needs a scroll bar, but I think these are acceptable caveats. Unless y'all have any idea what might be making the opening transition too fast?

firejake308:
Guys guys I think it finally works!

firejake308:
@gannonprudhomme Fixed the speed by adding a `React.memo`

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/ExpandedCourseCard.css
```diff
@@ -36,15 +37,25 @@
...
+.course {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    flex-grow: 1;
 }
 
 .content {
-    margin-top: 8px;
+    /* margin-top: 8px; */
```
gannonprudhomme, line 52:
```suggestion
```
Reminder to remove this (assuming you don't want it)

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -22,12 +25,34 @@ const CourseSelectColumn: React.FC = () => {
...
+      const header = cardEl.getElementsByClassName(cardStyles.header)[0] as HTMLElement;
+      const content = cardEl.getElementsByClassName(cardStyles.content)[0] as HTMLElement;
+      const sectionRows = cardEl.getElementsByClassName(sectionStyles.sectionRows);
+      // the actual height of content should include the fully expanded section rows, if present
+      const deltaRowHeight = sectionRows.length > 0
+        ? sectionRows[0].scrollHeight - sectionRows[0].clientHeight
+        : 0;
+      const expandedRowHeight = header.scrollHeight + content.scrollHeight
+        + parseFloat(getComputedStyle(content).marginTop) + deltaRowHeight;
+
```
gannonprudhomme, line 55: Do you think we could turn this into a function? It looks really funky and I think adding a function with a good name + description providing context would help a lot

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -37,16 +62,46 @@ const CourseSelectColumn: React.FC = () => {
...
+            // height of a collapsed row
+              otherKidsHeight += 38;
+            } else {
+            // height of the button at the top
+              otherKidsHeight += col.children[i].clientHeight;
+            }
+          }
+          // 500 is max height for row, 217 is height of rest of expanded card other than
+          // sectionRows
+          const availableHeight = Math.max(500, col.clientHeight - otherKidsHeight) - 217;
```
gannonprudhomme, line 86: We should make these `500` and `217` constants

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -22,12 +25,34 @@ const CourseSelectColumn: React.FC = () => {
   );
   const term = useSelector<RootState, string>((state) => state.termData.term);
   const dispatch = useDispatch();
+  const [loading, setLoading] = React.useState(true);
+  const [wasCourseRemoved, setCourseRemoved] = React.useState(false);
+
+  const removeCallback = React.useCallback((id: number) => {
```
gannonprudhomme, line 31:
```suggestion
  const removeCourseCallback = React.useCallback((id: number) => {
```
Think adding `course` here helps identify what this actually does at a glance

Also, what is the need for this exactly? How come the course cards can't just remove themselves?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
```diff
@@ -1,35 +1,202 @@
...
+        </Button>
+      </ButtonGroup>
+      {getCustomizationContent()}
+    </div>
+  );
+
+  return (
+    <Card className={styles.card}>
+      {header}
+      <Collapse in={!collapsed} appear enter={shouldAnimate} onEntered={resetAnimCb}>
```
gannonprudhomme, line 195: what does `enter` do? I can't find any mention of it in the docs

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -22,12 +25,34 @@ const CourseSelectColumn: React.FC = () => {
...
+    dispatch(removeCourseCard(id));
+    setCourseRemoved(true);
+  }, [dispatch, setCourseRemoved]);
+
+  const resetAnimations = React.useCallback(() => {
+    setCourseRemoved(false);
+  }, []);
```
gannonprudhomme, line 38: How come this doesn't have `setCourseRemoved` in it's deps, but `removeCallback` does? Does it make a difference either way?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -37,16 +62,46 @@ const CourseSelectColumn: React.FC = () => {
...
+            if (col.children[i] === expandedRowRef.current) {
+            // ignore own height
+            // eslint-disable-next-line no-continue
+              continue;
+            } else if (col.children[i].classList.contains(styles.row)) {
+            // height of a collapsed row
+              otherKidsHeight += 38;
+            } else {
+            // height of the button at the top
+              otherKidsHeight += col.children[i].clientHeight;
```
gannonprudhomme, line 81:
```suggestion
              // height of the button at the top
              otherKidsHeight += col.children[i].clientHeight;
```

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -37,16 +62,46 @@ const CourseSelectColumn: React.FC = () => {
...
+        if (col) {
+          let otherKidsHeight = 0;
+          for (let i = 0; i < col.children.length; i++) {
+            if (col.children[i] === expandedRowRef.current) {
+            // ignore own height
+            // eslint-disable-next-line no-continue
+              continue;
+            } else if (col.children[i].classList.contains(styles.row)) {
+            // height of a collapsed row
+              otherKidsHeight += 38;
```
gannonprudhomme, line 78:
```suggestion
              // height of a collapsed row
              otherKidsHeight += 38;
```

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -37,16 +62,46 @@ const CourseSelectColumn: React.FC = () => {
...
+      // adjust height of section rows
+      if (sectionRows.length > 0) {
+        const col = document.getElementById('course-select-container');
+        if (col) {
+          let otherKidsHeight = 0;
+          for (let i = 0; i < col.children.length; i++) {
+            if (col.children[i] === expandedRowRef.current) {
+            // ignore own height
+            // eslint-disable-next-line no-continue
+              continue;
```
gannonprudhomme, line 75:
```suggestion
              // ignore own height
              // eslint-disable-next-line no-continue
              continue;
```
Aligning comments with block

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css.d.ts
```diff
@@ -11,8 +11,12 @@ declare namespace CourseSelectColumnCssNamespace {
     courseSelectColumn: string;
     "expanded-row": string;
     "expanded-row-small": string;
+    "expanded-row-temp": string;
     expandedRow: string;
     expandedRowSmall: string;
+    expandedRowTemp: string;
```
gannonprudhomme, line 17: is `Temp` in `expandedRowTemp` intentional? Or was it like a testing name

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
```diff
@@ -1,35 +1,202 @@
...
+import * as styles from './ExpandedCourseCard/ExpandedCourseCard.css';
+import SectionSelect from './ExpandedCourseCard/SectionSelect/SectionSelect';
+import BasicSelect from './ExpandedCourseCard/BasicSelect/BasicSelect';
 
 interface CourseSelectCardProps {
   id: number;
+  collapsed: boolean;
+  shouldAnimate?: boolean;
+  removeCourseCard?: (index: number) => void;
+  resetAnimCb?: () => void;
```
gannonprudhomme, line 23: Can you add comments explaining what these are + what they're used for? (e.g. `collapsed` for memoization)

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -18,9 +18,10 @@ import SmallFastProgress from '../../../../../SmallFastProgress';
 
 interface SectionSelectProps {
   id: number;
+  onMounted?: VoidFunction;
```
gannonprudhomme, line 21: Is this ever actually assigned? I don't see any creation of `SectionSelect` that sets it

gannonprudhomme (changes requested):
Performance is much improved from the previously with the addition of `React.memo`. Other bugs are still occurring (still only care about # 1), so if you don't want to invest the time into it rn I'm completely fine with merging this and making it an issue to fix later

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
firejake308: Good point, I ended up removing this prop when I moved the `fixHeight` functionality up from `CourseSelectCard` to `CourseSelectColumn`

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css.d.ts
firejake308: Intentional, because the class is only applied for a short time before it is replaced by either `expandedRow` or `expandedRowSmall`

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
firejake308: I guess I should add it in. It shouldn't make a difference, since I'm pretty sure the identity of the `setCourseRemoved` function stays the same from one render to the next.

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
firejake308: Material UI internally uses React Transition Group, and the `enter` prop gets forwarded to it. See [docs](https://reactcommunity.org/react-transition-group/transition)

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
gannonprudhomme: Honestly really the only reason I mentioned it was since I was confused how exhaustive deps didn't catch it, especially since it's literally just `useMemo` under the hood

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
firejake308: When a card is removed, it used to dispatch a `removeCourseCard` action themselves. However, now, when a card is removed, it needs to notify the other cards not to animate since we said that we didn't like that animation. This is done by passing the information up to `CourseSelectColumn`, where `wasCourseRemoved` can be set to `true`

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -22,31 +25,98 @@ const CourseSelectColumn: React.FC = () => {
...
 
   const expandedRowRef = React.useRef<HTMLDivElement>(null);
-  // Use dynamic className to style expanded card
-  React.useLayoutEffect(() => {
+  const MIN_CARD_HEIGHT = 500;
+  // height of fixed card contents, that is, everything except section rows
+  const CARD_CONTENT_BASE_HEIGHT = 217;
+  const COLLAPSED_ROW_HEIGHT = 38;
+  // comes from padding-top in the .row class
+  const ROW_PADDING_TOP = 89;
```
gannonprudhomme, line 46:
```suggestion
  const ROW_PADDING_TOP = 8;
```
typo

gannonprudhomme:
Also update this with master real quick to fix the failing backend linting

gannonprudhomme:
Waiting for Ryan to do a quick sanity check then this will be good to go 🥳

ryan-conn (approved):
I found one bug in this, which is that whenever you close a card, the next time you expand one it won't play the animation. This happens on the next time a card is expanded whether there was one expanded when you closed a card or not.

That's pretty minor though so we can wait to fix it until after this is merged, just thought it was worth pointing out

gannonprudhomme (approved):

firejake308:
> I found one bug in this, which is that whenever you close a card, the next time you expand one it won't play the animation. This happens on the next time a card is expanded whether there was one expanded when you closed a card or not.

I swear this PR is the hydra of bugs. Every time I squash one bug, two more appear in its place. That is definitely something that needs fixing, but in the interest of launch, I'll make an issue for it and fix it later

PR #475, opened 2020-11-23
Cache terms api

### LINKED ISSUE #386: Cache /api/terms calls
Since we now have 2 components that need to fetch `/api/terms`, caching the results will let us avoid calling the route as much (it will also prevent it from being called whenever you go back to the homepage as a cool bonus)

Basically we'll need an async `getTerms` function that will check if it's been called before, and if not fetch `/api/terms` and put the data it gets into the cache. Then if it's called again, it will block until the first call has completed fetching (so that if it's called twice at the same time, it doesn't end up just fetching `/api/terms` twice), and once it has, return the terms from the first call.

ryan-conn:
## Description

Caches the result of `api/terms`, meaning it will only ever be called once

## Rationale

I used a self-executing function to store state instead of doing it in global scope to make the caching self-contained

Also the test doesn't technically assert that `api/terms` is only called once for a couple reasons:
- `fetchMock` isn't actually adding the api call to the results of `fetchMock.mock.calls` for some reason
- The test will fail if it's called twice anyways, since the component won't know how to handle the response

## How to test

What I was doing instead of checking out the navbar term select branch is just add multiple `<TermSelect />` components to the main page, you'll notice the call only executes once

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|
| image | | [terms fetching](https://user-images.githubusercontent.com/28655462/100021415-0e70e480-2da7-11eb-84ea-881851b56283.gif) |

## Related tasks
Closes #386

### DIFF
--- a/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
+++ b/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
@@ -8,6 +8,23 @@ import { useDispatch } from 'react-redux';
 import setTerm from '../../../redux/actions/term';
 import * as styles from './SelectTerm.css';
 
+/**
+ * Stateful function that caches the json for an api/terms call, so that the API route
+ * only has to be fetched once.
+ */
+const getTermsJson = ((): () => Promise<any> => {
+  let fetched = false;
+  let fetchedData: Promise<any>;
+
+  return async (): Promise<any> => {
+    if (fetched) return fetchedData;
+
+    fetchedData = fetch('api/terms').then((res) => res.json());
+    fetched = true;
+    return fetchedData;
+  };
+})();
+
 const SelectTerm: React.FC = () => {
   // anchorEl tells the popover menu where to center itself. Null means the menu is hidden
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
@@ -17,9 +34,9 @@ const SelectTerm: React.FC = () => {
 
   // Fetch all terms to use as ListItem options
   function getTerms(): void {
-    fetch('api/terms').then((res) => res.json()).then(
+    getTermsJson().then(
       (res) => {
-        const termsMap = new Map(Object.entries(res));
+        const termsMap = new Map<string, string>(Object.entries(res));
         setTermMap(termsMap);
         setOptions(Array.from(termsMap.keys()));
       },
--- a/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
+++ b/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
@@ -13,14 +13,12 @@ import * as styles from './SelectTerm.css';
  * only has to be fetched once.
  */
 const getTermsJson = ((): () => Promise<any> => {
-  let fetched = false;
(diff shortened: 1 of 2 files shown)


autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
```diff
@@ -8,6 +8,23 @@ import { useDispatch } from 'react-redux';
...
+/**
+ * Stateful function that caches the json for an api/terms call, so that the API route
+ * only has to be fetched once.
+ */
+const getTermsJson = ((): () => Promise<any> => {
+  let fetched = false;
+  let fetchedData: Promise<any>;
+
+  return async (): Promise<any> => {
+    if (fetched) return fetchedData;
```
gannonprudhomme, line 20: Could you remove `fetched` altogether and return `fetchedData` if its truthy?

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
```diff
@@ -8,6 +8,23 @@ import { useDispatch } from 'react-redux';
 import setTerm from '../../../redux/actions/term';
 import * as styles from './SelectTerm.css';
 
+/**
+ * Stateful function that caches the json for an api/terms call, so that the API route
+ * only has to be fetched once.
+ */
+const getTermsJson = ((): () => Promise<any> => {
+  let fetched = false;
+  let fetchedData: Promise<any>;
```
firejake308, line 17: For personal learning:

So the basic idea is that we're storing the `fetchedData` inside this closure so that we don't need to fetch it again, right? Also, if we were obtaining this data synchronously (not through fetch, maybe like reading from cookies or something), we wouldn't necessarily need to return a Promise, would we?

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
firejake308: Agreed, the `fetched` variable seems kinda redundant

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
ryan-conn: > So the basic idea is that we're storing the `fetchedData` inside this closure so that we don't need to fetch it again, right?
Yeah, it will reuse that same promise every time it's called instead of making a new request
> Also, if we were obtaining this data synchronously (not through fetch, maybe like reading from cookies or something), we wouldn't necessarily need to return a Promise, would we?
Yes, this could work in the same way but with an actual value instead of a promise. A cool thing to note is that since javascript isn't *really* parallel, you don't need any kind of flow control since the entire function will run to completion even if you have some incredibly expensive value to compute, you don't have to worry about race conditions or anything

autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.tsx
ryan-conn: Yeah I was planning on implementing this a bit differently at first but you're right, I made the change

gannonprudhomme (approved):

PR #478, opened 2020-12-04
Fix term select after redux-rearrange

gannonprudhomme:
## Description

> This PR is based off of #472, which is based on #349

In essence, this fixes all of the issues with term mismatches when switching terms that we discovered in #349, including:

- Course cards' `get_saved_courses` (filling in the course title + card info)
- Course cards' `api/sections` (filling in the sections for sessions)
- Saved schedules
- Availabilities

I would highly suggest going commit-by-commit with this, as I tried to make the commits self-contained in their changes.

Note that the commit `Made ADD_COURSE_CARD reject mismatched terms` (15c0d65) is what fixes the `get_saved_courses` issues from [Ryan's most recent comment](https://github.com/aggie-coding-club/Rev-Registration/pull/349#issuecomment-727628534) on #349.

I must have spent like 15 hours on term select at this point

## Rationale

So there are actually two "bugs" in this, but they're basically the same thing. If you do the steps mentioned in `How to test` for schedules or availabilities, it will correctly ignore the mismatched term's data but will hide the loading indicator b/c the loading indicator is just a boolean state. I think you could fix this by somewhere alongside the `showLoadingIndicator` state (like convert it to a dict or another state variable) store the term, but I mentioned it before and I think we said it was fine. Regardless, here's a gif of it:

schedules-bug

## How to test

The easiest way to test it is to add `time.sleep(3)` in the respective API view in either `user_sessions/views.py` or `scraper/views.py` for what you're trying to test for.

For instance, if you were testing schedules, you would fill in save 2 different schedules in each term. Next, you'd add a `time.sleep(3)` in `get_saved_schedules` in `users_sessions/views.py`. Then, you would reload the page, have it restore the term and start the loading of the schedules, then change the term and make sure it doesn't restore the initial term's schedules.

### DIFF
--- a/autoscheduler/frontend/src/tests/redux/Availability.test.ts
+++ b/autoscheduler/frontend/src/tests/redux/Availability.test.ts
@@ -483,7 +483,7 @@ describe('Availabilities', () => {
   });
 
   describe('skips set availabilities', () => {
-    test('when theres a term mismatch', () => {
+    test("when there's a term mismatch", () => {
       // arrange
       const store = createStore(autoSchedulerReducer);
       store.dispatch(setTerm('202031'));
--- a/autoscheduler/frontend/src/tests/redux/Availability.test.ts
+++ b/autoscheduler/frontend/src/tests/redux/Availability.test.ts
@@ -10,6 +10,7 @@ import {
 import 'isomorphic-fetch';
 import Availability, { AvailabilityType, argsToAvailability, AvailabilityArgs } from '../../types/Availability';
 import DayOfWeek from '../../types/DayOfWeek';
+import setTerm from '../../redux/actions/term';
 
 /**
  * Converts a pair of hours and minutes into a number of minutes past midnight
@@ -464,6 +465,7 @@ describe('Availabilities', () => {
     test('when setAvailabilities is called', () => {
       // arrange
       const store = createStore(autoSchedulerReducer);
+      store.dispatch(setTerm('202031'));
       const expected: Availability[] = [{
         ...dummyArgs,
         startTimeHours: 10,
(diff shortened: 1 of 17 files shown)


autoscheduler/frontend/src/tests/redux/Schedule.test.ts
```diff
@@ -446,4 +447,45 @@ describe('Schedule Redux', () => {
...
+      store.dispatch(setSchedules([fullSchedule1], '202031'));
+
+      // assert
+      expect(store.getState().termData.schedules.length).toEqual(1);
+      expect(store.getState().termData.schedules[0]).toEqual(fullSchedule1);
+    });
+  });
+
+  describe('skips set schedules', () => {
+    test('when theres a term mismatch', () => {
```
ryan-conn, line 473: Same as the other apostrophe comment

autoscheduler/frontend/src/tests/redux/CourseCards.test.ts
```diff
@@ -469,12 +472,116 @@ describe('Course Cards Redux', () => {
       expect(store.getState().termData.courseCards[0].course).toEqual('CSCE 121');
       expect(store.getState().termData.courseCards[1].course).toEqual('MATH 151');
     });
+
+    describe('rejects an update when theres a term mismatch', () => {
```
ryan-conn, line 476: Same as other apostrophe comment

autoscheduler/frontend/src/tests/redux/Availability.test.ts
```diff
@@ -473,10 +475,32 @@ describe('Availabilities', () => {
...
-      store.dispatch(setAvailabilities(expected));
+      store.dispatch(setAvailabilities(expected, '202031'));
 
       // assert
       expect(store.getState().termData.availability).toEqual(expected);
     });
   });
+
+  describe('skips set availabilities', () => {
+    test('when theres a term mismatch', () => {
```
ryan-conn, line 486: This is really minor but it slightly bugs me how some of the tests with apostrophes are named, here are the 2 best options:
- Use quotes for the string itself (ESlint is fine with this): `"there's"`
- Don't use a contraction: `'there is'`

ryan-conn (changes requested):
I'm still occasionally able to get schedules and availabilities to appear for the wrong term, although I've never had it happen with course cards by just spamming the term select, I'm pretty sure it's because the resetting of term data isn't all happening in the same place, so for a moment the redux store will look like this (keep in mind I didn't have any data for `202031`, they're carried over from the previous term):
image
You might be able to fix the issue by simply setting all of the term data in one action instead of having the unmount cleanup functions deal with resetting their own parts of state (meaning we shouldn't need any of the new `clear*` actions since they'll be taken care of in `setTerm`)

That's my only real complaint with this, the code is pretty clean and the loading indicator issue doesn't bother me *too* much. It would be nice to fix in this but the term select PR is already going to be gigantic as-is so we can come back to it later if we decide it needs fixing

ryan-conn (approved):
This looks good to me now. The only issue is that there are a few linter errors caused by the branch this is based off of:
image

So once this is merged into that branch, make sure those get fixed

PR #480, opened 2020-12-11
Instructional method icons

### LINKED ISSUE #432: Show instructional methods next to section numbers
To deal with the fact that there are now 5 instructional methods that are used for thousands of sections, we need to display this information to users. After #431, we should display the instructional method in a similar way to the following:
image
Note that we'll probably have to refactor section info to use `display: grid` instead of a table, as tables simply lack the flexibility to clearly display long instructional methods alongside remaining seats on low resolutions.

ryan-conn:
## Description

Adds icons for instructional methods.

## Rationale

We've discussed whether we like all the icons here so there shouldn't be anything too controversial in this PR

## How to test

You can see most of the icons by looking at courses for MATH 151, ENGL 210, and JAPN 491, see `InstructionalMethodIcon.tsx` to look at all of them

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|
| image | image |

## Related tasks
Closes #432.

### DIFF
--- a/autoscheduler/frontend/src/types/Section.ts
+++ b/autoscheduler/frontend/src/types/Section.ts
@@ -1,6 +1,19 @@
 import Instructor from './Instructor';
 import Grades from './Grades';
 
+// Instructional methods from the backend, make sure these are synced with Django's Section model
+export enum InstructionalMethod {
+  F2F = 'Face to Face',
+  INTERNSHIP = 'Internship',
+  NONTRADITIONAL = 'Non-traditional',
+  WEB_BASED = 'Web Based',
+  STUDY_ABROAD = 'Study abroad',
+  NONE = '',
+  REMOTE = 'Remote Only',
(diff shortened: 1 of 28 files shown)


autoscheduler/frontend/src/types/Section.ts
```diff
@@ -1,6 +1,19 @@
 import Instructor from './Instructor';
 import Grades from './Grades';
 
+// Instructional methods from the backend, make sure these are synced with Django's Section model
```
gannonprudhomme, line 4: ya know this is the one of the down sides with having our backend and frontend written in separate languages. It would've been nice to share the same type files across backend and frontend just so we know we're always synced (also not switching between camel case and snake case would be nice)

gannonprudhomme (approved):
Nice job with this, especially with that map (chef kiss emote here cause apparently github doesn't have it). Also really like how the laptop icon looks

Issue #354, opened 2020-09-17
Move "include full sections" to course cards


gannonprudhomme:
First, we should remove the "include full sections" from the configure card in general and make it a per-coursecard setting, as that's really what it should be. We'll also probably only want it to show in the Basic customization level, as you're free to choose whatever section you want in the Section customization level.

We'll also want to save this setting in the course card session, so this will need a backend change along with. Also will have to change the scheduling algorithm slightly.

PR #482, opened 2020-12-14
Move include full section into course card

### LINKED ISSUE #393: Move include full sections to basic customization level in SectionSelect
You shouldn't have to select `Include full sections` when you're in `Section` customization level and have a specific section chosen, and I think it makes a lot more sense to have it on a course by course basis. As such,  the logical conclusion is to move it to the basic customization level (probably above the filters?)

We'll also need to change `create_schedules` on the backend to handle these changes accordingly.

ThatJuanGuy:
## Description

This moves the include full sections toggle into the course cards and eliminates the configure card

## Rationale

This provides more control to the user when deciding which courses to include full sections for. It also removes the configure card because there is no longer any reason for it to exist.

## How to test

I didn't make any new tests for this because I thought it should be covered by the existing tests. You can also select a course(s) with restrictions such that a schedule can only be generated if full sections are included and try to generate a schedule. If it only returns schedules when the include full is included for those courses, it should be working.

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #393 #354


gannonprudhomme:
You'll need to merge master into this as you have quite a bit of conflicts

autoscheduler/frontend/src/redux/actions/schedules.ts
```diff
@@ -66,7 +66,7 @@ export const errorGeneratingSchedulesMessage = 'There was an error generating sc
  * throws an error with a message indicating what happened.
  * @param includeFull: Whether to generate schedules including sections with no empty seats
```
ryan-conn, line 67: Remove this comment since it's no longer a param

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicSelect.tsx
```diff
@@ -20,6 +23,10 @@ const BasicSelect: React.FC<BasicSelectProps> = ({ id }) => {
   const hasAsynchronous = useSelector<RootState, boolean>(
     (state) => state.courseCards[id].hasAsynchronous || false,
   );
+  const includeFull = useSelector<RootState, boolean>(
+    (state) => state.courseCards[id].includeFull || false,
+  );
+  const dispatch = useDispatch();
 
   // shows placeholder text if no course is selected
```
ryan-conn, line 31: I noticed that if a course doesn't have online or honors sections, the placeholder text will show up and disallow users from using the include full checkbox (ex: JAPN 101 in fall 2020). This should be changed so it's always visible, but I'm not sure if it would be more intuitive to leave the placeholder text below it or just remove it entirely. For now I'd leave it and we could discuss what we think in the next meeting

autoscheduler/frontend/src/redux/actions/schedules.ts
```diff
@@ -101,6 +101,8 @@ ThunkAction<Promise<void>, RootState, undefined, ReplaceSchedulesAction | Select
           honors: isBasic ? (courseCard.honors ?? filterDefault) : filterDefault,
           remote: isBasic ? (courseCard.remote ?? filterDefault) : filterDefault,
           asynchronous: isBasic ? (courseCard.asynchronous ?? filterDefault) : filterDefault,
+          // TODO something about full sections goes here?
+          includeFull: isBasic ? (courseCard.includeFull) : false,
```
ryan-conn, line 105: You should remove this comment

I also noticed that this line makes it so that section select won't include full sections unless you select sections manually. I don't know exactly how I feel about this since while I assume most users would prefer this, it makes the behavior different when no sections vs. all sections are selected. We can keep it as-is for now but I just wanted to point this out and maybe discuss it in the next meeting

autoscheduler/scheduler/utils.py
```diff
@@ -75,11 +75,13 @@ class CourseFilter(NamedTuple):
                       then all are valid
         honors: Whether to include honors, exclude honors, or show only honors sections
         remote: Whether to include remote, exclude remote, or show only remote sections
+        #TODO SHOULD THERE BE A NOTE ABOUT ASYNC HERE?
```
ryan-conn, line 78: There should, can you add it?

autoscheduler/scheduler/views.py
```diff
@@ -83,14 +85,13 @@ def post(self, request):
...
-        include_full = query["includeFull"]
 
         num_schedules = 5
 
         schedules = []
         message = ''
         try:
-            schedules = create_schedules(courses, term, unavailable_times, include_full,
+            schedules = create_schedules(courses, term, unavailable_times,
                                          num_schedules)
```
ryan-conn, line 95:
```suggestion
            schedules = create_schedules(courses, term, unavailable_times, num_schedules)
```
This fits on one line now, hooray

ryan-conn (changes requested):
For the most part this looks good, besides some small cleanup the only issue I found is that you can't use the checkbox on sections that don't have online or honors sections.

gannonprudhomme:
Also reminder you need to update this with master

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicSelect.tsx
```diff
@@ -44,6 +55,7 @@ const BasicSelect: React.FC<BasicSelectProps> = ({ id }) => {
       <FormLabel>Options</FormLabel>
       <table className={styles.tableContainer}>
         <tbody>
+          <BasicCheckbox id={id} value="includeFull" label="Include Full Sections" />
           {hasHonors
             ? <BasicOptionRow id={id} value="honors" label="Honors" />
             : null}
```
ryan-conn, line 61: I'd appreciate if you could rewrite this to only have one table (instead of creating an identical table for the "there are no honors or remote sections" text and another if there are honors/remote/async sections). Instead of returning the entire table if there are no honors/remote sections, determine the correct thing to render (either the `BasicOptionRow`s or the placeholder text) and store it in a variable

autoscheduler/scheduler/tests/api_tests.py
```diff
@@ -39,15 +39,17 @@ def test_parse_course_filter_is_correct(self):
...
         expected = CourseFilter(subject="CSCE", course_num="121", section_nums=["500"],
                                 honors=BasicFilter.EXCLUDE, remote=BasicFilter.EXCLUDE,
-                                asynchronous=BasicFilter.EXCLUDE)
+                                asynchronous=BasicFilter.EXCLUDE, include_full=False)
 
         # Act
         result = _parse_course_filter(course)
-
+        print("EXPECTED: ", expected)
+        print("RESULT: ", result)
```
ryan-conn, line 52: Remove these prints
```suggestion
```

ryan-conn (changes requested):
A couple of small comments

gannonprudhomme:
Merge with master when you can

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/BasicSelect/BasicSelect.tsx
```diff
@@ -21,6 +24,34 @@ const BasicSelect: React.FC<BasicSelectProps> = ({ id }) => {
...
+  const noOptionsMessage = (
+    <>
+      {(!hasHonors && !hasRemote && !hasAsynchronous)
+        ? (
+          <Typography className={styles.placeholderText} color="textSecondary">
+            There are no honors or remote sections for this class
+          </Typography>
+        ) : null}
+    </>
+  );
```
ryan-conn, line 53:
```suggestion
  const noOptionsMessage = hasHonors || hasRemote || hasAsynchronous ? null : (
    <Typography className={styles.placeholderText} color="textSecondary">
      There are no honors or remote sections for this class
    </Typography>
  );
```
`null` JSX elements render nothing, so moving the ternary to the outside makes it more clear what this is doing

autoscheduler/scheduler/create_schedules.py
```diff
@@ -102,8 +102,8 @@ def _get_meetings(course: CourseFilter, term: str, include_full: bool,
     # Get id for each valid section to filter and order meeting data
     # Also removes full sections if include_full is False
     sections = sections.values('id', 'current_enrollment', 'max_enrollment')
-    # If manually selected don't check if section is full before adding
-    if course.section_nums or include_full:
+    # if manually selected don't check if section is full before adding
```
ryan-conn, line 105:
```suggestion
    # If manually selected, don't check if section is full before adding
```
Keep this capital like it was before since that's how we format python comments

ryan-conn (changes requested):
A couple of pedantic commenst and this is good

ryan-conn:
Make sure you rebase this with master to fix the linting issues and implement my change about `noOptionsMessage`

ryan-conn (approved):
Looks good now. When I looked at this more the spacing issue I pointed out during the last meeting also happens in master so I'll make a separate issue for it.

PR #485, opened 2020-12-15
Sort section select

### LINKED ISSUE #353: Add sorting to SectionSelect
I feel like you should be able to sort courses during SectionSelect (e.g. if you did actually want them in pure section order), as it would probably help.

Might also be nice to have a filter to ignore sections that intersected with your marked availabilities.

I do think we'd have to get really clever with this design, as the SectionSelect is already pretty cluttered as is

eelyort:
## Description

Adds a menu and functionality for sorting the sections in section select.

## Rationale

Removed 4 previous unit tests because they tested the default sorting assuming sorting was done on the frontend, however, it is now done on the backend

## How to test

Change to section select on a large class such as ENGR 102 and click buttons in sort menu.

## Screenshots

pr1

## Related tasks

Closes #353


eelyort:
Realized all the changes from master were in these commits so I'm going to remerge master via cherry-picking to make it easier.

PR #497, opened 2021-01-25
Times are no longer covered by adjacent meetings

firejake308:
## Description

We currently have a CSS bug where the start/end times for a meeting can be covered by an adjacent meeting if there's not enough time between classes. It looks like we've tried to fix this before, because we already have a style rule that increases the `z-index` on hover, but it's not being applied due to the element style rule taking priority. I fixed this by adding `!important` to the `.schedule-card:hover` style rule.

## Rationale

I could have also considered changing the `zIndex` that is applied in JavaScript at line 82 of `ScheduleCard.tsx`, but `!important` felt less invasive and less likely to break other things.

## How to test

Add PHYS 202-513 to your schedule. The recitation and lab times are back to back, so this should generate an overlap on GCP but you should be able to see the time on this branch.

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

#219

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/Schedule/ScheduleCard/ScheduleCard.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/Schedule/ScheduleCard/ScheduleCard.css
@@ -14,7 +14,8 @@
 }
 
 .schedule-card:hover {
-    z-index: 4;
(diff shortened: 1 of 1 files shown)


gannonprudhomme (approved):
ezpz

Issue #297, opened 2020-07-13
Add professor names to schedule preview


ryan-conn:
Since mini-schedules take up more vertical than the schedule's text in most situations, we should be able to put instructor names on a separate line under section numbers without making schedules much larger. It should probably be grey text and smaller than other information.

PR #503, opened 2021-01-31
Schedule preview improvements

### LINKED ISSUE #491: Make schedule preview not show colors for asynchronous sections
We shouldn't show color badge icon things in the Schedule Preview when the sections are asynchronous (have no meetings)

ryan-conn:
## Description

Replaces color boxes for sections with no meeting times with an X, and shows instructor names in the preview

## Rationale
I based this branch off of `schedule-details` because I reorganized the schedule preview significantly in that branch. No tests because I don't think they'd be very helpful but I could try to add some if we want.

I also resized `<ColorBox>` to be 1px bigger because the old size made it impossible to center the X, it would always be 1px off because of rounding

## How to test
Mess around with adding async classes, also try resizing the window to see how the instructor names interact when cut off (they should truncate the text and add an ellipsis)

## Screenshots

Before|After
--|--
image|image

## Related tasks

Closes #491, #297.

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/ScheduleListItem.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/ScheduleListItem.tsx
@@ -59,13 +59,17 @@ const ScheduleListItem: React.FC<ScheduleListItemProps> = ({ index, onDetailsCli
     </ListItemSecondaryAction>
   );
 
-  const scheduleSections = sectionsForSchedule(schedule).map((sec: Section) => (
-    <span key={sec.id} className={styles.sectionLabelRow}>
-      <ColorBox color={meetingColors.get(sec.subject + sec.courseNum)} />
-      {`${sec.subject} ${sec.courseNum}-${sec.sectionNum}`}
-      <br />
-    </span>
-  ));
+  const scheduleSections = sectionsForSchedule(schedule).map((sec: Section) => {
+    const color = sec.asynchronous ? undefined : meetingColors.get(sec.subject + sec.courseNum);
+
+    return (
+      <span key={sec.id} className={styles.sectionLabelRow}>
+        <ColorBox color={color} />
+        {`${sec.subject} ${sec.courseNum}-${sec.sectionNum}`}
+        <br />
+      </span>
+    );
+  });
 
   const scheduleItemContent = (
     <span className={styles.scheduleContentContainer}>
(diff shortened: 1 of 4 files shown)


autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/ScheduleListItem.tsx
```diff
@@ -59,13 +59,23 @@ const ScheduleListItem: React.FC<ScheduleListItemProps> = ({ index, onDetailsCli
...
 
-  const scheduleSections = sectionsForSchedule(schedule).map((sec: Section) => (
-    <span key={sec.id} className={styles.sectionLabelRow}>
-      <ColorBox color={meetingColors.get(sec.subject + sec.courseNum)} />
-      {`${sec.subject} ${sec.courseNum}-${sec.sectionNum}`}
-      <br />
-    </span>
-  ));
+  const scheduleSections = sectionsForSchedule(schedule).map((sec: Section) => {
+    const color = !sec.asynchronous && meetingColors.get(sec.subject + sec.courseNum);
```
gannonprudhomme, line 63: This is like an alternative form of a ternary where the false case is undefined/null right?

gannonprudhomme (approved):
This is pretty straightforward and I'll approve it since I think it's mostly good and achieves what it's supposed to (the red X looks much better too), however:

- I'd honestly wait to merge this into schedule details and instead just merge it into master once schedule details is merged, but up to you (I wish I did that for term select and last updated indicator since they're loosely related but technically separate tbh)
- Honestly I think I still would like the instructors to be left-aligned with each other instead of being slightly offset. I think not having them aligned makes the schedule preview feel way more cluttered/disorganized, but ik y'all said otherwise last time we discussed this, so if you _really_ disagree then you can leave it (I do think it's work making a demo of it and showing it tho)

autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/ScheduleListItem.tsx
ryan-conn: Yeah although I probably should've just explicitly set it to undefined to make it more clear

gannonprudhomme: Yeah I'd say so, honestly I don't really like this syntax that much in this scenario (I'm more inclined to favor it when it's in actual JSX)

ryan-conn:
Updated with the updated schedule details PR and addressed Gannon's comments


> I'd honestly wait to merge this into schedule details and instead just merge it into master once schedule details is merged, but up to you (I wish I did that for term select and last updated indicator since they're loosely related but technically separate tbh)

Just for clarity's sake this is what I decided to do

gannonprudhomme:
Think you need to hard reset and cherry pick these

ryan-conn:
> Think you need to hard reset and cherry pick these

Correct, I really didn't want to deal with 100+ lines of merge conflicts so that's what I ended up doing

Conveniently I noticed that the sections for schedules were throwing errors because I forgot to give them keys, so I fixed that as well. Other than that this should be identical to before I cherry-picked

gannonprudhomme (approved):

PR #505, opened 2021-02-07
Term-select: Fixed get_all_terms

gannonprudhomme:
## Description

So it turns out that `get_all_terms` was working completely wrong (you can check by removing the changes I made and running the tests). This fixes it and adds tests for it.

## How to test

Just check that the tests are running how you would expected

### DIFF
--- a/autoscheduler/scraper/management/commands/utils/scraper_utils.py
+++ b/autoscheduler/scraper/management/commands/utils/scraper_utils.py
@@ -2,18 +2,20 @@ from datetime import datetime
 from typing import List
 from itertools import product
 
-def get_all_terms(year: int = -1) -> List[str]:
+def get_all_terms(year: int = -1, now=datetime.now()) -> List[str]:
     """ Generates all of the terms, from 2013 until now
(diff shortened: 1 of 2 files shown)


autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -22,9 +24,28 @@ def get_all_terms(year: int = -1) -> List[str]:
...
+    # Now we need to decide what of the recent terms we're going to include in this
+
+    date_format = '%m/%d/%Y'
+    summer_fall_reg_start = datetime.strptime(f'03/22/{current_year}', date_format)
+    spring_reg_start = datetime.strptime(f'10/26/{current_year}', date_format)
+
+    if year == -1 or year == current_year:
+        semesters = []
+
+        if now < summer_fall_reg_start:
```
gannonprudhomme, line 37: I'm almost positive this can be simplified but I can't think of anything

ryan-conn (approved):

PR #508, opened 2021-02-14
Reverse schedule preview course direction

### LINKED ISSUE #499: Reverse schedule preview course direction
Because we reversed the direction that course cards are added for the course select column, we also need to reverse the direction for the courses in the SchedulePreview. Should be a super quick fix.

|Course cards|SchedulePreview|
|--|--|
|||

gannonprudhomme:
## Description

Reverses the schedule preview course direction to map the reversed CourseSelectColumn order. I _could_ probably add tests to this, but it's so dang simple that I really didn't think it was necessary.

## How to test

Add two courses, generate a schedule, and ensure they match up with the CourseSelectColumn direction

## Screenshots

|Before|After|
|--|--|
|chrome_vL6NRpSMSS|chrome_bQYmDWQs7E|

## Related tasks

Closes #499


autoscheduler/frontend/src/components/SchedulingPage/SchedulePreview/ScheduleListItem/ScheduleListItem.tsx
```diff
@@ -84,7 +84,8 @@ const ScheduleListItem: React.FC<ScheduleListItemProps> = ({ index }) => {
               {`${sec.subject} ${sec.courseNum}-${sec.sectionNum}`}
               <br />
             </span>
-          ))
+          // Reverse the direction that the sections are displayed in to match to the CourseCards
```
ryan-conn, line 87:
```suggestion
          // Reverse the direction that the sections are displayed in to match the CourseCards
```

ryan-conn (changes requested):
Not much to comment on except a typo I noticed, works fine

Also probably worth pointing out that I should make the same change to the schedule details PR since it has the same reversed order as this

ryan-conn:
I just noticed we don't actually need this PR since I abstracted the logic mapping schedules to sections into its own function (so that I could reuse it for schedule details). Now that I added the reverse in the schedule details PR it also applies to the schedule list items

gannonprudhomme:
oh neato

PR #511, opened 2021-02-14
Extended FIRST_HOUR and LAST_HOUR by 1 hour

### LINKED ISSUE #370: Extend LAST_HOUR by 1 hour
Both CSCE 121 and MATH 151 have tests that go till 10:30 PM (those bastards), so we should extend it by 30 minutes as our current limit is 10:00 PM

gannonprudhomme:
## Description

Extends FIRST_HOUR and LAST_HOUR for the schedule by 1 hour.

## Rationale

Classes such as KINE-199 have early times, so we have to account for them. Likewise, MATH classes frequently have exams that end at 10:30, so we need to account for it as well.

## How to test

KINE 199-598 and MATH 151-521 for 202111 have early and late times. Likewise, CSCE 482 has 15 minute times

## Screenshots

|Before|After|
|--|--|
|chrome_Zk4h90oTTB|chrome_chRyg1wgfP|

## Related tasks

Closes #370

### DIFF
--- a/autoscheduler/frontend/src/tests/ui/Availability.test.tsx
+++ b/autoscheduler/frontend/src/tests/ui/Availability.test.tsx
@@ -12,7 +12,7 @@ import { createStore, applyMiddleware } from 'redux';
 import thunk from 'redux-thunk';
 import autoSchedulerReducer from '../../redux/reducer';
 import Schedule from '../../components/SchedulingPage/Schedule/Schedule';
-import { timeToEvent, LAST_HOUR } from '../../utils/timeUtil';
+import { timeToEvent, LAST_HOUR, FIRST_HOUR } from '../../utils/timeUtil';
 import setTerm from '../../redux/actions/term';
 
 describe('Availability UI', () => {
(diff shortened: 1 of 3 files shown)


autoscheduler/frontend/src/tests/ui/Availability.test.tsx
```diff
@@ -122,7 +122,7 @@ describe('Availability UI', () => {
         .toHaveAttribute('aria-valuetext', expectedEnd);
     });
 
-    test('with an end time of 10 PM and a size of 30 mins if dragged below the bottom', async () => {
+    test('with an end time of LAST_HOUR-1 PM and a size of 30 mins if dragged below the bottom', async () => {
```
firejake308, line 125:
```suggestion
    test('with an end time of LAST_HOUR PM and a size of 30 mins if dragged below the bottom', async () => {
```

firejake308: Since the expected end time is 11 PM and the LAST_HOUR is also 11 PM

firejake308 (changes requested):
I've attached a screenshot of how it looks on my monitor, which I expect is the smallest monitor that anyone will try to use the website on. There's not much to be done about super-early KINE 199's, and at least if you hover over it, it tells you the exact start and end time. CSCE 482 is just *barely* legible on my screen, so I'd say this is fine one you adjust the comment I pointed out.

image

autoscheduler/frontend/src/tests/ui/Availability.test.tsx
gannonprudhomme: Great catch!

firejake308 (approved):

PR #512, opened 2021-02-14
Basic options default to no preference

### LINKED ISSUE #501: Fix remote and no-meeting-times defaults to no preference
By default, the remote and no meeting times fields for a new course card should be `no preference` instead of `exclude` (like it is now). We should fix it and add tests to catch it next time (which is sadly my fault cause I did this before in #345 )': )

firejake308:
## Description

Default for BasicOptions is now No Preference instead of Exclude. Also added test.

## Rationale

We attempted to make this change a while ago in #345, but there are multiple places that reference a default value for basic options, and one of them was missed.

## How to test

I created a test to verify that the change worked. The test failed when I first wrote it, but passed after I made the change in `BasicOptionsRow.tsx`. You can also check in browser to see that it works.

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #501


firejake308:
Fixed Gannon's comment from last meeting

gannonprudhomme (changes requested):

gannonprudhomme:
Gotta fix the conflicts :P

firejake308:
Thanks for the reminder, I almost forgot

gannonprudhomme (approved):
🚀

PR #513, opened 2021-02-15
User session fix

### LINKED ISSUE #495: Fix log in auto-redirect bug
For some reason some accounts on the GCP cannot be logged in, and when you do log in it just eternally get redirected to the main page.

ryan-conn:
## Description

Fixes the issue where users would end up with invalid sessions, leaving them stuck on the homepage for all of eternity

More concretely:
- Sessions now take 10 years to expire instead of 2 weeks, meaning no more users should run into the issue (unless they don't use the site for 10 years in which case we should probably delete their data anyways). 
- If a user ends up with an invalid `Session` object associated with their `UserToDataSession`, a new one will be created (to fix the sessions of users who have already encountered the issue)

## Rationale
This performs an additional query every time `retrieve_data_session` is called in order to check the `Session` is valid, as there's no way to determine this with 100% accuracy using only a `SessionStore` object: empty sessions are treated the same as ones that don't exist.

## How to test
You can check if your affected accounts on deployment are fixed, as well as manually mess around with your `user_to_data_session` and `django_session` tables locally to test different situations and what happens

## Related tasks
Closes #495.

### DIFF
--- a/.github/workflows/deploy-workflow.yml
+++ b/.github/workflows/deploy-workflow.yml
@@ -1,8 +1,5 @@
 name: GCP Deploy
-on:
-  push:
-    branches:
-      - production
+on: push
 
 jobs:
   deploy-app-engine:
--- a/.github/workflows/deploy-workflow.yml
+++ b/.github/workflows/deploy-workflow.yml
@@ -1,5 +1,8 @@
 name: GCP Deploy
-on: push
+on:
+  push:
+    branches:
+      - production
 
 jobs:
   deploy-app-engine:
--- a/autoscheduler/autoscheduler/settings/base.py
+++ b/autoscheduler/autoscheduler/settings/base.py
@@ -63,6 +63,9 @@ LOGIN_URL = '/auth/login/google-auth2/' #or try -oauth2?
 LOGIN_REDIRECT_URL = '/'
 LOGOUT_REDIRECT_URL = '/'
 
+# SESSION_COOKIE_AGE must be an integer - make sessions take 10 years to expire
+SESSION_COOKIE_AGE = 10 * 365 * 24 * 60 * 60
+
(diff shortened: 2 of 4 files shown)


.github/workflows/deploy-workflow.yml
```diff
@@ -1,8 +1,5 @@
 name: GCP Deploy
-on:
-  push:
-    branches:
-      - production
+on: push
```
gannonprudhomme, line 2: Reminder to undo this

gannonprudhomme (dismissed):
I'm still not sure why this isn't working for me locally, but w/e it works on GCP and that's good enough for me

ThatJuanGuy:
So it works for me on GCP as well but locally one of my accounts I used to test this isn't working. It's still trying to access an expired session.

gannonprudhomme:
> So it works for me on GCP as well but locally one of my accounts I used to test this locally isn't working. It's still trying to access an expired session.

Should we halt merging this until we get it to work locally? I can't think of how we'd test it tbh

ThatJuanGuy:
I think it's because we deleted expired sessions on GCP but not locally

ryan-conn:
> I'm still not sure why this isn't working for me locally, but w/e it works on GCP and that's good enough for me

> So it works for me on GCP as well but locally one of my accounts I used to test this locally isn't working. It's still trying to access an expired session.

Should be fixed now, the problem was for sessions that were expired but not deleted (which still exist on deployment but are more rare because a lot of them got purged). I made it so that expired sessions have their expiration time updated, and therefore the `SessionStore` object will correctly fetch the session instead of ignoring the existing one.

gannonprudhomme (approved):
Thanks again for picking this up from me, I am happy you came to a fix so I didn't have to 😅. Ryan to the rescue once again!

PR #494, opened 2021-01-14
About us

gannonprudhomme:
## Description

This adds the about us credits dialog to the landing page.

## How to test

Click `About Rev Registration` on the landing page and make sure it's formatted correctly

## Screenshots

chrome_QpWc2gOFVz

chrome_8dJLUZmyJN

## Related tasks

Closes #174

PR #514, opened 2021-02-21
Fixed backend linting (on GitHub Actions)

gannonprudhomme:
## Description

I'm pretty sure the linting Action workflow we use, [GitHub Action for pylint](https://github.com/cclauss/GitHub-Action-for-pylint), upgraded it's pylint version, which is causing a few PR's to fail when you wouldn't expect them to (#511, #508, #494). This updates `pylint` (and `pylint_django`) to the most recent version, as well as fixes the linting errors that arose. As such, make sure to run `pip install -r requirements.txt` after this is merged so that you're up-to-date.

### NOTE

You'll need to fix your linting in VSCode by going to your `.vscode/settings.json` and adding the last line, `--django-settings-module=autoscheduler.settings.base` to your pylint arguments, which means it'll look like:

```json
	"python.linting.pylintArgs": [
		"--load-plugins",
		"pylint_django",
		"--django-settings-module=autoscheduler.settings.base"
	],
```

frankly I'm pretty sure it can be changed to

```json
	"python.linting.pylintArgs": [
		"--load-plugins=pylint_django",
		"--django-settings-module=autoscheduler.settings.base"
	],
```

but I don't think it matters.

I also updated the [Setup Pylint Wiki](https://github.com/aggie-coding-club/Rev-Registration/wiki/Setup-Pylint) with this info.

## How to test

If it passes the Actions linting step then we're good

### DIFF
--- a/.github/lint-requirements.txt
+++ b/.github/lint-requirements.txt
@@ -2,7 +2,7 @@ aiohttp==3.6.2
 beautifulsoup4==4.8.2
 Django~=2.2.6
 djangorestframework~=3.9.2
-pylint-django==2.0.11
+pylint-django==2.4.2
 PyPDF2==1.26.0
 requests==2.21.0
 python-dotenv==0.13.0
--- a/.github/workflows/backend-workflow.yml
+++ b/.github/workflows/backend-workflow.yml
@@ -20,4 +20,4 @@ jobs:
         uses: cclauss/GitHub-Action-for-pylint@master
         with:
         # Installs Django and pylint-django, then runs pylint on all of the files in autoscheduler/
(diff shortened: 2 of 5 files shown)


ryan-conn:
This is really petty but make sure you have git line endings configured correctly, you changed a few files from LF to CRLF

gannonprudhomme:
Fixed it - I'm pretty sure `banner_request_tests.py` is LF line endings and `banner_requests.py` is CRLF endings (and I think my git changed it from CRLF -> LF). Only the frontend actually requires files to be LF line endings

ryan-conn (approved):

Issue #496, opened 2021-01-20
Add an information page


gannonprudhomme:
We should add a page / dialog that discusses various things about the website, including:

- How often we scrape and such
- Maybe more "hidden" features of the website (like saving a schedule?)
- Origin of the project? Why we started, what we're aiming for, the fact that it's student made and such

gannonprudhomme:
Make sure you base this off of Gannon's privacy policy branch

PR #515, opened 2021-02-21
Privacy policy

### LINKED ISSUE #408: Add privacy policy page / popup

gannonprudhomme:
## Description

This adds the privacy policy to the landing page. This also adds an abstract `InfoDialog` component which both `AboutUs` and `PrivacyPolicy` use to remove the boilerplate/duplicate code from them.

Note that the method I used to center the links isn't very solid, but I know it's going to have to change anyways when Carlos does #496, so I think what I have now is probably fine.

## How to test

Read the privacy policy and make sure you agree with everything it says on it.

## Screenshots

|Landing Page|Privacy Policy|
|--|--|
|chrome_VPpn3rUM4k|chrome_5afGSIAul3|

## Related tasks

Closes #408


autoscheduler/frontend/src/components/LandingPage/LandingPage.css
```diff
@@ -6,15 +6,13 @@
 }
 
 .dialog-container {
-    width: 100%;
+    padding-top: 32px;
```
ryan-conn, line 9: Did you mean to add this in the latest commit? I don't remember us talking about changing this spacing, and both this and `dialog-link` have 32px top padding now which looks a bit odd to me. If it was intentional, you should only put the padding on one of the elements 

| Before | After |
|-|-|
| image | image |

ryan-conn (changes requested):
Looks good except for a change from the latest commit that I don't think was intentional, see my comment

autoscheduler/frontend/src/components/LandingPage/LandingPage.css
gannonprudhomme: Correct yeah I definitely didn't meant to do that lol, will undo

ryan-conn (approved):

gannonprudhomme:
I somehow missed that you approved this lmao

PR #524, opened 2021-03-21
Use -moz- vendor prefix for fit-content

ryan-conn:
## Description
Firefox doesn't support `fit-content` which makes it so parts of the websites that use it look odd, for example:
fit-content
This adds `-moz-fit-content` wherever `fit-content` is used to achieve the intended effect:
image

## Rationale
The basic options row is really wonky on firefox without this, and while the instructions dialog looks fine I figured I might as well change it so that it works the same across browsers.

## How to test
Use firefox lol

## Screenshots
See above

## Related tasks
None, but I pointed out this issue during the last meeting


gannonprudhomme (approved):
I can't reproduce this locally or on GCP but it doesn't regress anything so looks good to me

PR #526, opened 2021-03-21
Use production react/redux in production

### LINKED ISSUE #453: Use production versions of React/Redux on the live site
We should use the production version of React in `index.html` in production, see [this StackOverflow post](https://stackoverflow.com/questions/4247717/django-using-different-templates-for-production/4247901#4247901) for how we can do this

`index.tsx` also uses an enhancer to allow the use of Redux devtools, similarly we should remove those lines in production.

ryan-conn:
## Description
Uses minified production react when running on GCP, and doesn't give access to redux devtools when `npm run build` is run. I also improved the way globals work by letting our build tools know about their existence rather than having to manually import dummy variables to get around linter errors, since the old method only works if it uses a default export.

## Rationale
Using the new react versions in production should provide a significant speed boost (I've read ~2x-5x as fast from different sources— I'm surprised we hadn't really considered how much of a difference this might make), as well as saving ~150kB data transfer due to smaller file sizes. On the redux side I'm sure there are some small speed improvements since it doesn't need to add hooks to every dispatch.

## How to test
You can test the behavior for GCP by changing `_INDEX_CONTENT['IS_GCP']` in `autoscheduler/frontend/views.py` to `True` rather than using the environment variable, and `npm run build` to disable redux devtools.

## Related tasks
Closes #453.

### DIFF
--- a/autoscheduler/frontend/src/tsconfig.json
+++ b/autoscheduler/frontend/src/tsconfig.json
@@ -5,6 +5,10 @@
         "noImplicitAny": true,
         "module": "commonjs",
         "target": "es2018",
+        "typeRoots": [
+            "./node_modules/@types",
+            "./globals.ts",
+        ],
         "jsx": "react"
     }
 }
--- a/autoscheduler/frontend/templates/index.html
+++ b/autoscheduler/frontend/templates/index.html
@@ -50,9 +50,13 @@
         <div id="root"></div>
 
         <!-- Dependencies -->
-        <!-- I think we need these? -->
-        <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react/16.12.0/umd/react.development.js"></script>
-        <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.11.0/umd/react-dom.development.js"></script>
+        {% if IS_GCP %}
+            <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react/16.12.0/umd/react.production.min.js"></script>
+            <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.11.0/umd/react-dom.production.min.js"></script>
+        {% else %}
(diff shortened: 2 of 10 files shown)


autoscheduler/frontend/templates/index.html
```diff
@@ -50,9 +50,13 @@
         <div id="root"></div>
 
         <!-- Dependencies -->
-        <!-- I think we need these? -->
-        <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react/16.12.0/umd/react.development.js"></script>
-        <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.11.0/umd/react-dom.development.js"></script>
+        {% if IS_GCP %}
+            <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react/16.12.0/umd/react.production.min.js"></script>
+            <script crossorigin src="https://cdnjs.cloudflare.com/ajax/libs/react-dom/16.11.0/umd/react-dom.production.min.js"></script>
```
gannonprudhomme, line 55: So did you determine that we actually do need these?

autoscheduler/frontend/src/tsconfig.json
```diff
@@ -5,6 +5,10 @@
         "noImplicitAny": true,
         "module": "commonjs",
         "target": "es2018",
+        "typeRoots": [
+            "./node_modules/@types",
+            "./globals.ts",
+        ],
```
gannonprudhomme, line 11: Does this remove the need of having to export `STATIC_URL` and `IS_PRODUCTION`?

gannonprudhomme (approved):

autoscheduler/frontend/templates/index.html
ryan-conn: Yeah, with another project I was working on react imports weren't necessary but we do need them with this node/webpack setup

autoscheduler/frontend/src/tsconfig.json
ryan-conn: Yes, declaring them in `globals.ts` makes typescript recognize them and putting them in `globals` in the eslintrc will make the linter recognize them

PR #531, opened 2021-03-28
Filter duplicate meetings on schedule

### LINKED ISSUE #419: Combine meetings with the same start time on the calendar view
Overlapping meeting times show up in the section select as a single meeting that goes through :
image

However, on the calendar they show up separately as we don't call the function to combine meetings:
image

We should extract the function we use to combine meetings to a util file and call it in both of these places for consistency.

gannonprudhomme:
## Description

This adds the filtering/combining of duplicate meetings on the schedule

## How to test

There's an example in Fall 2020, CSCE 121 - 597

## Screenshots

|Before|After|
|--|--|
|chrome_Mreqe0x3kn|chrome_u3FPCoBBfw|

## Related tasks

Closes #419

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
@@ -22,7 +22,7 @@ import useMeetingColor from './meetingColors';
 import InstructionsDialog from './InstructionsDialog/InstructionsDialog';
 import createThrottleFunction from '../../../utils/createThrottleFunction';
 import SmallFastProgress from '../../SmallFastProgress';
-import { filterDuplicateMeetings } from '../../../utils/meetingsForSection';
+import meetingsForSchedule from '../../../utils/meetingsForSchedule';
 
 // Creates a throttle function that shares state between calls
 const throttle = createThrottleFunction();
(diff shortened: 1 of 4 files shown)


autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -372,7 +373,7 @@ const Schedule: React.FC = () => {
       );
     }
     return [DayOfWeek.MON, DayOfWeek.TUE, DayOfWeek.WED, DayOfWeek.THU, DayOfWeek.FRI].map(
-      (idx) => getMeetingsForDay(idx).map((mtg) => renderMeeting(mtg)),
+      (idx) => filterDuplicateMeetings(getMeetingsForDay(idx)).map((mtg) => renderMeeting(mtg)),
     );
```
ryan-conn, line 377: I'm fine with this implementation but I feel like moving all this logic to its own function in a helper file (like `meetingsForSection` in `meetingsForSection.ts`) would make this a bit cleaner, since in general it's a good idea to separate business logic from the components themselves

ryan-conn (dismissed):
I don't have a problem with this as-is but if you feel like it's a change worth making then consider implementing my comment

ryan-conn (approved):
Looks good, I always love when a PR adds functionality and ends up making the code more clean

PR #532, opened 2021-04-05
GCP Error Logging

### LINKED ISSUE #488: Extend error tracking capabilities in server / App Engine
Currently App Engine doesn't give descriptive error warnings - if you get an error 500 it won't tell you what caused it or the stack trace for it. As such, we need to look into getting more descriptive error warnings

ryan-conn:
## Description

Adds a middleware that makes it so using the `logging` module will propagate errors to GCP (errors in views will automatically be classified).

Also related to this is the fact that I set up email notifications for ERROR or CRITICAL level application logs on Stackdriver, if we want to make any custom errors we can use `logging.error('blah')` or `logging.critical('blah')` and it should notify us.

## Rationale
I used to have a custom error handler for errors encountered in views, but since the default handler for them is more robust I removed it.

## How to test
You can't haha, here's some confirmation that it works as intended though:
### Errors in views/django internals:
image

### Custom logging:
image

### Error notification:
image

## Related tasks
Closes #488.

### DIFF
--- a/autoscheduler/autoscheduler/middleware/logging_middleware.py
+++ b/autoscheduler/autoscheduler/middleware/logging_middleware.py
@@ -1,5 +1,3 @@
-import logging
-from traceback import format_exc
 from django.core.exceptions import MiddlewareNotUsed
 from autoscheduler.settings.base import _IS_GCP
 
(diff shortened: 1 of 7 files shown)


autoscheduler/autoscheduler/middleware/logging_middleware.py
```diff
@@ -24,3 +26,17 @@ def __call__(self, request):
         response = self.get_response(request)
 
         return response
+
+    def process_exception(self, request, _):
```
gannonprudhomme, line 30: Will this double the exception emails we get? E.g. if we get a 400 error (and it sends us an email), will this trigger an email as well?

ryan-conn: No, the emails for log-based errors are based on whether errors are encountered in 5-minute timeframes and since this should be logged pretty much simultaneously with the actual error itself the email alert should overlap 99% of the time

gannonprudhomme (approved):

PR #534, opened 2021-04-07
Fix instructor select bugs

firejake308:
## Description

After merging with master, we discovered that the changes to CSS grid at the end reverted some of the desired behaviors from  earlier work. Specifically, if there is a single meeting for a section, its background should be white, not red, and the section number is clearly no longer aligned with the instructional method icon. These have both been fixed by this patch.

## Rationale

Should be simple enough, as it's 2 lines of CSS. Ask if you have questions though

## How to test

Inspect visually in browser. See PSYC 107 for a class with 1-meeting sections and literally any class for the misalignes section number.

## Screenshots

resolutions/sizes.

|Before|After|
|--|--|
|image|image|

## Related tasks

Quick patch for #473

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
@@ -48,15 +48,13 @@ const SectionInfo: React.FC<SectionInfoProps> = ({
   const remainingSeatsColor = remainingSeats > 0 ? 'black' : 'red';
   // show section number and remaining seats if this is the first meeting for a section
   const sectionHeader = (
-    <Typography className={styles.denseListItem} component="div" style={{ display: 'flex', gridColumn: '1 / -1' }}>
-      <div style={{ flex: 1, display: 'flex', alignItems: 'center' }}>
+    <Typography className={`${styles.denseListItem} ${styles.sectionHeader}`} component="div">
+      <div className={styles.sectionNumAndIcon}>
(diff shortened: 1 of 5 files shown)


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
```diff
@@ -49,7 +49,7 @@ const SectionInfo: React.FC<SectionInfoProps> = ({
   // show section number and remaining seats if this is the first meeting for a section
   const sectionHeader = (
     <Typography className={styles.denseListItem} component="div" style={{ display: 'flex', gridColumn: '1 / -1' }}>
-      <div style={{ flex: 1, display: 'flex' }}>
+      <div style={{ flex: 1, display: 'flex', alignItems: 'center' }}>
```
gannonprudhomme, line 52: Is there a reason we're doing styles in `.tsx` instead of `.css`? Just don't want to make it a class?

gannonprudhomme (changes requested):
The first row is still maroon (I think you changed it for the only-1-section case?); instructional method alignment is fixed though

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
firejake308: Fixed

firejake308:
I made the first row always maroon, and I also moved the honors icon outside of the instructor button

gannonprudhomme (approved):

PR #540, opened 2021-04-23
One line navbar

### LINKED ISSUE #517: NavBar goes on two lines on small screens
Need to make it so that the navbar portion that keeps stretching onto two lines stays on one and uses the available space.

This is due to the last-updated-indicator wrapping

ryan-conn:
## Description
Makes the navbar stay on one line rather than become thicc af

## Rationale

It starts looking a bit weird when the width goes below 600px (term select clips with name):
image
but this should be fixed with the mobile functionality PR and is incredibly unlikely on desktop so I don't think it's that big a deal

## How to test

## Screenshots
lol yikes

|Before|After|
|--|--|
| image | image |

## Related tasks
Closes #517.

### DIFF
--- a/autoscheduler/frontend/src/components/LastUpdatedAt.tsx
+++ b/autoscheduler/frontend/src/components/LastUpdatedAt.tsx
@@ -110,17 +110,17 @@ const LastUpdatedAt: React.FC = () => {
   }, [lastUpdated]);
 
   return (
-    <div
+    <Typography
       style={{
-        display: 'flex',
-        alignItems: 'center',
         paddingLeft: 8,
+        whiteSpace: 'nowrap',
+        overflow: 'hidden',
+        textOverflow: 'ellipsis',
       }}
+      variant="body2"
     >
-      <Typography variant="body2">
-        {lastUpdatedText ? `Last updated ${lastUpdatedText}` : null}
-      </Typography>
-    </div>
+      {lastUpdatedText ? `Last updated ${lastUpdatedText}` : null}
+    </Typography>
   );
 };
 
--- a/autoscheduler/frontend/src/components/NavBar/NavBar.css
+++ b/autoscheduler/frontend/src/components/NavBar/NavBar.css
@@ -6,6 +6,4 @@
 
 .user-name {
     white-space: nowrap;
-    overflow: hidden;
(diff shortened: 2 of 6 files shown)


autoscheduler/frontend/src/components/NavBar/NavBar.css
```diff
@@ -1,4 +1,9 @@
-.name-and-button {
+.logout-container {
     display: flex;
     align-items: center;
+    min-width: 0;
```
firejake308, line 4: Why is this necessary?

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
```diff
@@ -110,17 +110,17 @@ const LastUpdatedAt: React.FC = () => {
...
   return (
-    <div
+    <Typography
       style={{
-        display: 'flex',
-        alignItems: 'center',
         paddingLeft: 8,
+        whiteSpace: 'nowrap',
+        overflow: 'hidden',
+        textOverflow: 'ellipsis',
```
firejake308, line 118: We should probably be doing CSS class for this, but since this is the only thing in the component that would need it, I think it's not worth creating a new CSS file just for this

firejake308 (commented):
Not sure I understand what min-width is doing, but it works great on my small computer running Chrome

autoscheduler/frontend/src/components/NavBar/NavBar.css
ryan-conn: It's here because the text for login/logout used to shrink just like the last updated text, but I forgot to remove this line when I changed it

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
ryan-conn: Yeah that was my thought as well, but I kept it as a style since that's how it was originally (which I think is fine until we need multiple of them)

firejake308:
The latest commit looks fine on all screens except iPhone 5/SE

localhost_8000_schedule(iPhone 5_SE) (1)

firejake308 (dismissed):
Works on every screen that I tried except for iPhone 5/SE, which I will handle in #544

PR #543, opened 2021-04-25
Fix default sort direction and add direction to sessions

### LINKED ISSUE #521: Fix default sort order and add sort to sessions
The default sort order is somehow reversed by default - it does default but ascending instead of descending

Also, the sort isn't stored in the course cards sessions, so it doesn't get remembered between sessions

gannonprudhomme:
## Description

Fixes the default sort direction (for the second course card, it works on the first one) and added sort direction to sessions.

## How to test

Add two course cards (one new one) and double check they have the correct default sort direction. Change the sort direction to something else and ensure it saves it in the session

## Related tasks

Closes #521


ryan-conn (approved):
Oops, thought I approved this earlier today but I can confirm it's working

PR #544, opened 2021-04-25
Mobile improvements

### LINKED ISSUE #522: Make the website semi-functional on mobile
Although we are of course not going to make a mobile website, we should at least make the site _semi_ functional on mobile - enough to be able to show people and such. Especially since when we advertise the site (e.g. Reddit, GroupMe's), most people are going to be on their phones when they click on the link.

- Should we just entirely disable availabilities?
  - I’d imagine not but it’s basically impossible to use due to touch input + small screen size
- Doesn’t scroll well for course cards, not sure how to fix
- Should format correctly in landscape

firejake308:
## Description

Makes the website semi-functional on mobile. Not sure if this fixes scrolling on iOS though

## Rationale

Applied a bunch of CSS media queries

## How to test

Open the device toolbar in Chrome DevTools and mess around with a bunch of different devices. Key ones to check include: iPhone 5/SE (the smallest), iPhone X (which a lot of users will have), iPad both horizontal and vertical, because the breakpoint is set between the two), and also regular desktop to make sure nothing broke.

## Screenshots

See #522

## Related tasks

Closes #522


gannonprudhomme:
Rebase this with master real quick so it doesn't fail linting anymore

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
```diff
@@ -44,6 +46,18 @@ export function getLastUpdatedAtText(prev: Date, now = new Date()): string {
...
 
+const useStyles = makeStyles({
+  root: {
+    paddingLeft: 8,
+    whiteSpace: 'nowrap',
+    overflow: 'hidden',
+    textOverflow: 'ellipsis',
+    [appTheme.breakpoints.down('xs')]: {
+      display: 'none',
+    },
```
gannonprudhomme, line 57: the heck is this lmao

firejake308:
Addressed #554. The landing page now looks better in landscape view on mobile, and users can scroll to see the remaining text. See screenshot below. Also made the "Login with Google" text wrap on extremely small screens (<348px), as discussed in #540.

|Before|After|
|--|--|
|localhost_8000_(iPhone X) (1)|localhost_8000_(iPhone X)|

One **important** note is that a double scroll problem was introduced after I merged with master. #471 made it so that the cards will keep a maximum height of 500px on mobile, which means that the section rows will be scrollable, within the overall page that is also scrollable. Personally, I think this behavior is annoying but acceptable, since you have to scroll through all the sections to reach the end regardless. One workaround is to drag in the gray margins, so we could considering beefing up the margins to compensate. If we want to restore the previous behavior (where course cards get expanded all the way on mobile) I could *probably* have the JS `fixHeight()` function check `window.innerWidth` as an alnalog to CSS media queries.

autoscheduler/frontend/src/components/LandingPage/SelectTerm/NavBarSelectTerm.css
```diff
@@ -7,9 +7,16 @@
...
 .menu-paper {
     max-height: 216px !important;
     max-width: 800px !important;
 }
+
+@media (min-width: 768px) {
+    .select-term-button {
+        display: flex !important;
+    }
+}
```
gannonprudhomme, line 22:
```suggestion
}

```

gannonprudhomme (changes requested):
The problem I'm running into mainly is that (in landscape mostly) that it has a scroll bar within some div that contains most of the site, rather than the entire page be scrolling (like how portrait feels on the /schedule view). These are all in Chrome on an iPhone 11. You can see it in this video (had no idea github could show videos):

https://user-images.githubusercontent.com/7295783/117484149-8aa55180-af1b-11eb-9fba-4763efe2797d.mp4

A few other situations that look a bit off:

|Portrait not showing nav bar|Landscape not taking 100% of height|Landscape feels zoomed in a bit too much|
|--|--|--|
|IMG_0831|IMG_0834|IMG_0832|

Also note that you need to reset this with master then cherry pick these commits

firejake308:
Welp, failed to replicate in Chrome on my Android device. Going to see if I can borrow my dad's iPhone tomorrow

firejake308:
I believe that I've fixed the layout of the website on iPhone 6 by adding a `flex-shrink: 0;` CSS rule. I also noticed that the dropdown arrow in the Select Term button was weirdly positioned, so I added a `bottom: 9px;` to position it absolutely. This seemed to improve the layout in browser and on LambdaTest. I'd like for everyone (especially @eelyort) to try out the latest version at today's meeting, but until then, here are some screenshots for y'all until then.

|Before|After|
|--|--|
|iOS-11 3-Firefox|iOS-11 3-Firefox (1)|

firejake308:
Fixed the width of the course column, as seen below:

|Before|After|
|--|--|
|127 0 0 1_8000_schedule(iPhone 6_7_8)|127 0 0 1_8000_schedule(iPhone 6_7_8) (1)|

autoscheduler/frontend/src/components/App/App.tsx
```diff
@@ -9,9 +10,15 @@ import LandingPage from '../LandingPage/LandingPage';
...
+const App: React.FC = function App() {
+  React.useEffect(() => {
+    const takeFullHeight = (): void => { document.body.style.height = `${window.innerHeight}px`; };
+    window.onresize = takeFullHeight;
+    return (): void => { window.removeEventListener('resize', takeFullHeight); };
+  }, []);
+
   return (
-    <div className={styles.appContainer}>
+    <Div100vh className={styles.appContainer}>
```
gannonprudhomme, line 21: is div100vh necessary still? I remember it being mentioned that it didn't do anything but I could be misremembering

autoscheduler/frontend/src/components/LastUpdatedAt.tsx
gannonprudhomme: bump

autoscheduler/frontend/src/components/LandingPage/SelectTerm/NavBarSelectTerm.css
gannonprudhomme: bump2

autoscheduler/frontend/src/components/App/App.tsx
```diff
@@ -9,9 +10,15 @@ import LandingPage from '../LandingPage/LandingPage';
 import SchedulingPage from '../SchedulingPage/SchedulingPage';
 import UnknownRoutePage from '../UnknownRoutePage/UnknownRoutePage';
 
-const App: React.SFC = function App() {
+const App: React.FC = function App() {
+  React.useEffect(() => {
+    const takeFullHeight = (): void => { document.body.style.height = `${window.innerHeight}px`; };
+    window.onresize = takeFullHeight;
+    return (): void => { window.removeEventListener('resize', takeFullHeight); };
+  }, []);
```
gannonprudhomme, line 18: I'd add a comment describing this briefly cause I'm mostly clueless on what this does

gannonprudhomme (changes requested):
Just some minor questions / comments, should be good to go functionality wise

autoscheduler/frontend/src/components/App/App.tsx
firejake308: I just ended up removing it because it doesn't actually help. It was basically a homemade implementation of Div100vh

autoscheduler/frontend/src/components/App/App.tsx
firejake308: I got rid of it and couldn't tell the difference, at least on my Android or Chrome DevTools. I think you're right because I never did fix the bug where you rotate your iPhone and it has the empty space at the bottom, and Div100vh was intended as a possible solution for that

firejake308: It's been a while but I think we decided it was fine as long as it looked good on iPhone 6 in portrait mode

firejake308:
Fixed the middle column width issue

gannonprudhomme (approved):
Confirmed in last meeting that this works as expected!

Thanks a bunch for your help Adel, you killed this as always 🚀

PR #546, opened 2021-04-26
Made select term on landing page golden

firejake308:
## Description

Makes the select term button on the landing page golden.

## Rationale

It's ugly but it attracts the user's attention, which will help compensate for the wall of text that is the landing page

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

None


gannonprudhomme (approved):
Going to take a while to get used to

Issue #548, opened 2021-04-27
Handle remove transitions more gracefully


firejake308:
I found one bug in this, which is that whenever you close a card, the next time you expand one it won't play the animation. This happens on the next time a card is expanded whether there was one expanded when you closed a card or not.

-Ryan
![bug](https://user-images.githubusercontent.com/28655462/116306361-94d08e80-a76a-11eb-8bb2-cc0c60aee68c.gif)

gannonprudhomme:
also the course titles shift when you expand/collapse cards

firejake308:
@gannonprudhomme Just to confirm, are you referring to the fact that there is 3px less space between the delete icon and "Remove" compared to the space between the delete icon and "BIOL 107"?

![transition](https://user-images.githubusercontent.com/10082177/116304377-2094eb80-a768-11eb-9775-9e09b20259d3.gif)

gannonprudhomme:
Nope - previously when you would transition the cards, the course title (`CARC 300 - CARC STUDY AB...`) would shift up and down, but it doesn't seem like it's doing it anymore. Also I can't find my original post on #474 where I showed it so maybe I'm just going crazy

firejake308:
> I found one bug in this, which is that whenever you ~close~ *remove* a card, the next time you expand one it won't play the animation. This happens on the next time a card is expanded whether there was one expanded when you closed a card or not.

@ryan-conn did you mean to say remove instead of close? If so, I can think of a potential fix

ryan-conn:
Yeah that's what I meant, I added a gif to the description with both cases it pops up

PR #553, opened 2021-04-27
Fixed bug where large cards take up extra space

firejake308:
## Description

Setting the height in JavaScript is tricky because it means manually handling a lot of things that CSS usually manages. For cards larger than 500px, I had set the height to take up the maximum available space, but in reality, the height needs to be capped at the height of sectionRows

## How to test

I originally missed this bug because my screen was too small, so make sure your screen is fairly large. Next, open a card that is larger than 500px but can fit within your screen. I used SPAN 301 atfer zooming out my screen to 50%. The card should take up only as much space as it needs, not the whole column.

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Bug originally caused by mistake in #474
Does not address #548, which will be a slightly more involved PR


gannonprudhomme (approved):
Appreciate the quick response with this!

PR #575, opened 2021-05-05
Restored transtion after removing collapsed card

### LINKED ISSUE #548: Handle remove transitions more gracefully

firejake308:
## Description

Fixes a bug from #471 that Ryan reported. Closes #548 

## Rationale

I chose not to fix the lack of a transition when the user removes an expanded card. This is because the transition that plays by default feels off, in that the removed card disappears, the newly expanded card teleports up into its place, and then the new card expands with a transition. During #471, we decided that this was worse than not having any transition at all, which is what led me to disable this specific transition. Instead, when you remove a card, the new card magically pops into place (as it did before #471). I acknowledge that this is not the ideal, which would be to gradually collapse the card to be deleted while expanding the new card, then delete the old card. However, when I attempted to implement this, there was a lag between the collapse transition and the deletion of the card. Here's a GIF of what that looks like:
collapse_then_remove

If y'all think that this GIF is not that bad, I could change to that behavior instead of playing no transition at all.

## How to test

1. Create at least 2 course cards
2. Collapse all of them
3. Remove one
4. Expand one of the remaining cards
5. Assert that the transition played

To test that this doesn't break normal functionality,
1. Create at least 2 course cards
3. Remove the expanded course card
4. Assert that no transition plays and the next course card is now expanded

## Screenshots

|Initial state|Before|After|
|--|--|--|
|Collapsed|before_collapsed|after_collapsed|
|Expanded|after_expanded|<-- same|

## Related tasks

Closes #548

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
@@ -29,9 +29,10 @@ const CourseSelectColumn: React.FC = () => {
   const [wasCourseRemoved, setCourseRemoved] = React.useState(false);
 
   const removeCallback = React.useCallback((id: number) => {
+    const needToDisableTransition = !courseCards[id].collapsed;
     dispatch(removeCourseCard(id));
-    setCourseRemoved(true);
-  }, [dispatch, setCourseRemoved]);
+    if (needToDisableTransition) setCourseRemoved(true);
+  }, [courseCards, dispatch]);
 
   const resetAnimations = React.useCallback(() => {
     setCourseRemoved(false);


gannonprudhomme (approved):
Looks good and works as expected. Great job as always!

firejake308:
@gannonprudhomme I added in some new GIF's showing how it works in different situations. Just wanted to confirm that it looks correct before merging?

gannonprudhomme:
Looks right to me, but I'll leave it up to @ryan-conn for the final approval just to be sure

ryan-conn (approved):
Looks good, I can verify the bug I reported was fixed and I prefer the behavior you chose in the case that a card is expanded at the same time as the deletion (the current behavior that you kept).

PR #580, opened 2021-05-24
Click on meeting goto section

### LINKED ISSUE #467: Make selecting meetings open the respective section in the course card
When you click on a meeting in the Schedule, it should open the respective section in the course card Section customization level

gannonprudhomme:
## Description

I would've made this a draft PR, but I figure it'll just make it stay longer.

Things that y'all requested to be changed but I haven't figured out how to do it:
- `ontransitioned` transition instead of setting a 10ms delay on the transition - doing so never triggered it when I attached it to the `row` class for the course cards (the part that gets `expanded-row` added to it)
- Using refs instead of `getElementById` - how can we possibly do this from the `MeetingCard`? Wouldn't we need to go down the tree, starting from `CourseSelectColumn`?

I also need to figure out how to fix this bug (think this is up to @firejake308 's expertise - any hints would be nice)

clickto-bug

## How to test

Click on a meeting and ensure it opens the respective section

## Screenshots

click-goto

## Related tasks

Closes #467

### DIFF
--- a/autoscheduler/frontend/src/components/GenericSnackbar.tsx
+++ b/autoscheduler/frontend/src/components/GenericSnackbar.tsx
@@ -1,0 +1,40 @@
+import * as React from 'react';
+import { Snackbar, IconButton } from '@material-ui/core';
+import CloseIcon from '@material-ui/icons/Close';
+
+interface GenericSnackbarProps {
+  snackbarMessage: string;
+  setSnackbarMessage: (message: string) => void;
+}
+
+const GenericSnackbar: React.FC<GenericSnackbarProps> = ({
+  snackbarMessage, setSnackbarMessage,
+}) => {
+  const handleSnackbarClose = (_event: any, reason: string): void => {
+    if (reason === 'clickaway') return;
+    setSnackbarMessage('');
+  };
+
+  return (
+    <Snackbar
+      open={!!snackbarMessage}
+      autoHideDuration={5000}
+      message={snackbarMessage}
+      onClose={handleSnackbarClose}
+      onMouseDown={(evt): void => evt.stopPropagation()}
+      action={(
+        <IconButton
+          aria-label="close"
+          onClick={(evt): void => {
+            evt.stopPropagation();
+            setSnackbarMessage('');
+          }}
+        >
+          <CloseIcon fontSize="small" style={{ color: 'white' }} />
+        </IconButton>
+      )}
+    />
+  );
+};
+
+export default GenericSnackbar;
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
@@ -15,7 +15,7 @@ import * as styles from './SectionSelect.css';
 import Section from '../../../../../../types/Section';
 
 export function generateSectionInfoID(section: Section): string {
-  return `${section.id}-0`;
+  return `${section.id}`;
 }
 
 interface SectionInfoProps {
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
@@ -12,6 +12,11 @@ import meetingsForSection from '../../../../../../utils/meetingsForSection';
 import MeetingTypeDisplay from './MeetingType/MeetingTypeDisplay';
 import InstructionalMethodIcon from './InstructionalMethodIcon/InstructionalMethodIcon';
 import * as styles from './SectionSelect.css';
+import Section from '../../../../../../types/Section';
+
+export function generateSectionInfoID(section: Section): string {
+  return `${section.id}-0`;
+}
 
 interface SectionInfoProps {
     sectionData: SectionSelected;
@@ -86,6 +91,7 @@ const SectionInfo: React.FC<SectionInfoProps> = ({
       disableGutters
       button
       component="li"
+      id={generateSectionInfoID(sectionData.section)}
     >
       <ListItemIcon className={styles.myListItemIcon}>
         <Checkbox
--- a/autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
@@ -1,8 +1,11 @@
 import * as React from 'react';
 import { Typography } from '@material-ui/core';
+import { useDispatch } from 'react-redux';
 import * as styles from './MeetingCard.css';
 import Meeting, { MeetingType } from '../../../../types/Meeting';
 import ScheduleCard from '../ScheduleCard/ScheduleCard';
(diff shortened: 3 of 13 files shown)


firejake308:
> I also need to figure out how to fix this bug (think this is up to @firejake308 's expertise - any hints would be nice)

The bug is due to line 99 of `CourseSelectColumn.tsx`. When I was working on #474, I initially set the height of `.section-rows` as the same as its `scrollHeight`, which would give it enough space to accomodate all of its children. However, this caused a useless scrollbar to appear, so I patched it by giving `.section-rows` 1x extra in addition to its `scrollHeight`. The problem is that if you call `fixHeight()` multiple times, then it will add 1px each time, causing the bug that you noticed in this PR. I'm going to try to write tests for both of these bugs to make sure that we don't accidentally break one or the other.

firejake308:
Okay, but before I make tests, I have to be able to reproduce the bug, and I can't seem to get it to show up with my current process. I know it needs to be on a card that has enough room to render all sections (i.e. one that doesn't scroll), but I have tried clicking on the same meeting card repeatedly as well as alternating between classes, but neither seems to make the sections grow. Any advice @gannonprudhomme ?

cant_reproduce

firejake308:
Pressed wrong button, my bad

gannonprudhomme:
This is ready to review now, since I'm not going to be doing the `ontransitioned` event

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
```diff
@@ -12,6 +12,11 @@ import meetingsForSection from '../../../../../../utils/meetingsForSection';
 import MeetingTypeDisplay from './MeetingType/MeetingTypeDisplay';
 import InstructionalMethodIcon from './InstructionalMethodIcon/InstructionalMethodIcon';
 import * as styles from './SectionSelect.css';
+import Section from '../../../../../../types/Section';
+
+export function generateSectionInfoID(section: Section): string {
+  return `${section.id}-0`;
+}
```
ryan-conn, line 19:
```suggestion
export function generateSectionInfoID(section: Section): string {
  return `${section.id}-0`;
}
```
Is there a reason you add `-0` to the end rather than doing something like `section-info-${section.id}`? I'd prefer we keep ids helpful, the current names seem arbitrary

autoscheduler/frontend/src/components/GenericSnackbar.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { Snackbar, IconButton } from '@material-ui/core';
+import CloseIcon from '@material-ui/icons/Close';
+
+interface GenericSnackbarProps {
+  snackbarMessage: string;
+  setSnackbarMessage: (message: string) => void;
+}
+
+const GenericSnackbar: React.FC<GenericSnackbarProps> = ({
```
ryan-conn, line 10: Why did you choose to have multiple snackbars rather than use a single one and update the message using redux? I doubt anyone will actually encounter this so it's not a huge deal (and idk if it's worth bothering to fix unless we end up adding more snackbars in the future), but if you cause an error with generating schedules and click a meeting card with no corresponding course card, it will make both snackbars appear at once:
image

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
```diff
@@ -23,41 +28,75 @@ const MeetingCard: React.FC<MeetingCardProps> = ({
...
   const [isBig, setIsBig] = React.useState(true);
+  const [snackbarMessage, setSnackbarMessage] = React.useState('');
+  const dispatch = useDispatch();
 
   // hide meeting type if the card is small
   const handleResize = React.useCallback((newVal: boolean): void => {
     setIsBig(newVal);
   }, []);
 
+  const onClick = (): void => {
```
ryan-conn, line 39: The proper name for react event handler functions is `handleX` rather than `onX` (you can see in the react documentation and our other event handlers): you can leave the prop as `onClick`, but when setting its value here it should be `handleClick`.
```suggestion
  const handleClick = (): void => {
```

autoscheduler/frontend/src/components/SchedulingPage/Schedule/ScheduleCard/ScheduleCard.tsx
```diff
@@ -95,8 +98,14 @@ const ScheduleCard: React.FC<ScheduleCardProps> = ({
...
       onMouseLeave={(): void => setHovered(false)}
       onMouseDown={(e): void => {
+        // Don't allow dragging in availability blocks
         if (onDragHandleDown && e.target === e.currentTarget) e.stopPropagation();
+        // Ignore mouse down if we have an onClick handler (for MeetingCard -> CourseCard jumping)
+        // which prevents creating an availbility on top of a meeting
+        if (onClick) e.stopPropagation();
       }}
+      onClick={onClick}
+      onKeyPress={onClick}
```
ryan-conn, line 108: Why is this also an `onKeyPress` event? It doesn't seem to do anything and if I'm reading this right, the intention is to go to the section if you press a button over a meeting card (unless I'm reading it wrong)?

autoscheduler/frontend/src/redux/actions/termData.ts
```diff
@@ -118,7 +119,12 @@ export interface UpdateSortTypeAction {
   sortIsDescending: boolean;
 }
 
+export interface ScrollSectionIntoViewAction {
+  type: 'SCROLL_SECTION_INTO_VIEW';
+  section: Section;
+}
+
```
ryan-conn, line 126: This action actually just expands a card if it exists, or shows an error if it doesn't and sets it to section view, correct? The current name is misleading since no scrolling actually occurs here

ryan-conn (changes requested):

autoscheduler/frontend/src/components/GenericSnackbar.tsx
gannonprudhomme: I thought we had discussed about it but maybe we didn't - but it was just much simpler to make them both use a Generic snackbar than doing it through Redux.

I honestly hadn't even considered this had a possibility of happening so good catch, we can definitely switch it to doing redux if we think it's necessary

ryan-conn: Nah, like I said I don't think anyone will ever actually do that so it's not worth the effort

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionInfo.tsx
gannonprudhomme: I had originally put it there as a dummy, as I intended to have that there and count how many times each section ID occurred so we could have two course cards of the same type. I realized that was difficult to do and not necessary, and this is just left over.

autoscheduler/frontend/src/redux/actions/termData.ts
gannonprudhomme: Yeah this is a great point, was really just left over from my initial plans of how this would work

autoscheduler/frontend/src/components/SchedulingPage/Schedule/ScheduleCard/ScheduleCard.tsx
gannonprudhomme: I did it because otherwise you get `jsx-a11y/click-events-have-key-events`: `Visible, non-interactive elements with click handlers must have at least one keyboard listener ` and just kind of assumed it would work tbh

gannonprudhomme:
Other than the `onKeyPress` thing did all requested changes

ryan-conn (approved):
Looks good after addressing my changes

Issue #572, opened 2021-05-03
Use no arrow for tooltip on GPA in GradeDist


gannonprudhomme:
![image.png](https://images.zenhubusercontent.com/5da358de33f74500019c3b7f/009bccee-2e6e-44e6-9074-ffe8ce33474f)

Either need to change the arrow direction or just make it use the no arrow Tooltip (might as well for consistencies sake)

firejake308:
The positioning is also funky, in that it covers up the GPA text. Will removing the arrow fix that?

gannonprudhomme:
> The positioning is also funky, in that it covers up the GPA text. Will removing the arrow fix that?

Was figuring it would - I assumed the arrow was forcing it to remain above it, but it clearly can't

ryan-conn:
Do together with #571

PR #582, opened 2021-05-25
GradeDist fixes

### LINKED ISSUE #571: GPA tooltip z-index smaller than instructor bar in SectionSelect

ryan-conn:
## Description
Fixes #571 and #572

## Rationale
I fixed #571 by adding `zIndex` as a parameter of `<ProfessorGroup />`: due to [how stacking contexts work in CSS](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), since each professor name header is a `sticky` element with identical z indices the tooltips will render under sequential headers.
There are only two ways to fix this:
- Remove tooltips from the flow of the page so that they can be placed in a stacking group with higher priority (e.g. use portal again)
- Give headers descending z indices so that the tooltips won't show below sequential headers
I opted for the second choice given we've determined we don't want to ơ̴̬̩̈́̆̿̓p̸̘̙̎̑͌ȩ̸̛̳̉̊͆ṅ̶̢̪̻̼̰̊̀͘͝ ̷͚̐̐̆̅͝t̸̗̪̥̘̹͑̈́h̷̛͙̭̐e̴̻͖̓̎̈́̓͝ ̶͕̕ṕ̸͚̍͗o̷̡̰͂͌̈́ŕ̶̛͕̞̭̿̂̾t̵̨̛͉̬͔̀a̸̬̖͓̩̦̾́ḽ̶́͊̇ ̶͖͔̬̩̄̄͛̈͒a̸̘͖̝̋̊̄̈́͒g̸̗̭̣̉á̵͈̼̖͖̙̿̒̾͒ȉ̸̢͚̽͐n̷͍̅͛̍́̃. Note that if there was a case where tooltips could be positioned over previous section headers the problem would still happen in that case, but I don't think this is possible. As such I think this is an adequate fix for the issue.

## How to test
Hover tooltips and notice the lack of arrow as well as tooltips now showing correctly over professor name headers

## Screenshots
|Before|After|
|--|--|
| image | image |
## Related tasks

Closes #571 and #572.

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/GradeDist/GradeDist.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/GradeDist/GradeDist.tsx
@@ -62,7 +62,6 @@ const GradeDist: React.FC<GradeDistProps> = ({ grades }) => {
     <div className={styles.gradesContainer}>
       <Tooltip
         title={makeFromTotalSectionsTooltipText(grades.count)}
-        arrow
         PopperProps={{
           disablePortal: true,
         }}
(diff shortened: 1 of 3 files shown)


gannonprudhomme (approved):
Can't say I 100% understand CSS stacking, but can confirm that this works as expected

PR #583, opened 2021-05-25
Frontend/add filters full

### LINKED ISSUE #364: Add filters to SectionSelect
We should add filters to section selects, such as honors-only or online-only. Should also be able to filter out sections that are full, or one that intersect with your availabilities.

eelyort:
## Description

Added filters, allowing the user to filter through the sections and find the one they want easier. In addition, removed basic vs. section select as the filters basically replaces basic section select.

## How to test

Should be immediately visible in the course cards. In particular Fall 2021 PSYC 107 is good since it has all three filters.

## Screenshots

5-25-2021-demo

## Related tasks

Tests for filters need to be done. One failing test was skipped.

Closes #364


eelyort:
Moved to new pull request

PR #585, opened 2021-05-27
Change calendar width

### LINKED ISSUE #568: Increase size of course select column
Probably the most frequent suggestion we've been getting is that the course select column doesn't use enough space. As such, we should shrink the space the schedule gets and increase the space the course select column gets (since it's the most important part, after all)

From a bit of preliminary testing we released that around ~750px for max-width of the schedule probably is good, but we could even go as low as 550px (from what I can tell).

### LINKED ISSUE #570: Increase size of grade distributions in course select column
The grade distributions max-width is 10rem - not sure how we want to do it, but we should definitely increase it's ability to grow, as there's no reason it should be this small:

|1440p|1080p|
|--|--|
|||

firejake308:
## Description

Changes how space is distributed between the 3 columns on the SchedulingPage. Course cards will now always render on one line, with no wrapping. This is achieved by setting `min-width: 400px`. The remaining space is split in a 1:1:2 ratio between course column, the middle column, and the schedule, respectively. However, the middle column has a maximum width of 550px, after which point the remaining space is split between the course card column and the schedule in a 1:2 ratio. Neither the first nor the last column has a maximum width.

This PR also increases the `max-width` of grade distributions from `10rem` to `20rem`. 8px of margin have also been added between the instructor name and grade container in order to make this look good on small screens.

## Rationale

Earlier, we had discussed possibly setting a maximum width for course cards to prevent them from getting too stretched out. However, the cards look reasonable to me on all screen sizes (720p, 1080p, 1440p, and 4k)

## How to test

### 720p:
image

### 1080p:
127 0 0 1_8000_schedule (4)

### 1440p:
127 0 0 1_8000_schedule (3)

### 4k:
127 0 0 1_8000_schedule (2)

## Related tasks

Closes #568 
Closes #570

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/GradeDist/GradeDist.css
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/GradeDist/GradeDist.css
@@ -2,8 +2,7 @@
     display: flex;
     align-items: center;
     flex-grow: 1;
-
-    max-width: 10rem;
+    max-width: 20rem;
 }
 
 .grades-dist {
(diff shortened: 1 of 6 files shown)


ryan-conn:
This all looks good to me anywhere from 720p to 4k, I'll leave it up to @gannonprudhomme to approve to make sure he agrees.

I'm curious how this would work with #544 considering they both change the page CSS: it seems like they should work fine together considering mobile improvements mostly performs media queries, but it's worth considering how you want to handle merging both of these

gannonprudhomme (approved):

PR #586, opened 2021-05-30
Fix scrape courses --year parameter, check terms have data before saving

### LINKED ISSUE #547: scrape_courses --year option doesn't work correctly
It broke whenever we did the last updated indicator (#464). When you do `--year` it does all of the terms, so it's like you passed in nothing

ryan-conn:
## Description
The initial plan was for this PR to just fix the `--year` parameter for `scrape_courses`, but trying to do so I noticed that terms were being saved that didn't have any courses. This made me realize we could simplify the logic by ignoring these extra terms instead of the band-aid fix of manually checking the terms to fix (which could break if TAMU hasn't posted classes for a term when we scrape, leading to a term that shows up but has no courses).

## Rationale
This could really be a 2-line PR if we just wanted to restore the previous behavior, but I think the simplification of our codebase and more robust term checking is worth the extra effort.

## How to test
Try scraping the current year as well as next year, and notice that only terms with data are being saved. I added tests for terms without courses being saved and moved existing get_all_terms tests to get_recent_terms tests with its expected behavior, since that function didn't have as many tests made for it.

## Related tasks
Closes #547.

### DIFF
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -325,7 +325,7 @@ def save_models(instructors: List[Instructor], sections: List[Section], # pylint
 
     print(f"Saved all in {elapsed_time:.2f} seconds")
 
-def save_terms(terms, options):
+def save_terms(terms, courses, options):
     """ Creates terms objects to save """
 
     start = time.time()
@@ -340,10 +340,7 @@ def save_terms(terms, options):
             queryset = Term.objects.all()
 
         # Only save terms that actually have courses so that users don't see empty terms
-        terms_with_courses = set(
-            course['term'] for course in
-            Course.objects.distinct('term').filter(term__in=terms).values('term')
-        )
+        terms_with_courses = set(course.term for course in courses)
         terms_to_save = [
             Term(code=term, last_updated=now) for term in terms
             if term in terms_with_courses
@@ -390,6 +387,6 @@ class Command(base.BaseCommand):
 
         instructors, sections, meetings, courses = get_course_data(depts_terms)
         save_models(instructors, sections, meetings, courses, terms, options)
-        save_terms(terms, options)
+        save_terms(terms, courses, options)
 
         print(f"Finished scraping in {time.time() - start_all:.2f} seconds")
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -383,7 +383,7 @@ class Command(base.BaseCommand):
         if options['term']:
             terms = [options['term']]
         elif options['year']:
-            terms = get_all_terms(options['year'])
+            terms = get_all_terms(year=options['year'])
         elif options['recent']:
             terms = get_recent_terms()
         else:
--- a/autoscheduler/scraper/management/commands/scrape_courses.py
+++ b/autoscheduler/scraper/management/commands/scrape_courses.py
@@ -353,7 +353,7 @@ def save_terms(terms, options):
         queryset.delete()
         Term.objects.bulk_create(terms_to_save)
 
-        term_len = len(terms) if terms else 1
+        term_len = len(terms_to_save) if terms else 1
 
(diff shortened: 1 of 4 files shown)


autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -331,55 +330,64 @@ def save_terms(terms, options):
...
+            if term in terms_with_courses
+        ]
+        print(terms_to_save)
+
         queryset.delete()
-        Term.objects.bulk_create(term_models)
+        Term.objects.bulk_create(terms_to_save)
 
-        term_len = len(terms) if terms else 1
+        term_len = len(terms_to_save) if terms else 1
```
gannonprudhomme, line 356:
```suggestion
        term_len = len(terms_to_save) if terms_to_save else 1
```
is this even stilll necessary? I forgot my logic when I originally made it lol

autoscheduler/scraper/management/commands/utils/scraper_utils.py
```diff
@@ -2,51 +2,22 @@
 from typing import List
 from itertools import product
 
-def get_all_terms(year: int = -1, now=datetime.now()) -> List[str]:
+def get_all_terms(*_, year: int = None, now=datetime.now()) -> List[str]:
```
gannonprudhomme, line 5: image

mfw when you reduce my 40 line function to 8 lines

gannonprudhomme (changes requested):

autoscheduler/scraper/management/commands/scrape_courses.py
ryan-conn: Nope, I noticed that too and decided not to change it since it wasn't necessary, but I'll still make the change since you pointed out

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -331,55 +330,64 @@ def save_terms(terms, options):
...
+        # Determine terms to resave
         if options['year'] or options['recent'] or options['term']:
             queryset = Term.objects.filter(code__in=terms)
         else: # If no options are provided, then we're scraping all terms
             queryset = Term.objects.all()
 
+        # Only save terms that actually have courses so that users don't see empty terms
+        terms_with_courses = set(
+            course['term'] for course in
+            Course.objects.distinct('term').filter(term__in=terms).values('term')
```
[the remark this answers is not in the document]
ryan-conn: I haven't actually timed it, but neither of those operations should take more than 10-20ms so it doesn't really matter which way we do it— I did it this way because I thought it was pretty clear how the terms with courses are being determined, and we don't have to pass in the scraped courses

autoscheduler/scraper/management/commands/scrape_courses.py
```diff
@@ -331,55 +330,64 @@ def save_terms(terms, options):
...
     """ Gets course information from banner and adds it to the database """
 
     def add_arguments(self, parser):
-        parser.add_argument('--term', '-t', type=str,
-                            help="A valid term code, such as 201931")
-        parser.add_argument('--year', '-y', type=int,
-                            help="A year to scrape all courses for, such as 2019")
-        parser.add_argument('--recent', '-r', action='store_true',
-                            help="Scrapes the most recent semester(s) for all locations")
+        time_args = parser.add_mutually_exclusive_group()
```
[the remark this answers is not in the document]
ryan-conn: I didn't either until I looked at the documentation because I figured there must be an easier way to handle this use case

autoscheduler/scraper/management/commands/utils/scraper_utils.py
ryan-conn: To be fair it's because I made it so the function doesn't need to handle edge cases anymore since terms without courses are ignored, but it's always nice when something ends up being simplified in a PR

autoscheduler/scraper/management/commands/scrape_courses.py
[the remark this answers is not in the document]
gannonprudhomme: We talked about it last week, but make this change to search through the list and this will be good to go

gannonprudhomme (approved):
🚀

Issue #566, opened 2021-04-29
Add ability to expand schedule


gannonprudhomme:
One of the most frequent requests is to expand the size of the schedule (once people are done selecting their courses, typically)

gannonprudhomme:
Updated this to be a comprehensive/screenshot schedule view rather than full screen, since simply making it wider really isn't all that helpful.

PR #587, opened 2021-05-30
Added fullscreen schedule view

gannonprudhomme:
## Description

This adds the ability to make the schedule fullscreen. Screen-shottable/save-as-image schedules coming later. I'm considered of the performance for this tbh - it should be really snappy but it is not and I'm not exactly sure why

## How to test

Create different schedules and fullscreen them, and ensure they look right. Always ensure there is at least 8 hours on the schedule at a time (per the `getFirstAndLastHour` function).

CSCE 482 is a good test case, as it has a 15 minute section in them.

## Screenshots

fullscreen-schedule-demo

## Related tasks

I converted #566 to be just screenshot a schedule, so no issue for this one


gannonprudhomme:
Reminder to myself for rounding the corners of the bottom fullscreen button bar

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
```diff
@@ -3,32 +3,157 @@ import { Typography } from '@material-ui/core';
...
+    if (heightDiff > 30) {
+      setCardSize(MeetingCardSizes.large);
+    } else if (heightDiff > -15) { // Just enough for two rows of text
+      setCardSize(MeetingCardSizes.medium);
+    } else {
+      setCardSize(MeetingCardSizes.small);
+    }
   }, []);
 
+  const fullscreenLargeCard = (
```
gannonprudhomme, line 49: Should I declare these in the switch so it doesn't construct these until we actually need them, since we're making 3 always but only ever need 1 per card? Or does it not matter since it's not rendered/executed until we actually render them

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.css
```diff
@@ -14,6 +14,20 @@
 .schedule-container {
     flex-grow: 1;
     flex-basis: 0;
+    height: calc(100% - 8px);
```
ryan-conn, line 17:
```suggestion
    height: 100%;
```
The current calculation has too much bottom padding
| 100% | 100% - 8px |
|-|-|
| image | image |

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -48,6 +57,23 @@ const SchedulingPage: React.FC<SchedulingPageProps> = ({
       </div>
       <div className={styles.scheduleContainer}>
         <Schedule />
+        <div
+          className={styles.fullscreenButtonContainer}
+          style={{ backgroundColor: theme.palette.primary.main }}
+        >
+          <ThemeProvider theme={whiteButtonTheme}>
+            <Typography color="primary" className={styles.totalHoursText}>
+              Total Hours: 15
```
ryan-conn, line 66: Might want to dummy this out before merging, but leave the structure so it can be readded later with minimal effort

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -36,8 +44,9 @@ const SchedulingPage: React.FC<SchedulingPageProps> = ({
   }, [dispatch, termCurr]);
 
   return (
-    <div className={styles.pageContainer}>
-      <div className={styles.leftContainer}>
+    <div className={styles.pageContainer} style={fullscreen ? { justifyContent: 'flex-end' } : null}>
+      {/* Hide the left & middle column if we're fullscreen */}
+      <div className={styles.leftContainer} style={fullscreen ? { display: 'none' } : null}>
```
ryan-conn, line 49: I don't like using style here, use additional classes instead (same applies to NavBar)

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -48,6 +57,23 @@ const SchedulingPage: React.FC<SchedulingPageProps> = ({
       </div>
       <div className={styles.scheduleContainer}>
         <Schedule />
+        <div
+          className={styles.fullscreenButtonContainer}
+          style={{ backgroundColor: theme.palette.primary.main }}
+        >
```
ryan-conn, line 63: imo this should use the `GenericCard` component and add an option to make it maroon (a `color: 'primary' | secondary'` prop?), it should also have rounded edges like all the other cards on the page

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.css
```diff
@@ -1,11 +1,12 @@
 @value v-time-label-margin: 48px;
+@value fullscreen-row-height: 48px;
 
 .calendar-container {
     display: flex;
     flex-direction: column;
     padding: 8px;
     font-family: "Roboto", "Helvetica", "Arial", sans-serif;
-    height: calc(100% - 24px);
+    height: calc(100% - 24px - fullscreen-row-height);
```
ryan-conn, line 9: What's preventing you from making the parent element flex rather than manually subtracting the fullscreen row height?

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
ryan-conn: Yeah do it in the switch

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
```diff
@@ -3,32 +3,157 @@ import { Typography } from '@material-ui/core';
...
+      fullscreenCard = fullscreenMediumCard;
+      break;
+    case MeetingCardSizes.small:
+      fullscreenCard = fullscreenSmallCard;
+      break;
+    default:
+      break;
+  }
+
+  const normal = (
```
ryan-conn, line 131: Instead of having `normal` and `fullscreenCard` variables, just have a single `card` variable that determines what card type to use

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
```diff
@@ -3,32 +3,157 @@ import { Typography } from '@material-ui/core';
 import * as styles from './MeetingCard.css';
 import Meeting, { MeetingType } from '../../../../types/Meeting';
 import ScheduleCard from '../ScheduleCard/ScheduleCard';
+import meetingTimeText from '../../../../utils/meetingTimeText';
+import { meetingBuildingWithRoom } from '../../../../utils/meetingBuilding';
+
+enum MeetingCardSizes {
```
ryan-conn, line 9: Small but typescript enum names are singular, not plural since it makes enum member access look more natural (sizes.large vs. size.large)
```suggestion
enum MeetingCardSize {
```

ryan-conn (changes requested):
This needs a test checking that room scraping works correctly, afaik there isn't one

I also noticed that the fullscreen page breaks when the height is low:
image

ryan-conn:
Make the following diff to fix the issue with schedule buttons we talked about last meeting:
image

The issue is that `useDimensions` only updates on window resize (as javascript doesn't have an element-level resize event), so opening the console will cause `setDimensions` to be called when it's not being rendered due to the `display: none` around the component.

I should note that apparently [resize observers](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) should work, but I can't get them to return the correct `top` and `left` values, nor find a node package that actually uses them and returns the `top`/`left` values with over 100 downloads, so I'll call this solution good enough for now.

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
gannonprudhomme: I will do it now, but just want to note that there is a chance that I have to re-do this change in save schedules as image, as at the moment I render both of them at once and `display: none` one of them (and `display: none` the normal one when we're saving it as an image). Only doing it now as there's also a chance that I don't do it this way

autoscheduler/scraper/tests/api_tests.py
```diff
@@ -107,6 +107,9 @@ def setUpTestData(cls):
         # Convert INSTRUCTIONAL_METHOD_CHOICES to dict
         cls.instructional_methods = dict(Section.INSTRUCTIONAL_METHOD_CHOICES)
 
+    def setUp(self):
+        self.maxDiff = None
+
```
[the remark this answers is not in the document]
gannonprudhomme: Yeah def just a debug thing I left in

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.css
gannonprudhomme: Am I missing something? I did this intentionally - this bottom padding makes it line up with the bottom padding of the middle column and the course card column.

|100%|100% - 8 px|
|--|--|
|chrome_7VK7tqDy9M|chrome_pjxp3LkOqc|

autoscheduler/scraper/models/section.py
```diff
@@ -67,6 +67,7 @@ class Meeting(models.Model):
     """
     id = models.BigIntegerField(primary_key=True) # id is primary key in scraped data
     building = models.CharField(max_length=6, null=True)
+    room = models.CharField(max_length=6, null=True)
```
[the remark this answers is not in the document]
gannonprudhomme: Great catch - `RDR` (Rudder tower probs) has a `THEATRE` room that was seemingly introduced in 202031

autoscheduler/frontend/src/utils/meetingBuilding.ts
```diff
@@ -1,5 +1,12 @@
...
 export default function meetingBuilding(meeting: Meeting): string {
-  return meeting.building || 'ONLINE';
+  return meeting.building ? meeting.building : 'ONLINE';
+}
+
+export function meetingBuildingWithRoom(meeting: Meeting): string {
+  // A few sections have ONLINE as both the building * room, so return ONLINE for these instead
+  return meeting.building && meeting.building !== 'ONLINE'
+    ? `${meeting.building} ${meeting.room}`
+    : 'ONLINE';
```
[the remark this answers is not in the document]
gannonprudhomme: Yes - in fact there are 1703 sections (2437 meetings)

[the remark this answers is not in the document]
gannonprudhomme: Okay yeah scraped all terms and queried again - 527 sections and 579 meetings, so significantly less but they still exist

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
gannonprudhomme: What did you have in mind for dummying this out? I was just going to leave this here and set the color to be transparent (`display: none` makes the fullscreen button shift left into the total hours text space b/c space-between)

gannonprudhomme:
> I also noticed that the fullscreen display breaks when the height is low:
> image

lmao did you notice that the time range is fucked on the side?

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.css
gannonprudhomme: Good call - I honestly thought I had done this and it didn't work but I guess I did something wrong the first time.

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
gannonprudhomme: I wouldn't be opposed to making a `GenericHeader` if you'd like, but I don't think `GenericCard` would be appropriate since it won't actually have any children (and thus won't be a card). I also have to edit the `GenericCard` a bit to get it to work how we want it to be, b/c we have to change the border radius and height of its internal `header` div, which isn't possible currently (basically just also need to add the ability to change the styles of the header, currently it's just for the card content). Not that I mind doing it ofc, I just think it's going beyond the functionality that a `GenericCard` needs in most other cases and as such it should be something different probably.

I do agree it should have rounded edges, even left a comment to remind myself but I forgot lol. I'll add them now and we can follow up about making this generic later if you have some more ideas on it

gannonprudhomme:
Implemented all comments except for the `GenericCard` one (also haven't "dummied" out the total hours anymore than what I had other than hiding it). Also haven't figured out how to handle the shrinking full screen schedule effectively, as it seems that one of the components of the schedule have a min height that is preventing it from shrinking anymore (and thus it shrinks the full screen button container). Going to ask adel about that one

gannonprudhomme:
Also during review try to confirm that I didn't undo changes from #594

autoscheduler/frontend/src/utils/timeUtil.ts
```diff
@@ -36,3 +38,55 @@ export const timeToEvent = (h: number, m: number, offset = 0, clientHeight = 100
...
+
+  // Use earliest & latest if we're fullscreen
+  let first = FIRST_HOUR;
+  let last = LAST_HOUR;
+  if (fullscreen && schedule.length > 0) {
+    first = earliest;
+    last = latest + 1;
+  }
+
+  return { first, last };
```
ryan-conn, line 91: Since none of the logic above this gets applied unless `fullscreen` is set, I would just return early by checking if it's set at the beginning to make it more clear how these numbers are set

ryan-conn (changes requested):
Functionality looks good, a couple nitpicks to make the code a bit higher quality

I'd also like if you could add/modify a test or two to make sure rooms are being correctly set, it seems like something easy to overlook when we're making changes

autoscheduler/frontend/src/components/SchedulingPage/Schedule/MeetingCard/MeetingCard.tsx
```diff
@@ -5,44 +5,173 @@ import * as styles from './MeetingCard.css';
...
+                {meetingTimeText(meeting)}
+              </Typography>
+            </Typography>
+          </>
+        );
+        break;
+      default:
+        break;
+    }
+  } else { // default fullscreen card
```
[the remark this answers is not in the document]
gannonprudhomme: Yep! Meant default meeting card

ryan-conn (approved):
Looks good now that the changes from last time have been made

PR #523, opened 2021-03-20
Filters for section select

firejake308:
## Description

If a user wants to select indidivual sections but narrow them down, they currently have no option besides scrolling through every section and manually checking which ones meet their criteria. This PR adds filters to SectionSelect, so that users can limit to/exclude honors, remote, or asynchronous sections.

Since we're close to launch and we want this feature by then, I've made a draft PR to start collecting feedback, even though the feature is not yet complete. TODO:

- ~Add UI~
- Add filtering functionality
- Save to sessions
- Fix transition heights
- Fix tests

## Rationale

I added a bottom divider to help distinguish the filters part from the rest of the SectionSelect

## How to test

Actual functionality is still TODO, so far only the UI is finished

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related tasks

Closes #364

PR #590, opened 2021-05-31
Fixed bug with infinitely growing cards

### LINKED ISSUE #588: Fix course card height increasing by 1px upon re-render
E.g.:

Also happens if you enable/disable the course card, and upon initial render/refresh

firejake308:
## Description

This bug was originally discovered in #580, read the comments there for more context.

## Rationale

I elected to address both of the issues by adding 2px of padding-bottom to `.section-rows`, since we were going to do it anway in #523 

## How to test

Bug 1 - Disable, then enable a course card repeatedly. The card should not grow. There is also a Jest test to ensure that this bug does not relapse.

Bug 2 - Open SPAN 301 in Chrome (not Firefox), and make sure your window is big enough to accommodate all sections without scrolling. If not, zoom out, then close and open the course card. There should be no scroll bar in the list of sections.

## Screenshots

|Before|After|
|--|--|
|image|image|

## Related Tasks

Closes #588

### DIFF
--- a/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
+++ b/autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
@@ -82,6 +82,7 @@ const CourseSelectCard: React.FC<CourseSelectCardProps> = ({
         <Switch
           aria-label="Disable"
           checked={!disabled}
+          inputProps={{ 'aria-checked': !disabled }}
           onClick={(evt): void => {
             dispatch(updateCourseCard(id, { disabled: !disabled }));
             evt.stopPropagation();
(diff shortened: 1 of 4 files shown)


gannonprudhomme:
You have my commits from #580 in here - you'll need to hard reset and cherry-pick these commits

gannonprudhomme (changes requested):
shift-bug

this does still seem to be occurring in Chrome unfortunately :( Also occurs in Firefox for me, but it doesn't do it on the initial render like Chrome does

firejake308:
image

Now that you mention it, I am also seeing the bug as well. Time to debug, I guess

firejake308:
Okay, looks like the problem was that on each render, I was setting the element `height` to be equal to the current `scrollHeight`, which actually included the element height + padding. By changing the padding to margin, it seems to resolve both bugs on my computer.

gannonprudhomme:
Note that you still have my commits from #580 in here - otherwise I believe this is good to go

firejake308:
Rebased with master

gannonprudhomme (approved):

PR #591, opened 2021-06-05
Term select loading indicator

### LINKED ISSUE #375: Add loading indicator for SelectTerm
Occasionally (especially when you navigate to the website for the first time in a bit and it's starting up slowly), the `api/terms` route takes a bit too long to resolve and as such, the `SelectTerm` drop-downs are empty. We should add a loading indicator that prevents them from opening the `SelectTerm` drop-down until `api/terms` has actually resolved.

In actuality `api/terms` should basically always be instant, since the response only changes like once a semester, but since we don't have caching there isn't a super simple way of making this faster.

ryan-conn:
## Description
Adds a loading indicator to `<SelectTerm />` that shows up after loading for half a second

## Rationale
We said in the meeting last week that it would make the website seem unresponsive if the loading indicator showed up immediately, so I added a 500ms delay before starting to display the loading indicator: before that, the component displays as it did before.

## How to test
Make the following change to make fetching terms take long enough to see the loading indicator locally:
image

## Screenshots
loading

## Related tasks
Closes #375.

### DIFF
--- a/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.css
+++ b/autoscheduler/frontend/src/components/LandingPage/SelectTerm/SelectTerm.css
@@ -14,3 +14,9 @@
     width: 55% !important;
     max-width: 800px !important;
 }
+
+.loading-indicator {
+    display: flex;
+    justify-content: center;
+    min-width: 200px;
+}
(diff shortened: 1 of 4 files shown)


gannonprudhomme (approved):
Wait until the meeting so everyone can confirm this is how we want it - but this looks good to me

PR #592, opened 2021-06-06
Frontend/add filters full 3

### LINKED ISSUE #364: Add filters to SectionSelect

eelyort:
## Description

Added filters, allowing the user to filter through the sections and find the one they want easier. In addition, removed basic vs. section select as the filters basically replaces basic section select.

## How to test

Should be immediately visible in the course cards. In particular Fall 2021 PSYC 107 is good since it has all three filters.

## Screenshots

5-25-2021-demo

## Related tasks

Tests for filters need to be done. One failing test was skipped.

Closes #364


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
```diff
@@ -4,18 +4,27 @@ import RemoveIcon from '@material-ui/icons/Delete';
...
 
+const Collapse = withStyles({
+  container: {
+    display: 'flex',
+  },
+  wrapper: {
+    width: '100%',
+  },
+})(CollapseBase);
+
```
ryan-conn, line 27: I noticed that this PR breaks the transition for course cards (cards opening "snap" open all the way open from closed at the very end of the animation). This branch is on the left, production is on the right:
animationbroken.

Removing this override for the `Collapse` component seemed to fix this issue for me, why is it here considering it just seems to break the animation?

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -73,43 +157,87 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
...
    */
+  let allSelected = true;
   const makeList = (): JSX.Element[] => {
+    /**
+     * Function to filter based off of selected filters.
+     * We need to use this here and in professor group,
+     * since professor group only gets a start and end index.
+     */
+    const filterSections = (sectionData: SectionSelected): boolean => {
+      const toBool = (filter: SectionFilter | undefined, val: boolean): boolean => {
```
ryan-conn, line 174:
```suggestion
    const shouldIncludeSection = (sectionData: SectionSelected): boolean => {
      // Whether the section should be included based on a SectionFilter and its value
      const filterOnVal = (filter: SectionFilter | undefined, val: boolean): boolean => {
```
I don't like the name `filterSections`, since it only checks a single section (whereas the name makes it seem like it should filter all of them at once), try using the new name I provided (the same applies to the `ProfessorGroup component's property with the same name).

Similarly `toBool` isn't a very good name: it's hard to tell what it does: changing the name and adding a comment will help with readability

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -73,43 +157,87 @@ const SectionSelect: React.FC<SectionSelectProps> = ({ id }): JSX.Element => {
...
+
     let lastProf: string = null;
     let lastHonors = false;
     let currProfGroupStart = 0;
-    return sections.map((sectionData, secIdx) => {
-      const firstInProfGroup = lastProf !== sectionData.section.instructor.name
+    // since we will be filtering, we need to store the index somewhere
+    return sections
+      .map((sectionData, secIdx) => ({ sectionData, secIdx }))
+      .filter(({ sectionData }) => filterSections(sectionData)).map(({ sectionData, secIdx }) => {
```
ryan-conn, line 195:
```suggestion
    return sections
      // Remember original index of sections so that ProfessorGroup uses the correct indices
      .map((sectionData, secIdx) => ({ sectionData, secIdx }))
      .filter(({ sectionData }) => filterSections(sectionData))
      .map(({ sectionData, secIdx }) => {
```
I tried rewriting this because it looked messy before realizing there was actually a reason you kept the indices, make sure to add comments for minor stuff like this so that other people can understand what your code is doing better. Also make sure to keep your indentation consistent to help with readability.

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -1,36 +1,110 @@
 import * as React from 'react';
 import { useDispatch, useSelector } from 'react-redux';
 import {
-  List, Typography, Checkbox, Button, Menu, MenuItem, IconButton, Tooltip,
+  List, Typography, Checkbox, Button, Menu, MenuItem, IconButton,
+  Tooltip, ExpansionPanel as ExpansionPanelBase, ExpansionPanelSummary as ExpansionPanelSummaryBase,
+  ExpansionPanelDetails as ExpansionPanelDetailsBase,
```
ryan-conn, line 6: Use `Accordion` instead of `ExpansionPanel` for these components:
image

ryan-conn (changes requested):

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.tsx
```diff
@@ -180,7 +179,9 @@ const CourseSelectColumn: React.FC = () => {
     if (card) {
       // Grow this card if it is focused so that
       // it can be viewed properly on low resolutions
-      const isExpandedRow = (card.collapsed === false);
+      const isExpandedRow = (card.collapsed === false
+        && !card.loading
+        && card.course);
```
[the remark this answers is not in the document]
eelyort: This is because we want a different message depending on if it has no sections vs. if there is no course selected. Previously this was done both here and in SectionSelect, but I moved both here since its easier.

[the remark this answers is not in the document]
eelyort: This is because we want a different message depending on if it has no sections vs. if there is no course selected. Previously this was done both here and in SectionSelect, but I moved both here since its easier.

autoscheduler/frontend/src/tests/ui/CourseSelectCard.test.tsx
```diff
@@ -309,7 +305,7 @@ describe('Course Select Card UI', () => {
   });
 
   describe('shows appropriate placeholder text', () => {
-    describe('when the customization level is Basic', () => {
+    describe('when the customization level is Select', () => {
```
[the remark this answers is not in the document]
eelyort: Done

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectColumn.css
```diff
@@ -45,7 +45,7 @@
 
 .expanded-row {
     overflow-y: hidden;
-    min-height: 500px;
```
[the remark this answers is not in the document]
eelyort: Can we talk about this in the meeting, I think we decided 273px as any larger would result in white-space when everything is fully collapsed.

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
eelyort: When its removed it breaks scrolling, let me test some alternate solutions.

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
eelyort: Our version of material ui seems super outdated, Accordion doesn't exist yet lol.

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
ryan-conn: I probably shouldn't have made this a suggestion because you need to change the usage of `toBool`, it broke the most recent build. I'm unresolving this just to make a note to fix it

eelyort: Oh sorry I fixed that already, didn't push yet

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
eelyort: Sounds good thanks.

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/CourseSelectCard.tsx
eelyort: This is fixed now

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
```diff
@@ -1,36 +1,110 @@
...
+    sortIsDescending: boolean;
+    // filters
+    hasHonors: boolean;
+    hasRemote: boolean;
+    hasAsynchronous: boolean;
+    honors: SectionFilter;
+    remote: SectionFilter;
+    asynchronous: SectionFilter;
+    includeFull: boolean;
+  } = useSelector<RootState, any>((state) => state.termData.courseCards[id]);
```
ryan-conn, line 107: This is a good example of when *not* to use destructuring: these declarations take 30 lines and only serve to make the code more ambiguous imo (since it's not immediately clear all these variables are all retrieved from course card data)

autoscheduler/frontend/src/tests/ui/CourseSelectColumn.test.tsx
```diff
@@ -22,20 +22,20 @@ beforeEach(() => {
...
+//       try {
+//         expect(element).toBeVisible();
+//         return true;
+//       } catch {
+//         return false;
+//       }
+//     }
+//     return false;
+//   };
+// }
```
ryan-conn, line 38: Make sure to remove unused code blocks like this instead of commenting them out when you commit

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.tsx
eelyort: Do you want me to go back to how it was before with the multiple `useSelector()`'s?

firejake308: I would recommend using a single `useSelector`, saving that as a temporary variable called something like `courseCardData`, and then pulling the fields individually from the `courseCardData` object

autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/SectionSelect.css
```diff
@@ -184,6 +184,101 @@
...
+    text-align: left;
+    display: inline-flex;
+    flex-direction: row;
+}
+/* sections stuff */
+.table-container .accordian-details .sections-wrapper {
+    display: flex;
+    flex-direction: column;
+    width: 100%;
+    height: fit-content;
```
firejake308, line 276: So this line is what fixes the scroll bar issue, but it also causes transitions to bug out?

ryan-conn (changes requested):
There are still a few bugs I found testing this out:
- No animation when collapsing cards
- No animation when expanding the sections accordion (interestingly it works correctly when expanding the filters)
- Expanding cards will immediately set the height to something like 800px, then slowly move to the correct size (especially noticeable with small cards since they will snap to a large size then shrink)
- Course cards no longer have a minimum height based on their content (instead being 38px), leading to situations where they're completely unreadable like below:
image

gannonprudhomme:
Closing this in favor of Ryan's later PR down the line

PR #596, opened 2021-06-28
Save schedule as image

### LINKED ISSUE #566: Add ability to expand schedule

gannonprudhomme:
## Description

Adds the ability to save schedule as an image. Note that this is based off of #587.

A few bugs that it has that I'm not sure how to fix:
- Can't save the schedule as an image on Chrome iOS. Might be able to fix by using [file-saver](https://www.npmjs.com/package/file-saver)
- Zoom on mobile messes up screenshot. Might be able to reset before the image is saved using styles
- If the window is too small and you save as an image, the meeting card text will be the smaller variant.

It is also failing tests due to some weird import thing that I plan on figuring out

## How to test

Press the save as image button in various circumstances.

## Screenshots

save-as-image-sc

## Related tasks

Closes #566


autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -465,17 +504,24 @@ const Schedule: React.FC = () => {
...
+    if (screenshot) return;
+
     throttle('', () => {}, 2 ** 31 - 1, true);
-  }, []);
+  }, [screenshot]);
+
+  // The screenshottable-schedule does not render corrctly if isLoadingAvailabilities == true
+  React.useEffect(() => {
+    if (screenshot) setIsLoadingAvailabilities(false);
+  }, [screenshot]);
```
ryan-conn, line 521:
```ts
const [isLoadingAvailabilities, setIsLoadingAvailabilities] = React.useState(!screenshot);
```
Isn't it a better idea to set the initial state correctly like above rather than set it in a hook?

autoscheduler/frontend/src/components/SchedulingPage/SchedulingPage.tsx
```diff
@@ -68,6 +99,17 @@ const SchedulingPage: React.FC<SchedulingPageProps> = ({
...
+                <div className={styles.loadingIndicatorContainer}>
+                  <SmallFastProgress />
+                </div>
+              ) : (
+                <Tooltip title="Save as image">
+                  <IconButton onClick={(): void => handleClick()}>
+                    <SaveAlt color="primary" />
+                  </IconButton>
+                </Tooltip>
+              )}
```
ryan-conn, line 112: Comment for this and the other ternary in the return value: I prefer moving conditional rendering outside of component returns to keep it somewhat organized

autoscheduler/frontend/src/utils/html2canvas.js
```diff
@@ -0,0 +1,11 @@
+import * as html2canvas from 'html2canvas'; // This import is incorrect, so we made this a JS file
```
ryan-conn, line 1: Allegedly html2canvas already has type definitions, are they incorrect? I also noticed `package.json` uses v1.0.0 whereas the package is at v1.3.2, what's the reason for using an old version?

autoscheduler/frontend/src/utils/saveImage.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Given a data URL of an object, saves it as the given filename and displays the save as download
+ * prompt.
+ * @param url The (data) URL of the object to save
+ * @param filename The name of the file to save
+ */
+export default function saveImage(url: string, filename: string): void {
```
ryan-conn, line 7:
```suggestion
export default function saveDataURL(url: string, filename: string): void {
```
Should probably change the name to be more accurate since it's filetype-agnostic

ryan-conn (changes requested):
A few comments, one other thing I'd like to see changed is having a solid colored background rather than the mix of white and gray:
image
It seems like you should be able to do this by just setting `background-color: initial` on the `hour-label` CSS class for schedule images

autoscheduler/frontend/src/utils/html2canvas.js
gannonprudhomme: We actually use `v1.0.0-rc.7` which was the most up-to-date version when I made this 3 months ago - they only started recently updating it again apparently, but I will update it to use the most up-to-date version.

And yes, the type definitions were incorrect. Seems like per https://github.com/niklasvh/html2canvas/issues/2628 that if we updated to TypeScript 3.8 (currently on 3.6.4) the types would work correctly, which I will try to do

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
```diff
@@ -31,27 +31,48 @@ const throttle = createThrottleFunction();
...
-  );
-  // Needed for saving availabilities
-  const term = useSelector<RootState, string>((state) => state.termData.term);
 
-  const fullscreen = useSelector<RootState, boolean>((state) => state.fullscreen);
+  // No need to listen for changes to these whenever we're rendering the duplicate
+  // screenshottable schedule
+  const availabilityList = useSelector<RootState, Availability[]>((state) => {
+    if (!screenshot) return state.termData.availability;
+    return [];
```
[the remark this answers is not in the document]
gannonprudhomme: Agreed

autoscheduler/frontend/src/components/SchedulingPage/Schedule/Schedule.tsx
gannonprudhomme: Yeah brainfart lol

autoscheduler/frontend/src/utils/html2canvas.js
gannonprudhomme: Ok so updated both and the issue is still occurring. I'm going to update `html2canvas` to `1.3.2` anyways but the typescript stuff is still messed up so the weird layout is going to remain the same.

PR #599, opened 2021-07-20
Fix Google Analytics bounce rate

### LINKED ISSUE #476: Fix Google Analytics bounce rate
B/c we're a "single page app" (b/c of our use of reach-router) we need to tell Google Analytics that the user is changing pages. See this for more: https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications

gannonprudhomme:
## Description

Basically just adds a .js and .d.ts file for calling the global `gtag` function.

Note that jest is also failing for some weird reason, which is identical to the failing tests in #596

## How to test

Honestly the only way I can think of testing this is to just deploy it and see if the bounce rate changes, so that's what I'm going to do.

## Related tasks

Closes #476

### DIFF
--- a/autoscheduler/frontend/src/components/LandingPage/LandingPage.tsx
+++ b/autoscheduler/frontend/src/components/LandingPage/LandingPage.tsx
@@ -7,12 +7,17 @@ import * as styles from './LandingPage.css';
 import About from './About/About';
 import PrivacyPolicy from './PrivacyPolicy/PrivacyPolicy';
 import setTerm from '../../redux/actions/term';
+import ga from '../../utils/ga';
 
 const LandingPage: React.FC<RouteComponentProps> = () => {
   const dispatch = useDispatch();
   React.useEffect(() => {
     dispatch(setTerm(null));
   }, [dispatch]);
+
+  React.useEffect(() => {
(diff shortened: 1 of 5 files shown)


gannonprudhomme:
Just deployed from the `deploy-fix-bounce-rate` branch that was based off of #544

gannonprudhomme:
So previously I don't think this was working since I wasn't calling `ga('event', 'page_view')`. Adding this seems to have fixed it, but we'll be able to be positive that it works after checking the bounce rate in a week after I deploy it.

gannonprudhomme:
Also need to confirm that the site still functions fine if an analytics is being ad-blocked

gannonprudhomme:
chrome_JlOP99TeVo

Confirmation that this has worked 🎉

ryan-conn:
Make sure you fix the testing/linting issues on this

> Also need to confirm that the site still functions fine if an analytics is being ad-blocked

It works with me on firefox when googletagmanager.com is being blocked, is this what you mean?

gannonprudhomme:
> Make sure you fix the testing/linting issues on this

The testing issues will have to wait until #596 is merged, as the babel + `html2canvas` fixes also apply to what I'm doing with `ga` here

PR #600, opened 2021-08-03
Automate scrape grades & scrape depts

### LINKED ISSUE #510: Add automating of scrape_grades every semester
A few times a year (around 3) grade distributions are released by the University. We should scrape these on a schedule so the entire process is automated and hands-free for when we go into LTS mode.

We should also figure out when they're typically released so we can scrape for them accordingly.

You can check when they're released here: https://web-as.tamu.edu/gradereports/

### LINKED ISSUE #529: Add automating of scrape_depts every semester
We just ran into the problem that the website showed Fall 2021, but no courses were actually scraped. This is because we didn't run `scrape_depts`!

As such, like #510, we should run `scrape_depts` every once in awhile.

A good solution might be to just run it with `scrape_courses` when we fall on the "cutoff days" - 3/22 and 10/26 - although it would probably be better to do it more regularly just in case it gets missed during then.

Might not hurt to just run it once a day every day?

gannonprudhomme:
## Description

This adds a `--recent` flag to both scrape grades and scrape depts, as well as the `--discord` flag for sending success/failure messages to a specified Discord channel. It also adds a workflow with cron schedules for running the jobs on a schedule.

## How to test

There's a few options, but the easiest would probably be to comment out the `on: schedule: ...` part like I did in f108fc7 and force push it to a separate branch then check actions (or the Discord channel) to confirm it works. 

To test the Discord bot locally (just the `--discord` flag), message me for the DISCORD_BOT_TOKEN and DISCORD_SCRAPE_CHANNEL_ID to add to the `.env` file.

## Related tasks

Closes #510, closes #529


ryan-conn (changes requested):
Looks good with a couple nitpicks

Also make sure you run this to make sure the discord integration works before we merge this

ryan-conn (approved):
Thanks for testing the discord integration
image

PR #601, opened 2021-08-08
Feedback form

### LINKED ISSUE #597: Make a feedback form component
We should make a little widget that lets you give a rating and provide quick feedback, similar to what gt-scheduler.org does (shameless feature steal). It would simply send a route to the backend, which would add a record to a table as well as send a message in our discord server.

ryan-conn:
## Description
Adds a feedback form accessible from the navbar. Note that even though we wanted to implement discord notifications, I'm waiting on Gannon to make that possible so we can just ignore it for now and do that in another PR or make the change to this once it's done.

## How to test
(Make sure you apply migrations before testing) click the button next to the login button on the navbar and try filling it with different ratings/comments, then check the DB to make sure they're submitted correctly

## Screenshots
|Before|After|
|--|--|
| image | image |

image

## Related tasks
Closes #597.


autoscheduler/feedback/views.py
```diff
@@ -0,0 +1,18 @@
...
+from rest_framework.parsers import JSONParser
+from rest_framework.response import Response
+from feedback.models import FeedbackFormResponse
+
+@api_view(['POST'])
+@parser_classes([JSONParser])
+def submit_feedback(request):
+    """ Creates a FeedbackFormResponse object corresponding to the sent data """
+    rating = request.data.get('rating')
+    comment = request.data.get('comment')
```
gannonprudhomme, line 11: Any particular reason you chose to put the information in the request body rather than as query parameters? Don't see why this route couldn't just be used like `/feedback/submit?comment="asdfassadf"&rating=5`

gannonprudhomme (changes requested):
Looking great! The only thing I really want to be changed is
- There should be an X at the top right of the dialog like our other dialogs
- Needs to be based off of #600 and add functionality to send the Discord message to the respective channel

gannonprudhomme:
Also can you add the feedback folder to `backend-workflow.yml` for linting? We have to manually add the folder for it (we also forgot to do the same for `user_sessions` =P)

autoscheduler/feedback/views.py
ryan-conn: Browsers have max URL lengths (generally 2MB which is problematic if you submit feedback close to the length limit) and I also want to avoid any problems with URL encoding/decoding from potentially cropping up

gannonprudhomme:
Btw this needs to be updated with master!

PR #603, opened 2021-08-19
Video web conf instructional method

ryan-conn:
## Description
Adds support for the synchronous video/web conf instructional method

## How to test
Scrape courses and look at AERO 301-501 or PGAdmin and filter:
`
SELECT * FROM public.sections WHERE instructional_method = 'SYNCVD';
`

## Related tasks
I couldn't actually find the issue for this but it's necessary for obvious reasons


autoscheduler/frontend/src/components/SchedulingPage/CourseSelectColumn/CourseSelectCard/ExpandedCourseCard/SectionSelect/InstructionalMethodIcon/InstructionalMethodIcon.tsx
```diff
@@ -11,35 +11,42 @@ interface InstructionalMethodIconProps {
...
-const instructionalMethodIcons = new Map<InstructionalMethod, JSX.Element>([
-  [InstructionalMethod.F2F, <People color="action" fontSize="small" />],
-  [InstructionalMethod.INTERNSHIP, <Work color="action" fontSize="small" />],
-  [InstructionalMethod.NONTRADITIONAL, <Search color="action" fontSize="small" />],
-  [InstructionalMethod.WEB_BASED, <Language color="action" fontSize="small" />],
-  [InstructionalMethod.STUDY_ABROAD, <Flight color="action" fontSize="small" />],
-  [InstructionalMethod.NONE, null],
-  [InstructionalMethod.REMOTE, <Laptop color="action" fontSize="small" />],
-  [InstructionalMethod.F2F_REMOTE_OPTION, (
+const instructionalMethodIcons: Record<InstructionalMethod, JSX.Element> = {
```
ryan-conn, line 14: fyi before making a comment about this: using typescript's Record means that every value must be defined vs. a map where properties are optional. Most places we're using maps we should be using Records instead to prevent dumb mistakes where we forget to set a value

gannonprudhomme (approved):
Looks good! Thanks for making this PR so quick :)

